PR #1, opened 2014-07-11
[ADD] mrp_merge_productions module

veloutin:
Add a module that allows merging Manufacturing Orders.


mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+                      })
+        m_prod = self.pool.get("mrp.production")
+        m_picking = self.pool.get("stock.picking")
+        # force unlinking of MOs through super calls
+        orm.Model.unlink(m_prod, cr, uid, rm_prod_ids)
+        orm.Model.unlink(m_picking, cr, uid, rm_picking_ids)
+        return None
+
+
+mrp_production_merge_wizard()
```
pedrobaeza: This is not needed anymore from 6.1

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+                "move_lines": [(5, )],
+                "move_lines2": [(5, )],
+            })
+
+            rm_prod_ids.append(prod.id)
+
+        target.write({"move_lines": [(4, mid) for mid in moves_to_link],
+                      "move_lines2": [(4, mid) for mid in moves_to_link2],
+                      })
+        m_prod = self.pool.get("mrp.production")
```
pedrobaeza: You can use abbreviated form: self.pool['mrp.production']

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+        m_picking = self.pool.get("stock.picking")
+        # force unlinking of MOs through super calls
+        orm.Model.unlink(m_prod, cr, uid, rm_prod_ids)
```
pedrobaeza: Usual way of doing this is:

m_prod,unlink(cr, uid, rm_prod_ids).

You also need to propagate context to avoid collateral effects.

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+                "move_lines2": [(5, )],
+            })
+
+            rm_prod_ids.append(prod.id)
+
+        target.write({"move_lines": [(4, mid) for mid in moves_to_link],
+                      "move_lines2": [(4, mid) for mid in moves_to_link2],
+                      })
+        m_prod = self.pool.get("mrp.production")
+        m_picking = self.pool.get("stock.picking")
```
pedrobaeza: These kind of variables are usually named with _obj suffix instead your nomenclature.

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm, osv, fields
```
pedrobaeza: You can remove osv if you change osv.except_osv to new orm.except_orm

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+        "selected_merge_ids": fields.many2many("mrp.production", "m2m_prod_merge_wizard_sel_ids",
+                                               "wizard_id", "production_id", "Selected Productions"),
+        "possible_merge_ids": fields.one2many("mrp.production.merge.choice", "wizard_id", "Merge Choices"),
+    }
+
+    def _get_merge_choices(self, cr, uid, target_id):
+        mrp = self.pool.get("mrp.production")
+        target = mrp.browse(cr, uid, target_id)
+        if not target:
+            raise osv.except_osv(
```
pedrobaeza: orm.except_orm

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+            ('location_dest_id', '=', target.location_dest_id.id),
+            ('company_id', '=', target.company_id.id),
+        ])
+
+        return production_ids
+
+    def _get_single_target(self, cr, uid, target_id):
+        target = self.pool.get("mrp.production").browse(cr, uid, target_id)
+        if target.state in ('done', 'cancel'):
+            raise osv.except_osv(
```
pedrobaeza: orm.except_orm

mrp_merge_productions/wizard/merge_productions.py
```diff
@@ -0,0 +1,309 @@
...
+                                # Don't allow a change of product or unit
+                                if to_merge.product_id.id != orig_product:
+                                    break
+                                if id_of(to_merge.product_uom) != orig_uom:
+                                    break
+
+                        # Found a target, stop searching
+                        break
+                else:
+                    raise osv.except_osv(
```
pedrobaeza: orm.except_orm

pedrobaeza:
Hi, Vincent,

Thank you very much for your contribution.

I have commented some things that you can improve to fulfill v7 and OCA conventions and style.

You have also to remove company logo in the module.

Regards.

mrp_merge_productions/wizard/merge_productions.py
veloutin: In this case, m_prod.unlink(...) fails since the unlink() implementation in mrp_production does not allow unlinking productions in states other than draft or cancel. This is why this code calls it through orm.Model.unlink().

pedrobaeza: Yeah, but this way, you are overpassing unlik method override chain, possibly inhibiting some restrictions. I'm thinking an alternative way, passing a context argument and processing it on an unlink override, but the problem remains the same. Probably we need to do it that way...

veloutin:
I have removed the wizard that appeared when you asked to merge, since it was clumsy and didn't actually help for anything. Instead, a production is selected for merging automatically. There is also a test now which can be added to if the current merge behavior has issues.

mrp_merge_productions/merge_productions.py
```diff
@@ -0,0 +1,297 @@
...
+                break
+        else:
+            raise orm.except_orm(
+                _("Unable to merge 'To Produce' move"),
+                _("No candidate to merge move %s") % (to_merge.id, ),
+            )
+
+
+class mrp_production(orm.Model):
+    _inherit = 'mrp.production'
```
bwrsandman: Please name the script after the model
`mrp_production.py`

bwrsandman:
@veloutin https://github.com/OCA/manufacture/pull/1/files#diff-5
Did you mean to include this?

bwrsandman:
:+1:

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,297 @@
...
+
+            # Keep the workcenter lines if any
+            workcenter_lines.extend(line.id for line in prod.workcenter_lines)
+
+            # Moves "to produce" must be merged with the ones from the target
+            # production.
+            merge_moves(self, cr, uid,
+                        prod.move_created_ids, moves_to_produce,
+                        move_updates, moves_to_cancel)
+            # Moves "ton consume" should be merged as well
```
rvalyi: typo: to consume

rvalyi:
:-1: 

Hello,

this looks nice but I have a still little problem with the _do_merge_productions method, 

1) it's nearly 100 lines. That's not explicit and hard to override.
I'm not too sure, but what about for instance extracting a method in the for prod in productions loop?
https://github.com/OCA/manufacture/pull/1/files#diff-599990eab23f785a590a000fcd560ce0R213
Somewhere else?

2) _do_merge_productions is returning None. Again not great for overriders. Isn't there anything more meaningful about what has been done that can be returned instead?

max3903:
:+1:

rvalyi:
Hello, for the record I still maintain my remarks here: https://github.com/OCA/manufacture/pull/1#issuecomment-49804061

I don't want to be boring with existing legacy codebase we have to maintain, but as for new modules I prefer having them avoid these issues.

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,308 @@
...
+
+
+class mrp_production(orm.Model):
+    _inherit = 'mrp.production'
+
+    def _check_productions_mergeable(self, productions):
+        if any(production.state in ('done', 'cancel', 'in_production')
+               for production in productions):
+            raise orm.except_orm(
+                _("Invalid Selecetion"),
```
gdgellatly: Typo for selection

gdgellatly:
Hi,

Interesting module.  I just read through the code so no test but I like it.  Nicely written.

I am not 100% sure but the comment on line 41 seems different to the sort key on line 42 in that it ignores "in production" state in mrp_merge_productions/mrp_production.py.  That comment is in conflict with the documentation in openerp.py also on line 41.  Also in openerp.py intended is spelt wrong.  On line 159 Selection is spelt wrong.

bwrsandman:
:+1:

veloutin:
@gdgellatly Thanks for pointing out the typos and outdated comment, all fixed.

gdgellatly:
Hi,

I just took another look at this.  There are a whole heap of functions that will only ever pertain to manufacturing orders kept outside the class.  I know its not explicitly in the style guide, but where a method really does belong to a class or instance, but does not need to use self I far prefer them included as static methods of the class.  Just makes it clearer.   Add them into the rest and we have in effect a non odoo style overidable ~300 line method

As for the other more general helper methods, the reality is no one is ever going to make their module depend on this module just to access them so maybe they coud be stuck in the class as well seeing they aren't used anywhere else, or maybe if for some reason they were going to get widespread adoption then in a completely different module.

max3903:
@veloutin did/can you take @rvalyi 's comments into account ?

veloutin:
@max3903 , @rvalyi : I agree that there is a function that's a bit on the heavy side. The fact that there are 10 fields to merge makes it grow really fast. As far as extending, I did not have anything meaningful to return. Perhaps someone who wants to extend will know what to change and what to return.

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+
+def id_of(obj):
+    """ For o2m relations, returns the foreign id if it is set or None """
+    if obj:
+        return obj.id
+    else:
+        return None
+
+
+def get_target(productions):
```
clonedagain, line 38: A docstring would help make the intent clear

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+
+    productions = sorted(productions, key=sort_key)
+    return productions[0], productions[1:]
+
+
+def name_of(obj):
+    return getattr(obj, "name", repr(obj))
+
+
+def moves_mergeable(source, target):
```
clonedagain, line 55: A docstring would be welcome here too

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+               for production in productions):
+            raise orm.except_orm(
+                _("Invalid Selection"),
+                _("You cannot merge manufacturing orders that are done, "
+                  "in production or cancelled."),
+            )
+
+        unique_attrs = (
+            ("product_id", _("Product")),
+            ("bom_id", _("Bill of Materials")),
```
clonedagain, line 170: Should'nt you consider the routing too ?
I'd not want to merge MOs that are not made using the same manufacuturing process

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+        moves_to_produce = [move for move in target.move_created_ids]
+        moves_to_consume = [move for move in target.move_lines]
+        scheduled_lines = [line for line in target.product_lines]
+        moves_to_link2 = []
+        moves_to_cancel = []
+        workcenter_lines = []
+        move_updates = {}
+        prodlines_to_delete = []
+        prodline_updates = {}
+        # XXX Should we merge notes or add notes of any kind?
```
clonedagain, line 228: Yes merging notes probably makes sense, but duplicate strings should be avoided.
I don't think adding notes is wanted, but writing to the document's chat/wall could certainly be useful.
That's not a blocking point however

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def lines_mergeable(source, target):
+    return source.product_id.id == target.product_id.id
+
+
+def merge_prodlines(obj, cr, uid, source_lines, target_lines, updates,
+                    to_delete):
+    for to_merge in source_lines:
+        for mergeable in target_lines:
```
clonedagain, line 73: why not imply call your iteration variables "source" and "target"?

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+                )
+                _logger.debug("Merging prodline %d into %d (%d %s)",
+                              to_merge.id, mergeable.id,
+                              to_merge.product_qty, to_merge.product_id.name)
+                to_write["product_qty"] += to_merge.product_qty
+                if to_merge.product_uos_qty:
+                    to_write["product_uos_qty"] += to_merge.product_uos_qty
+                to_delete.append(to_merge.id)
+                break
+        else:
```
clonedagain, line 88: Is the indentation right here ? Looks like an unusual "for ...: / else: " construct.
Probably it is, just making sure.

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+def lines_mergeable(source, target):
+    return source.product_id.id == target.product_id.id
+
+
+def merge_prodlines(obj, cr, uid, source_lines, target_lines, updates,
+                    to_delete):
+    for to_merge in source_lines:
+        for mergeable in target_lines:
+            if lines_mergeable(to_merge, mergeable):
+                to_write = updates.setdefault(
```
clonedagain, line 75: Humm, using SETdefault to actually GET a dict from a dict...
A comment would make the intent clearer.

mrp_merge_productions/mrp_production.py
```diff
@@ -0,0 +1,320 @@
...
+            "picking_id": target.picking_id and target.picking_id.id,
+        }
+
+        rm_picking_ids = []
+        rm_prod_ids = []
+
+        moves_to_produce = [move for move in target.move_created_ids]
+        moves_to_consume = [move for move in target.move_lines]
+        scheduled_lines = [line for line in target.product_lines]
+        moves_to_link2 = []
```
clonedagain, line 222: why 2 ?

veloutin:
_needs fixing_

pedrobaeza:
@carlosliebana, are you going to work on this?

pedrobaeza:
I think this one is not going to be resumed, so I close.

PR #11, opened 2014-12-30
Add french translations for mrp_bom_product_details

arthru:


pedrobaeza:
Files are technically correct, so :+1:, but I cannot say about the translation.

anajuaristi:
:+1: 

Single doubt.

Why Bill of materials is translated as Nomenclature and not like "liste des matériaux"?

max3903:
@anajuaristi This is the right term to use in the current context and also the one used in the official module:
https://github.com/odoo/odoo/blob/8.0/addons/mrp/i18n/fr.po#L369

max3903:
:+1:

PR #12, opened 2014-12-31
Create initial german translation (new de.po for each module)

rschnapka:
created de.po in i18n for each module


pedrobaeza:
Seems correct technically speaking. :+1:

yvaucher:
:+1:

PR #22, opened 2015-03-22
7.0 mrp production autocomplete

sebastienbeau:
This module will set in done all manufacturing Order for the product which have the option "Complete Manuf. Order" tic. The validation is done by a scheduled action.

A case of use is when you have an external system to manage the production, but this system can not trigger when the manufacturing order have been done. So as you still want to manage the consumation of goods and the processing of move for the picking, You can automatically set the production order in done with a schedule action.


Viggor:
Except pep8 validation, all is ok. :+1:

PR #25, opened 2015-04-30
[ADD] Module "mrp_production_merge_component".

JordiBForgeFlow:
This module was written to extend the manufacturing capabilities of Odoo,
and allows manufacturing orders to compute the required components, so that
the same components of the BOM found at different levels of the  hierarchy
are merged into the same manufacturing order line.

This allows for a more optimized procurement process, since all the components
will be ordered in the same Purchase Order Line.

This solution is specific to 7.0. In version 8.0 onwards the procurement
process will merge various procurement orders for the same product and
supplier into the same Purchase Order Line.


pedrobaeza:
Check PEP8:

```
./mrp_production_merge_component/model/mrp.py:22:1: F401 'fields' imported but unused
./mrp_production_merge_component/model/mrp.py:30:80: E501 line too long (81 > 79 characters)
./mrp_production_merge_component/model/mrp.py:67:80: E501 line too long (80 > 79 characters)
```

The rest seems good, but I'm concerned about possible extensions that add some fields to the computation, because you're overwriting lines. Maybe you can copy all possible values in your new lines?

PR #29, opened 2015-06-02
Add mrp_tracking_label

florian-dacosta:


sebastienbeau:
This PR should be proposed here : https://github.com/OCA/manufacture-reporting

PR #36, opened 2015-08-05
[ADD] Migrate < mrp_production_partner_note > from odoomrp-utils

oihane:


mrp_partner_note/README.rst
```diff
@@ -0,0 +1,44 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Partner production notes
+========================
+
+This module adds the possibility of defining a note for production in the
+partner so when the MO is automatically created it will be written this note
```
pedrobaeza: s/created it will written this note/created, this note will be written

mrp_partner_note/models/procurement_order.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class ProcurementOrder(models.Model):
+    _inherit = 'procurement.order'
+
+    @api.model
+    def _prepare_mo_vals(self, procurement):
+        res = super(ProcurementOrder, self)._prepare_mo_vals(procurement)
+        sale_proc = procurement.move_dest_id.procurement_id
+        res['notes'] = sale_proc.sale_line_id.order_id.partner_id.mrp_notes
```
pedrobaeza: Don't overwrite this field if there's no value defined (maybe another module write this value)

pedrobaeza:
Module name should _mrp_production_partner_note_

mrp_partner_note/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+    "contributors": [
+        "Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>",
+        "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
+        "Ana Juaristi <anajuaristi@avanzosc.es>",
+    ],
+    "category": "Tools",
+    "depends": [
+        "sale",
+        "mrp",
+        "mrp_production_note",
```
nhomar: if you depend of mrp_production_note which depends of mrp you do not need mrp here.

oihane:
Module renamed
Description changed
'Note' rewritting avoided.
`mrp` dependency removed

nhomar:
I'll give my +1 here because it is a fairly simple feature, but IMHO there is several cases where this note can be automatically written then may be in a future we can improve the behavior.

Regards.

pedrobaeza:
What cases do you have in mind, @nhomar? It not developed, they can be put in the Roadmap section.

oihane:
@pedrobaeza done changes

pedrobaeza:
:+1:

nhomar:
2015-08-05 4:41 GMT-05:00 Pedro M. Baeza notifications@github.com:

> What cases do you have in mind, @nhomar https://github.com/nhomar? It
> not developed, they can be put in the Roadmap section.
- Information about the product (the same case that you have here related
  to partner but related to the product).
- Information about the raw of materials.
- About the location.

....

It means we have several many2one fields were the notes can be
automatically pulled.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

pedrobaeza:
Yeah, that's why I say about not overwriting the field if there's nothing to put.

mrp_production_partner_note/models/procurement_order.py
```diff
@@ -0,0 +1,19 @@
...
+class ProcurementOrder(models.Model):
+    _inherit = 'procurement.order'
+
+    @api.model
+    def _prepare_mo_vals(self, procurement):
+        res = super(ProcurementOrder, self)._prepare_mo_vals(procurement)
+        sale_proc = procurement.move_dest_id.procurement_id
+        mrp_notes = sale_proc.sale_line_id.order_id.partner_id.mrp_notes
+        if mrp_notes:
+            res['notes'] = mrp_notes
```
nhomar: Here you are overwritting.

It should be an append dont you think?

Something like:

res['notes'] = '%s\n\n---\n\n%s' % (res.get('notes'), notes)

With that you ensure if any other module is creating notes you will not lose them.

pedrobaeza: Yeah, you're right. This should be done this way then:

```
if mrp_notes:
    res['notes'] = (res.get('notes', '') + "\n") if res.get('notes') else ''
    res['notes'] += mrp_notes
```

oihane: But in that case, if there is not 'notes' in res it will write the line breaks...

pedrobaeza: @oihane , see my comment for the solution

nhomar: I personally do not like sum strings.

`<<This is a personal preference>>`
I prefer the format instead (but it is a personal preference to avoid any "type problem").

I prefer format or concatenation instead the (+) but it is ok.

pedrobaeza: In this case, there's no problem, as you are concatenating strings. The problem comes when passing other variable types.

pedrobaeza:
:+1:

mrp_production_partner_note/models/procurement_order.py
nhomar: yea, but somebody can reuse your method and pass something else, the validation for sum is less explicit than the one for format.

``` python
Python 2.7.6 (default, Sep  9 2014, 15:04:36)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'text' + 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: cannot concatenate 'str' and 'int' objects
>>> 'text  {value}'.format(value=1)
'text  1'
>>>
```

It is more secure use the correct approach for strings than the + element at least that you want a hiden error in some moment.

pedrobaeza: But there's no such possibility, because values can't be other type than string... but I agree it's more secure in general terms (but the length of the line is higher also, that's why in this case I preferred the other formula to fit in less lines).

oihane: So how about this...

```
if mrp_notes:
    res['notes'] = ('%s\n' % res.get('notes', '')) if res.get('notes') else ''
    res['notes'] += mrp_notes
```

nhomar:
``` python
if mrp_notes:
    res['notes'] = res.get('notes') and ('{other}\n'.format(other=res.get('notes', '')) or ''
    res['notes'] = '{mine}'.format(mine=mrp_notes)
```

The same 2 lines :-)

pedrobaeza: Cool, any of these variations fit in a line, so :+1:

oihane: Something seems to be missing in @nhomar 's solution... second line rewrites res['notes']...

pedrobaeza: Yeah, it should be `+=` in the second line

mrp_production_partner_note/models/procurement_order.py
```diff
@@ -0,0 +1,22 @@
...
+    @api.model
+    def _prepare_mo_vals(self, procurement):
+        res = super(ProcurementOrder, self)._prepare_mo_vals(procurement)
+        sale_proc = procurement.move_dest_id.procurement_id
+        mrp_notes = sale_proc.sale_line_id.order_id.partner_id.mrp_notes
+        if mrp_notes:
+            res['notes'] = (
+                res.get('notes') and
+                ('{other}\n'.format(other=res.get('notes', ''))) or '')
+            res['notes'] += '{mine}'.format(mine=mrp_notes)
```
oihane: @nhomar is it better like this?

mrp_production_partner_note/models/procurement_order.py
nhomar: I am so sorry you are right.
 the `+=` sum also using numeric elemente.

The correct approach.

``` python
    old = res.get('notes') and ('{other}\n'.format(other=res.get('notes', '')) or ''
    res['notes'] = '{old}{mine}'.format(old=old, mine=mrp_notes)
```

mrp_production_partner_note/models/procurement_order.py
nhomar: I commented above but I put you here:

I am so sorry you are right.
 the `+=` sum also using numeric elemente.

The correct approach.

``` python
    old = res.get('notes') and ('{other}\n'.format(other=res.get('notes', '')) or ''
    res['notes'] = '{old}{mine}'.format(old=old, mine=mrp_notes)
```

nhomar:
For me now is perfect. :+1:

anajuaristi:
:+1:

PR #37, opened 2015-09-08
[IMP] Port mrp_bom_product_details to 8.0

mistotebe:


LoisRForgeFlow (approved):
It has plurals in the name but also in v9 and v10 (merged), so I think we can merge this as well.

LoisRForgeFlow:
I tried to merge as it was green, but seems that after rebasing it is not. I will close as this is stale, please reopen and fix it if you think it is still relevant.

mistotebe:
AFAIK tests look like postgres failed but I don't care anymore, not about 8.0 anyway :)

LoisRForgeFlow:
Yes, travis definition should be fixed, if someone is still interested in v 8.0 can propose a fix for that branch (you can check branch 10.0 for example).

PR #40, opened 2015-10-01
[MIG] mrp_bom_version: migrated from odoomrp-wip

oihane:


mrp_bom_reference_selection/views/mrp_bom_view.xml
```diff
@@ -2,7 +2,7 @@
 <openerp>
     <data>
 
-        <record id="mrp_bom_tree_view" model="ir.ui.view">
+        <record id="mrp.mrp_bom_tree_parent_view" model="ir.ui.view">
```
oihane, line 5: By doing this we create a base view so that both modules can be installed separatelly and they won't collide between each other.

Correct option will be to have the view in `mrp` module but, as it is not, I think this is the best option.

oihane:
This is how it looks now
image

mrp_bom_version/README.rst
```diff
@@ -0,0 +1,60 @@
...
+    :alt: License: AGPL-3
+
+=================
+MRP - BoM Version
+=================
+
+This module provides a state in the BoM whether to allow their use in
+manufacturing, to do the following states are defined:
+
+* **Draft**:
```
pedrobaeza, line 17: Can you move this explanation to Usage section?

mrp_bom_version/models/mrp_bom.py
```diff
@@ -0,0 +1,152 @@
...
+            'version': self.version + 1,
+            'active': (self.company_id.active_draft if self.company_id else
+                       self.env.user.company_id.active_draft),
+            'parent_bom': self.id,
+        })
+        return new_bom
+
+    @api.multi
+    def button_activate(self):
+        self.ensure_one()
```
pedrobaeza: Why are you limiting to one record?

mrp_bom_version/models/mrp_bom.py
```diff
@@ -0,0 +1,152 @@
...
+                             copy=False, default=1)
+    parent_bom = fields.Many2one(
+        comodel_name='mrp.bom', string='Parent BoM')
+    old_versions = fields.Many2many(
+        comodel_name='mrp.bom', string='Old Versions',
+        compute='_get_old_versions')
+
+    @api.multi
+    def button_draft(self):
+        self.ensure_one()
```
pedrobaeza: Use write and don't limit to one record

mrp_bom_version/models/mrp_bom.py
```diff
@@ -0,0 +1,152 @@
...
+    def button_activate(self):
+        self.ensure_one()
+        self.write({
+            'active': True,
+            'state': 'active'
+        })
+
+    @api.multi
+    def button_historical(self):
+        self.ensure_one()
```
pedrobaeza: The same

mrp_bom_version/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+# (c) 2015 Oihane Crucelaegui - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    active_draft = fields.Boolean(
```
pedrobaeza: Why is this defined by company instead a global value?

oihane: Cause the only way I found to do is this way. It is defined in res.config... but saved in res.company

pedrobaeza: You can save it in a ir.value instead.

oihane: Can you define an ir.value by clicking in a res.config check???

pedrobaeza: You do it the load/save methods of the config model. See https://github.com/OCA/server-tools/pull/76/files#diff-30a60a6e49d7e65247cfbf6c6911e349R73 (with the recommendations)

oihane: I manage to set it to true, but it show as false, in all cases and I can't then set it to false back.

pedrobaeza:
:+1:

oihane:
@pedrobaeza something goes akward... it seems to not to execute coveralls.

pedrobaeza:
The link is not put in this PR, but it's executing it.

mikelarre:
:+1:

CarlosMolinaP:
:+1:

PR #45, opened 2015-11-02
[ADD] New module quality_control_stock

oihane:


pedrobaeza:
Can you squash a bit the commits? For example, a PR and its revert

oihane:
Ready for review, it has coveralls :100: %

quality_control_stock/models/qc_trigger.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, fields
+
+
+class QcTrigger(models.Model):
+    _inherit = 'qc.trigger'
+
+    picking_type = fields.Many2one(
+        comodel_name="stock.picking.type", readonly=True)
```
pedrobaeza: Put here `ondelete='cascade'`

quality_control_stock/models/stock_picking_type.py
```diff
@@ -0,0 +1,42 @@
...
+        if vals.get('name') or vals.get('warehouse_id'):
+            qc_trigger_model = self.env['qc.trigger'].sudo()
+            qc_trigger = qc_trigger_model.search(
+                [('picking_type', '=', self.id)])
+            qc_trigger.name = self.complete_name
+        return res
+
+    @api.multi
+    def unlink(self):
+        self.env['qc.trigger'].sudo().search(
```
pedrobaeza: Thinking about this, with an `ondelete="cascade"` in the field definition is enough.

oihane: are you sure, here it uses "sudo"

quality_control_stock/models/stock_picking_type.py
```diff
@@ -0,0 +1,42 @@
...
+from openerp import models, api
+
+
+class StockPickingType(models.Model):
+    _inherit = 'stock.picking.type'
+
+    @api.multi
+    def _create_qc_trigger(self):
+        qc_trigger = {
+            # Synchronize all translations
```
pedrobaeza: I think this is a TODO, not a comment, because this doesn't synchronize all translations. In fact, I think the name is not translatable, so we need to do that first.

quality_control_stock/tests/test_quality_control_stock.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def test_inspection_create_for_product(self):
+        self.product.qc_triggers = [(
+            0, 0, {
+                'trigger': self.trigger.id,
+                'test': self.test.id,
+            }
+        )]
+        self.picking1.do_transfer()
+        self.assertEqual(self.picking1.created_inspections, 1,
```
pedrobaeza, line 61: You should check in all that the created inspection is for the selected test on the trigger.

quality_control_stock/tests/test_quality_control_stock.py
```diff
@@ -0,0 +1,222 @@
...
+                         'Only one inspection must be created')
+
+    def test_inspection_create_for_product_partner(self):
+        self.product.qc_triggers = [(
+            0, 0, {
+                'trigger': self.trigger.id,
+                'test': self.test.id,
+                'partners': [(6, 0, [self.partner1.id])],
```
pedrobaeza: You can do instead: `'partners': [(6, 0, self.partner1.ids)],`

pedrobaeza:
Please add demo data with:
- A trigger for an existing demo product and the demo test linked.
- An inspection for any demo stock.move.

quality_control_stock/models/stock_picking_type.py
oihane: @pedrobaeza you developed this code, I can't say the meaning of this comment

pedrobaeza: Yes, please put it as TODO an in known issues.

oihane:
I did not manage to create a trigger for a demo product and the demo test.
It does not create the trigger until post-init, so there is not trigger to select to create the trigger_line.

pedrobaeza:
OK, I see. There's no possibility of such demo data then for now. One option can be to hard-coded any demo trigger and change post-init to only create non-existant triggers, but I don't know if this deserves the time.

agaldona:
:+1:

pedrobaeza:
:+1:

damendieta:
Hi, I'd like to migrate this module to v9 but the base repository is marked as "unknown repository".

Where can I get the code to base the migration?.

Thanks a lot.

Daniel.

PR #46, opened 2015-11-06
[ADD] New Module mrp_production_hierarchy

andrea4ever:
Add a hierarchy for class mrp.production and show in a tree like Product for Category


andrea4ever:
Wrong Target

PR #63, opened 2015-12-09
[FIX] mrp_bom_reference_selection: Fixed onchange error closes #62

### LINKED ISSUE #62: mocalculate data on MO does not bring the product_UOM
Create a bill of materials with a product in KG
Go to manufacturing order
Calculate data
The assigned product UOM in "planned materials" is Unit, so it's wrong and it's causing this error:

Field(s) product_uom failed against a constraint: You try to move a product using a UoM that is not compatible with the UoM of the product moved. Please use an UoM in the same UoM category.

This issue was reported  by me and @ybofr on this PR but you said it was not related to new hooks. I think it is, since the error appairs with "calculate data" button.
https://github.com/OCA/manufacture/pull/57

@pedrobaeza @rafaelbn @oihane

oihane:


mrp_bom_reference_selection/models/mrp_bom_line.py
```diff
@@ -49,23 +50,22 @@ def onchange_product_id(self, product_id):
 
         return res
 
-    @api.one
     @api.depends('product_id')
     def _get_child_bom_lines(self):
```
pedrobaeza: call it _compute_...

oihane:
@rafaelbn @pedrobaeza this fixes error reported by @anajuaristi @ybofr in the PR of _mrp_hook_ and _mrp_operations_extension_

pedrobaeza:
:+1: 

Fast-tracking for solving the current problem.

PR #72, opened 2016-01-14
[MIG] Make mrp_bom_note installable in v9

oihane:
- mrp_bom_note

### DIFF
--- a/mrp_bom_note/README.rst
+++ b/mrp_bom_note/README.rst
@@ -1,11 +1,21 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
     :alt: License: AGPL-3
 
-MRP - BoM notes
-===============
+==========================
+Notes in Bill of Materials
+==========================
 
 This module creates in BoM a new field to add notes
 
+Usage
+=====
+
+* Go to Manufacturing > Products > Bill of Materials
+* Edit an existing BoM and fill in the Notes field on tab Notes
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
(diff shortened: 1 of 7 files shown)


pedrobaeza:
One module per PR, please.

oihane:
Ok, changed, but what about the other two modules that one depends on the other?

mrp_bom_note/views/mrp_bom_view.xml
```diff
@@ -2,18 +2,18 @@
 <openerp>
     <data>
```
max3903: `<data>` can be removed.

oihane: @max3903 is that something about v9? What in case it is `noupdate`?

mrp_bom_note/README.rst
```diff
@@ -1,11 +1,18 @@
...
-===============
+==========================
+Notes in Bill of Materials
+==========================
 
 This module creates in BoM a new field to add notes
 
+Usage
+=====
+
```
max3903: * Go to Manufacturing > ...
* Edit an existing BOM and fill in the Notes field on tab X after field Y

mrp_bom_note/views/mrp_bom_view.xml
max3903: @oihane I don't know if `noupdate` works in `<odoo>`, but you don't have any here so it should work.

oihane:
@pedrobaeza @max3903 there is some issue with coveralls, but module is ready for review

max3903:
:+1: 

@oihane To answer your question about dependency, you need to migrate the dependencies first and once merged, you rebase to migrate your target module.

pedrobaeza:
:+1: Merging to continue the rest of the modules migration

PR #73, opened 2016-01-15
[MIG] Make mrp_production_note installable in v9

oihane:


mrp_production_note/views/mrp_production_view.xml
```diff
@@ -1,17 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
```
max3903: You can use `<odoo>`.

max3903:
@oihane In travis.yml of the repo, `matrix` is wrongly indented. As a consequence, the tests are not run properly. It should be at the same level as `global`.

Example in v8:
https://github.com/OCA/manufacture/blob/8.0/.travis.yml

Can you fix that please?

oihane:
@pedrobaeza @max3903 there is some issue with coveralls, but module is ready for review

max3903:
@oihane There is no issue with Coveralls. It computes the code coverage **difference** between your branch and OCA/9.0. As there is no code coverage yet on OCA/9.0, it stays in pending state.

:+1:

pedrobaeza:
There were a problem in Coveralls reported by its status agent, but now is correct.

:+1:

PR #74, opened 2016-01-16
[FIX] mrp_operations_project - fix operators time in WO

ddico:
I believe "project_task_work_mrp" view is defined in "mrp_operations_project" and not in "mrp_project". 

Also we need to add the "name" (Work Summary) in "project_task_work_mrp" view, otherwise we get a KeyError at save and you probably need the work summary anyway.

Now the "Operators Time" looks like this :

screen shot 2016-01-16 at 6 26 57 pm

### DIFF
--- a/mrp_operations_project/views/mrp_production_view.xml
+++ b/mrp_operations_project/views/mrp_production_view.xml
@@ -12,6 +12,7 @@
                            on_change="onchange_task_id(task_id)"
                     />
                     <field name="user_id"/>
+                    <field name="name"/>
                     <field name="date"/>
                     <field name="hours" widget="float_time"/>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1: Fast-tracking it...

Only one clarification. The field _name_ is not mandatory in standard, and that's why is not included by default, but it doesn't hurt also.

ddico:
Great! Thank you

JMODaERP:
Thank you very much again, for me it is very important to know the name of the operator ( expert) assigned to an operation . I vote to be included in the standard , albeit only on the text ; Done By. Although it is " merged " I approve too.     :+1: 
Excellent.

PR #76, opened 2016-01-20
[FIX] mrp_operations_project - fix incorrect operator name in tasks

ddico:
A workcenter with multiple operators will trigger identical tasks for the same operation.

screen shot 2016-01-20 at 1 50 35 am


anajuaristi:
Good point
If technically is rigth :+1 from my side

pedrobaeza:
Totally right :+1: 

Merging.

ddico:
That was fast :)

JMODaERP:
Ddico Many thanks , that's what I was looking for !, because that operate 90 % of the companies in Spain . Thank you very much insist . Excellent, although this " merged " I also give you my approval. :+1: 
 Excellent.

PR #77, opened 2016-01-20
[FIX] mrp_production_real_cost - create analytic lines before done

ddico:
Calling "action_done" before creating "account_analytic_line" triggers an "account_move_line" amount without the value of the last operation.

If we call "action_done" after the "account_analtic_line" is created it will give us the correct amount (including the last operation).

### DIFF
--- a/mrp_production_real_cost/models/mrp_production_workcenter_line.py
+++ b/mrp_production_real_cost/models/mrp_production_workcenter_line.py
@@ -97,7 +97,7 @@ class MrpProductionWorkcenterLine(models.Model):
 
     @api.multi
     def action_done(self):
-        result = super(MrpProductionWorkcenterLine, self).action_done()
         self._create_analytic_line()
         self._create_pre_post_cost_lines(cost_type='post')
+        result = super(MrpProductionWorkcenterLine, self).action_done()
         return result


pedrobaeza:
We are not using accounting in moves, but seems logic :+1: 

cc @rafaelbn @antespi

rafaelbn:
Tested, thanks! :+1:

oihane:
LGTM :+1: Thanks @ddico for the contribution

pedrobaeza:
We are now seeing that this breaks another thing: end_date is not assigned until super is called, so machine time is never filled. Don't you have any other method for getting accounting?

ddico:
Is like the chicken or the egg...

I don't have a clean way, even if I find a solution I don't think is going to make other users happy?

How about we do something similar like in:
https://github.com/OCA/manufacture/blob/8.0/mrp_operations_time_control/models/operation_time.py#L21
before calling "_create_analytic_line" (or in _create_analytic_line)

Do you think that will work? Would that be OK for everyone?

cc: @pedrobaeza @rafaelbn @antespi

pedrobaeza:
Done in https://github.com/OCA/manufacture/pull/78. Thanks for the tip.

PR #78, opened 2016-01-20
[IMP] mrp_production_real_cost: added test for onchange to increase c…

oihane:
…overalls

### DIFF
--- a/mrp_production_real_cost/models/mrp_production_workcenter_line.py
+++ b/mrp_production_real_cost/models/mrp_production_workcenter_line.py
@@ -31,16 +31,17 @@ class MrpProductionWorkcenterLine(models.Model):
                     (product.default_code or ''))
             general_acc = workcenter.costs_general_account_id or False
             price = -(workcenter.costs_hour * operation_line.uptime)
-            analytic_vals = production._prepare_real_cost_analytic_line(
-                journal_id, name, production, product,
-                general_account=general_acc, workorder=self,
-                qty=operation_line.uptime, amount=price)
-            task = task_obj.search([('mrp_production_id', '=', production.id),
-                                    ('workorder', '=', False)])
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Waiting for Travis result, :+1:

mrp_production_real_cost/models/mrp_production_workcenter_line.py
```diff
@@ -97,7 +97,7 @@ def action_pause(self):
 
     @api.multi
     def action_done(self):
+        result = super(MrpProductionWorkcenterLine, self).action_done()
```
oihane: This order change is because machine uptime is calculated with `start_date` and `end_date` if we create the analytic line before uptime will be 0, as `end_date` won't be defined (if the workorder has not been previously paused)

pedrobaeza: But this has been just changed in https://github.com/OCA/manufacture/pull/77 due to the reasons exposed there.

mrp_production_real_cost/models/mrp_production_workcenter_line.py
```diff
@@ -97,6 +98,8 @@ def action_pause(self):
 
     @api.multi
     def action_done(self):
+        not_paused_records = self.filtered(lambda x: x.state != 'pause')
+        not_paused_records._write_end_date_operation_line()
```
oihane, line 102: When a workorder is paused the end_date is written, if I do not do this, it will be changed.

pedrobaeza:
Nice! :+1: 

cc @rafaelbn @JavierIniesta

JavierIniesta:
:+1:

rafaelbn:
Thanks :+1:

PR #79, opened 2016-01-22
[IMP] mrp_operations_extension: WO finish, do not consume with the remaining raw materials

Daniel-CA:
- New behaviour added when finishing WO and there are pending movements

all info:
https://github.com/odoomrp/odoomrp-wip/pull/966

### DIFF
--- a/mrp_operations_extension/__openerp__.py
+++ b/mrp_operations_extension/__openerp__.py
@@ -30,6 +30,7 @@
     "data": [
         "data/mrp_operations_extension_data.xml",
         "wizard/mrp_workorder_produce_view.xml",
+        "wizard/workcerter_line_finish_view.xml",
         "views/mrp_workcenter_view.xml",
         "views/mrp_routing_operation_view.xml",
         "views/mrp_production_view.xml",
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Please add tests.

Daniel-CA:
Can someone check this? The coverage/coveralls is 100%

agaldona:
:+1:

Daniel-CA:
Updated tittle of the PR to clarify what is the reason for this:
When an operation is finished, it should not produce with the remaining raw materials.

anajuaristi:
It's not correctly explained yet. Produce is produce and consume is
consume. You don't produce with remaining row materials. It's wrong aproach.
It would be better something like ... to finish one operation all materials
have to be consumed or cancelled.

Daniel-CA:
Right, in operations you consume, I just misspelled, changed.

pedrobaeza:
OK, seems to be good enough :+1:

anajuaristi:
do not consume remaining raw materials serís entonces más correcto

oihane:
:+1: LGTM

anajuaristi:
@oihane @Daniel-CA @pedrobaeza 

We need this merged and it has got 3 :+1:  LGTM + mine so... I will merge in short time if there is no problem.

PR #80, opened 2016-01-23
[FIX] mrp_production_real_cost - override _costs_generate to avoid duplicates

ddico:
I would like to suggest that we override "_costs_generate" in order to avoid creating duplicate entries in account_analytic_line when action_production_end. 

I don't think this behaviour was intentional. The records created in "_costs_generate" don't have MO#, TASK#, WO# so it is not possible to relate them to the originating MO. If I understand correctly the scope of mrp_production_real_cost module I believe we should prevent these records to be created.

I'm going to attach a picture from account_analytic_line table as it's easier to see all the above.

screen shot 2016-01-20 at 11 52 33 pm

I hope i'm not missing anything here...

Thank you

### DIFF
--- a/mrp_production_real_cost/models/mrp_production.py
+++ b/mrp_production_real_cost/models/mrp_production.py
@@ -73,3 +73,13 @@ class MrpProduction(models.Model):
             'product_uom_id': product.uom_id.id,
             'general_account_id': general_account.id,
         }
+
+    @api.multi
+    def _costs_generate(self):
+        """
+        As we are generating the account_analytic_lines for MO in the
+        current module, we override this method in order to avoid
+        duplicates created in the parent class. Any other module
+        inheriting this method should take this into account!
+        """
+        return


pedrobaeza:
Uhm, curious, I would say that the method was inhibited somewhere, but it seems that not. Thanks for finding it.

:+1:

pedrobaeza:
Hi, @ddico, I'm thinking that better if we just rename our method to this name to keep maximum compatibility.

ddico:
@pedrobaeza - done

pedrobaeza:
I'm afraid that you have renamed the wrong method, because that one is for getting the compute fields, not for creating the costs.

ddico:
Sorry @pedrobaeza, I don't see which one are you referring to?

pedrobaeza:
OK, sorry, I saw it badly, because there's no equivalent method in current code, as the real costs are added on each workcenter line. Please return to the previous proposal and sorry for the noise.

ddico:
done

pedrobaeza:
Again :+1:

@anajuaristi @rafaelbn please approve

rafaelbn:
Thanks @ddico :+1:

oihane:
Good job :+1:

PR #82, opened 2016-01-24
[IMP] mrp_production_real_cost - save WO on AAL

ddico:
When account_analytic_line(s) are created they don't have the originating WO saved. With the exception of the first item in the materials all other are missing.

Workorder is defined as a related field (on task_id) here:
https://github.com/OCA/manufacture/blob/8.0/mrp_project/models/account_analytic_line.py#L14
So we don't need to add workorder to the analytic_vals, instead we just need to ensure the correct task is saved.

Before this PR:

screen shot 2016-01-24 at 9 52 17 am

With this PR:
screen shot 2016-01-24 at 9 53 16 am

### DIFF
--- a/mrp_project/models/account_analytic_line.py
+++ b/mrp_project/models/account_analytic_line.py
@@ -12,6 +12,5 @@ class AccountAnalyticLine(models.Model):
     mrp_production_id = fields.Many2one(
         comodel_name='mrp.production', string='Manufacturing Order')
     workorder = fields.Many2one(
-        comodel_name='mrp.production.workcenter.line', string='Work Order',
-        related="task_id.workorder", store=True)
+        comodel_name='mrp.production.workcenter.line', string='Work Order')
     task_id = fields.Many2one('project.task', 'Project Task')


ddico:
@pedrobaeza - not sure why this failed? The 3 files I changed are 100% in coveralls.

anajuaristi:
Hi @ddico

Wo can not be related to task since tasks are only created if there is at
less one operator asigned to the routing step.
Its correct that if task exists on analytic line, wo should be taken from
it but if there is no task for the operation we should be able to add costs
to it.
From my side I wouldn't merge the proposed changes.

ddico:
Hi @anajuaristi 

I'm aware that one task is created for the MO (I believe this was called before "General Task") and also that a new task is created for each end every operator in a routing operation and that we can track time for ALL of these tasks - and I'm not preventing any of that.

My PR was adding clarity to what is the "source" of an account analytic line. What WO was processed and generated this analytic line. 

If this level of detail is not required or useful then I do not see the point in having the "workorder" field in the "account_analytic_line" table. What is this field supposed to store and in which context? At the moment does not show anything useful except some incomplete information unfortunatelly.

pedrobaeza:
Instead of replacing what you have done, it's only needed to pass as argument the workcenter in the corresponding place. I will point you where to put it.

mrp_production_real_cost/models/stock_move.py
```diff
@@ -24,11 +24,11 @@ def action_done(self):
                     '-' + (record.product_id.default_code or ''))
             analytic_vals = (production._prepare_real_cost_analytic_line(
                 journal_id, name, production, record.product_id,
-                workorder=record.work_order, qty=record.product_qty,
+                qty=record.product_qty,
```
pedrobaeza: Maybe this one is not correctly pointed.

mrp_production_real_cost/models/mrp_production_workcenter_line.py
```diff
@@ -75,10 +75,10 @@ def _create_pre_post_cost_lines(self, cost_type='pre'):
         if price:
             analytic_vals = production._prepare_real_cost_analytic_line(
                 journal_id, name, production, product,
-                general_account=general_acc, workorder=self,
+                general_account=general_acc,
```
pedrobaeza: This should point to the workcenter, not the workcenter line

mrp_production_real_cost/models/mrp_production_workcenter_line.py
```diff
@@ -34,11 +34,11 @@ def _create_analytic_line(self):
             if price:
                 analytic_vals = production._prepare_real_cost_analytic_line(
                     journal_id, name, production, product,
-                    general_account=general_acc, workorder=self,
+                    general_account=general_acc,
```
pedrobaeza: Also to the worcenter instead of the workcenter line.

mrp_production_real_cost/models/mrp_production_workcenter_line.py
pedrobaeza: No, sorry, it's correct. It contains the workorder, not the workcenter.

pedrobaeza:
Seeing the code, I can't figure out where is the problem why the workorder is not passed, but the principle is correct (be passed as argument). If any problem as it may seems, what we have to correct is the passed argument.

ddico:
Hi @pedrobaeza 

In my PR I was taking advantage of what ORM does for a related field and is working as expected if we don't pass the workorder and if we have the correct task_id. 

As you can see in: https://github.com/OCA/manufacture/blob/8.0/mrp_project/models/account_analytic_line.py#L14

workorder = fields.Many2one(
        comodel_name='mrp.production.workcenter.line', string='Work Order',
        **related="task_id.workorder"**, store=True)

If the **related** field is not you wanted and you want to control what goes in this field then maybe this is what needs to be changed. 

I'm trying to understand this... so you would like having the workorder in the associated account_analytic_line but you don't want to have the task_id in there then? Because if you do want the task_id then why don't let the related field do what is supposed to do and save the workorder as well?

Thank you

pedrobaeza:
That's the problem then! That field shouldn't be related. I switch to related some fields, but after talking with @anajuaristi, I undo this change because the same reason she explained here, but I should forget this field. Can you please change that field for not being related?

anajuaristi:
@ddico...
Just to give you one case... supose that you have got one operation made
inside the routing but you only want to compute cost in generic way. If you
make wo related field you always need having a task. But you are rigth...
if task is created, wo id should be the same
@pedrobaeza rigth. I remember we had a talk about this subject.

ddico:
@pedrobaeza i've changed workorder in account_analytic_line and now is not a related field anymore. Now workorder info in account_analytic_line is correct as per the discussion above.

I would like to mention for everyone else interested in this subject, now all account_analytic_line records created when a MO is processed are linked to the task_id of the "General Task" (a generic task created by a MO in its project)

Thank you @anajuaristi for clarification.

mrp_project/models/account_analytic_line.py
```diff
@@ -13,5 +13,5 @@ class AccountAnalyticLine(models.Model):
         comodel_name='mrp.production', string='Manufacturing Order')
     workorder = fields.Many2one(
         comodel_name='mrp.production.workcenter.line', string='Work Order',
-        related="task_id.workorder", store=True)
+        store=True)
```
pedrobaeza: You can also remove `store`, as all regular fields are stored.

pedrobaeza:
Where is the code where analytic lines are binded to the general task?

ddico:
I'll remove `store`.

I was referring to:
`task = task_obj.search(
                    [('mrp_production_id', '=', production.id),
                     ('workorder', '=', False)])
`
and what that means.... that we are searching the "General Task" when `'workorder', '=', False`

ddico:
@pedrobaeza - done.

pedrobaeza:
:+1: 

I still don't have clear if the search you mentioned is already included in the repo or in one PR.

ddico:
It's on the repo, in multiple places. Example: [here](https://github.com/OCA/manufacture/blob/8.0/mrp_production_real_cost/models/mrp_production_workcenter_line.py#L41).

ddico:
Coverage fails because of unrelated issues!?

@pedrobaeza - what can I do to help?

pedrobaeza:
No special problem. See that coverage is not really dropping. It's not a problem.

oihane:
What a pitty the coveralls, code reviewed :+1:

JMODaERP:
O.K. :+1:

PR #83, opened 2016-01-24
[ADD] mrp_production_real_cost - add cost per cycle

ddico:
After a MO is completed we end up with following account analytic lines for each work order:
- Pre-operation cost - if pre-operation costing product is defined in the work center
- Material(s) cost - if any materials are to be consumed in this work order 
- Cost per hour - if defined in in the work centre
- Post-operation cost -  if post-operation costing product is defined in the work center

This PR will add one more account analytic line:
- Cost per cycle - if defined in in the work centre

Hope this helps


anajuaristi:
Tis functionality was present from begining. I didn't know it had been
erased.
For me is correct that operation has a cost per hour and/or cost per cycle
so if technically is correct +1 to merge from my side

pedrobaeza:
Yeah, I removed it because it creates duplicated cost for the same concept it you fill both cycles and duration. It was discussed with you, Ana, saying that I give priority to cycles.

anajuaristi:
But i told you that both conceps are different ones. Most of people is
using cost by hour for machines but there is companies using cost by cycle
and even both so from my side of view is wrong that behavior has been
erased.
If someone is not using one of them just put zero on cost hour or cost
cycle field and it's enough.
That's the way it was done when module was built. I don't understand the
reason of erasing that functionality so I see rigth that it is added again

pedrobaeza:
I told you that it confuses a lot the user, because they tend to fill both without knowing that it's going to impact 2 times. @ddico, do you know how standard code makes this?

anajuaristi:
I think that's users training issue not system issue. It can also be
explained on field help.
From my point of view It's more confusing having a field on workcenter
that's doing nothing.

I say again that for me the change is rigth because imho not having it
causes more issues than having it.

pedrobaeza:
It does something. It's used when the other data is empty.

ddico:
I agree with @anajuaristi. It's confusing when you see that cost per cycle does nothing. If a user does not need cost per cycle there is always the option of not filling in the amount.

If you agree, I could improve the field help and/or readme file.

pedrobaeza:
@rafaelbn @JMODaERP, opinion?

rafaelbn:
Hi,
- "It's confusing when you see that cost per cycle does nothing" - agree
- "If a user does not need cost per cycle there is always the option of not filling in the amount" - disagree, usability is very important.
- "I told you that it confuses a lot the user, because they tend to fill both without knowing that it's going to impact 2 times" - agree
- "I think that's users training issue not system issue" - **a good system prevents user errors**

I think cost per cycle must be a calculate fields. Cost per hour, never changes. So:
- If cost per cycle is not filled it's calculated
- If filled what is calculated is time for 1 cycle (hour)



So I'm not agree adding cost per cycle in this way.

ddico:
Hi @rafaelbn 

Thank you for your review.

I'm concerned with what you have stated and the way you have calculated "cost per cycle" in the image above.

My PR is calculating the cost per cycle as:
 `mrp_workcenter.costs_cycle` multiplied with `mrp_production_workcenter_line.cycle` 
and is has nothing to do with "cost per hour".

I think before we can move on with this we should all agree on the definition of these two costs and why we have them.

#### Odoo's definition:

**Cost per Hour** -  cost of work centre per hour.
**Cost per Cycle** - cost of work centre per cycle.

Very generic no? So, what does this mean? 

In this module `mrp_production_real_cost` we are trying to give users an accurate calculation of the cost of the product.

#### The cost of the product is made of:
- **material** costs (we have the BOM for this)
- **labour** costs (we have the timesheet for this)
- and finally the **overhead**

How much is this overhead? As we are trying to accommodate many different manufacturing processes, I believe the intention was to have:

##### Cost per Cycle -  sum of all overhead costs associated with running one _work center cycle_

##### Cost per Hour -  sum of all overhead costs associated with running the work center (excluding any costs covered in cost per cycle) divided by the number of hours the work center is operating.

Additionally we have two other costs PRE and POST which are fixed costs before and after the operations of the work center. 

It's up to the user to determine these as accurately as possible and use them accordingly to their manufacturing process. Some might chose only one of these costs, some all of them or maybe none.

Would this make sense to you?

anajuaristi:
@ddico It completely makes sense to me

@rafaelbn The misunderstanding on your point of view (and @pedrobaeza's
one) against our point of view is that in your case you see both field
related and you see the same cost. For us, they are both completely
different cost.

cost by hour is the cost of the machine running during several time
independently of how many products are produced
cost by cycle is the cost on this operation for each piece (cycle) produced
independently of how long it takes to produce (This is overhead)

Your vision could be correct but you are loosing a lot of use cases and you
don't need one of the fields.
In our vision both fields are needed and user is selecting to use one, the
other or both together. The generated costs ARE NOT duplicated costs are
different costs (in our vision). If we do like you say we are avoiding
users to use both fields, if we do like we say you are giving user the
option of using whatever they need.

The last point is that Standar Odoo creates 2 costs if you fill the
correspondent analytic account (it's not right and those account fields
should be made invisible on that point but ..) creating a cost line for
each line, they are showing the meaning of having those 2 cost fields there

My2cents

2016-01-25 21:50 GMT+01:00 ddico notifications@github.com:

> Hi @rafaelbn https://github.com/rafaelbn
> 
> Thank you for your review.
> 
> I'm concerned with what you have stated and the way you have calculated
> "cost per cycle" in the image above.
> 
> My PR is calculating the cost per cycle as:
> mrp_workcenter.costs_cycle multiplied with
> mrp_production_workcenter_line.cycle
> and is has nothing to do with "cost per hour".
> 
> I think before we can move on with this we should all agree on the
> definition of these two costs and why we have them.
> Odoo's definition:
> 
> _Cost per Hour_ - cost of work centre per hour.
> _Cost per Cycle_ - cost of work centre per cycle.
> 
> Very generic no? So, what does this mean?
> 
> In this module mrp_production_real_cost we are trying to give users an
> accurate calculation of the cost of the product.
> The cost of the product is made of:
> - _material_ costs (we have the BOM for this)
> - _labour_ costs (we have the timesheet for this)
> - and finally the _overhead_
> 
> How much is this overhead? As we are trying to accommodate many different
> manufacturing processes, I believe the intention was to have:
> Cost per Cycle - sum of all overhead costs associated with running one _work
> center cycle_ Cost per Hour - sum of all overhead costs associated with
> running the work center (excluding any costs covered in cost per cycle)
> divided by the number of hours the work center is operating.
> 
> Additionally we have two other costs PRE and POST which are fixed costs
> before and after the operations of the work center.
> 
> It's up to the user to determine these as accurately as possible and use
> them accordingly to their manufacturing process. Some might chose only one
> of these costs, some all of them or maybe none.
> 
> Would this make sense to you?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/manufacture/pull/83#issuecomment-174658767.

## 

CEO Avanzosc, S.L http://www.avanzosc.es : Office phone / Tfono oficina:
(+34) 943 02 69 02
Ana Juaristi Olalde http://www.anajuaristi.com/: Personal phone: 677 93
42 59. User/usuario skype: Avanzosc
www.openerpsite.com

_El contenido de esta comunicación y de toda su documentación anexa es
confidencial y se dirige exclusivamente a su destinatario. El uso no
autorizado de esta información está prohibido por la legislación vigente.
Si usted no es el destinatario le rogamos nos lo indique, no comunique su
contenido a terceros y proceda a su destrucción. Disculpe las molestias que
le haya ocasionado la recepción indebida de este e-mail. Sus datos figuran
en un fichero cuyo titular es Avanzosc, S.L., a quien usted puede dirigirse
para ejercer sus derechos de acceso, rectificación, cancelación y oposición
en Klara Donea 13, 20720, Azkoitia (Gipuzkoa), Tef. 943 02 69 02 -
*_administracion@avanzosc.
soporte@avanzosc.eses*

_Komunikazio honen edukia eta dokumentazio erantsia konfidentziala da eta
hartzaileak bakarrik jaso beharko luke. Indarrean dagoen legeriak debekatu
egiten du bertan eskainitako informazioa baimenik gabe erabiltzea.
Komunikazioa zuri iritsi bazaizu, baina zu ez bazara hartzailea, mesedez,
guri jakinarazi, eta jasotako informazioa ez inori jakinarazi eta suntsitu.
Barkatu okerreko email hau jasotzeak eragindako eragozpenak. Zure datuak
Avanzosc, S.L. enpresaren fitxategietan sartuta daude. Zure datuak atzitzea
eska dezakezu, bai eta, datuak zuzentzea, ezereztea eta tratamenduari aurka
egitea ere. Horretarako, enpresara jo dezakezu, helbide honetan: *_Klara
Donea 13*\* 20720, Azkoitia (Gipuzkoa), telefonoa: 943 02 69 02 - *
_administracion@avanzosc. soporte@avanzosc.eses_ _This message and all
documents attached to it are confidential and intended only for the person
or entity to which it is addressed. Any use of this information by
unauthorised persons is prohibited under current legislation. If you
received this message by error, please advise us, destroy it and refrain
from communicating its contents to third parties. We apologise for any
inconvenience receiving this email improperly may cause to you. Your
personal data are included in a file owned by Avanzosc, S.L. If you want to
exercise your rights of access, correction, erasure and objection you can
contact the Controller at *_Klara Donea 13*\* 20720, Azkoitia (Gipuzkoa), T:
943 02 69 02 – administracion@avanzosc. soporte@avanzosc.eses*

pedrobaeza:
OK, then let's explain very well in the UI this use case and maybe introduce another fields for making what @rafaelbn wants to do. What do you say, @ddico?

ddico:
I will make a proposal for explaining all this it the UI / readme.

@rafaelbn please create a PR for your use case.

Thank you

rafaelbn:
It is possible to set both fields, is this correct? is correct to set a cost per hour and also a cost per cycle?

rafaelbn:
It is possible to set both fields, is this correct? is correct to set a cost per hour and also a cost per cycle?

Depending on this maybe we should block one of them when the other is filled.

From: https://github.com/OCA/manufacture/pull/83#issuecomment-174329321

> Yeah, I removed it because it creates duplicated cost for the same concept it you fill both cycles and duration. It was discussed with you, Ana, saying that I give priority to cycles.

Is correct to duplicate cost?

We must think in user and not in our opinions.

@ddico this happens

> I'm concerned with what you have stated and the way you have calculated "cost per cycle" in the image above

is because a manager from a factory never user cost per cycle in his life.

From Odoo and we must be in the same line, first think to make it easy, user friendly and intuitive, second do it.

oihane:
Travis fails, PEP8 because of:

```
mrp_production_real_cost/models/mrp_production_workcenter_line.py:62:17: E122 continuation line missing indentation or outdented
mrp_production_real_cost/models/mrp_production_workcenter_line.py:63:17: E122 continuation line missing indentation or outdented
mrp_production_real_cost/models/mrp_production_workcenter_line.py:64:17: E122 continuation line missing indentation or outdented
mrp_production_real_cost/models/mrp_production_workcenter_line.py:66:25: E126 continuation line over-indented for hanging indent
mrp_production_real_cost/models/mrp_production_workcenter_line.py:67:25: E128 continuation line under-indented for visual indent
```

There are also 2 warnings in pylint check, I don't know if you have seen them.

```
************* Module mrp_production_real_cost.__openerp__
mrp_production_real_cost/__openerp__.py:7: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
mrp_production_real_cost/__openerp__.py:7: [C8102(manifest-required-key), ] Missing required key "images" in manifest file
```

@ddico you should also check the test, because know it creates more analytic lines.

ddico:
@oihane - I fixed the PEP8 issues.
Test are failing as more analytic entries are generated because `cost_cycle` is $0.05 in the demo data. The purpose of this PR is to generate an extra account analytic line record for the cycle cost. Test will have to be adjusted eventually.

@rafaelbn - there is always the option to keep `cost_cycle` zero and no analytic entry will be created.
How many users, after they have filled in an amount in the `cost_cycle` field, will think the calculation is wrong because is not adding up... probably they won't call that friendly nor intuitive. 

I'm not sure what else I can do or say about this (except maybe changing the test)... so you guys will have to decide if you want it or not.

rafaelbn:
OK. Agree on this finally, we will check user impact after some month running in customers.

Let's see why travis fails and going on with this PR

Thanks!

ddico:
I'm planing to add the image below to the readme of the `mrp_production_real_cost` module together with a description of the costs as follows:

mrp real cost

`Cost/cycle` = all overhead costs associated with running one work centre cycle (where a cycle is the `Number of Cycles` as calculated in the Work Order). 

`Cost/hr` = calculated as all overhead costs associated with running the work centre (excluding any costs covered in the `cost per cycle`) / the number of hours the work centre is operating.

`Pre-operation cost` = `Time before prod.` x cost of the `Pre-operation costing product`

`Post-operation cost` = `Time after prod.` x cost of the `Post-operation costing product`

It’s up to the user to determine overhead costs as accurately as possible and use all, some or none of them in order to reflect their manufacturing process.

```

Would this make sense to you @anajuaristi @pedrobaeza @rafaelbn @oihane ?
```

pedrobaeza:
Wow, that's an amazing explanation that you can include in the README.rst file!

rafaelbn:
clap your hands all together! :open_mouth: 

This make sense for me :+1: 

Thanks you @ddico !!!

cc @JavierIniesta @JMODaERP @antespi

anajuaristi:
Great graphic but just a point...
Hour account and cycle account shouldn't be empty?  I don't like to use
those account fields because they are generic by routing so all cost from
all MO's are mixed in a single account. We are setting a different account
per MO so I think those account fields are not needed anymore.
Do you think we should make them invisible or advice somehow to user on
readme? Or they are rigth as they are?

ddico:
I agree with you the analytic accounts `Hour Account` and `Cycle Account` are not in use. Odoo traditionally will save them into the `ref` field in `account_analytic_line`, but that detail was lost till one of my PRs I made this week when i've added them back into the `ref` field. I think they could be useful if someone wants to summarize some data based on them. I don't think people will be confused now if they see them in the ref field... what do you think?

anajuaristi:
In my point of view I would completely erase them or at less I would make
them invisible. About having them as summarize field I don't know if this
could be usefull. In case of having the need to summarize would be much
better to use analytic journals since they are designed on that purpose.
IMHO.

ddico:
@anajuaristi - I don't have a strong feeling about that so I removed them from the picture 

@pedrobaeza @rafaelbn @oihane - I've updated the readme and test to reflect the cycle count, please double check.

Thank you

JMODaERP:
Go I have fallen down in the suitable place. 

This debate is very interesting, and I believe that undoubtedly Rafael is right. If  a center has a capacity of 1000 units (capacity_per_cycle del objeto mrp.workcenter) for hour (time_cycle del mrp.workcenter) and according to the product center of production (product_id del objeto mrp.workcenter )  its cost is 30 $ per hour it remains clearly clear that the cost of a cycle (number of operations can be realized in a period of time) is 30 $. 

Then the cost cycle is a variable dependent on the cost hour and of the hours cycle and how dependent variable must calculate by means of the values of others two.

 Now then, for what we are looking is the cost machine of the center for a cycle of production or that is the same, the cost machine which we incur when we make 1000 units. Therefore as soon as the work order was finished, the system should allow us to change the number of imputed hours, or the number of cycles, thing that now although it allows it, then he does not consider the changes in the quantity of the analytics Journal item not in the calculation of the quantity del analytics Journal item neather Total Real Costy.

In short: that the cost cycle is a studied fact and that the number of hours affects to the real cost when we modify it at the end of the order.

JMODaERP:
I tested this runbot and:

I have modified to hand Machine Up time, as a real  case, the user need modify the WO and:

First: 
1 2016-01-30 21_15_53-work orders - horas de trabajo ingles

2 2016-01-30 20_56_04-costs revenues - analitics
Second:

4 2016-01-30 21_26_59-manufacturing orders - total real cost

Third: 

What is this?.

3 2016-01-30 20_52_56-ordenes de trabajo -extra information

ddico:
Hi @JMODaERP,

I think your definition of cost per cycle is different than the one we had in this thread. 

If you want to change the time, you need to do it before you `Finish` a work order. Changing the time after you have finished the work order will not recalculate the analytic lines, regardless if it's cycle or hourly cost. I think you have modified the machine up time after the work order was finished.

If you believe the hourly cost calculation is not correct, please keep in mind that this module does not use `Number of Hours` - instead this module is using `Machine up time` in order to calculate the cost. As you can see the cost is calculated [here](https://github.com/OCA/manufacture/blob/8.0/mrp_production_real_cost/models/mrp_production_workcenter_line.py#L36) as:
price = -(workcenter.costs_hour \* operation_line.uptime)

And if I'm not mistaken in the example you have provided you are referring to hourly cost while my PR is about the missing cycle cost.

pedrobaeza:
@ddico, can you rebase your PR?

ddico:
@pedrobaeza - done.

pedrobaeza:
:+1:

oihane:
LGTM :+1:

Issue #81, opened 2016-01-24
8.0 [BUG] MRP Total Real Cost


JMODaERP:
The costs charged (direct) in an order of production standard, whatever the industry sector, are calculated based on the cost of raw materials and auxiliary materials, plus the cost of operators, plus the cost of machinery. Also many companies include preparing machines (CAM,CAD,setting dies, tooling etc ) and often include cleaning the room or production facility , like pre and post tasks.

Currently, the calculations performed Odoo OCA , the costs of operators that are being 33% of the total real costs, and allocation of hours which are ignored have created the " Operators time" tab and the employee has been included HR Settings and product_id of hr_employee . 

Maybe it's an oversight (BUG) or a special way to calculate the total real costs.

If you see fit to do this repair VS improving the resolution love me , otherwise I would open a debate on the subject.

Thank you.

For example : In a production

Current Behavior.

![1 2016-01-24 12_34_48-2016-01-24 12_06_18-costs revenues - odoo-operarios 1-ingles](https://cloud.githubusercontent.com/assets/15608043/12536403/45811b70-c2a4-11e5-803b-da1a8e412a5b.jpg)

```
                                                            Visual 1
```

![2 2016-01-24 12_07_39-costs revenues - odoo-operarios 2](https://cloud.githubusercontent.com/assets/15608043/12536405/5e4b2998-c2a4-11e5-950a-6a89f924d33d.jpg)

```
                                                            Visual 2
```

![3 2016-01-24 12_08_06-costs revenues - odoo-operarios 3](https://cloud.githubusercontent.com/assets/15608043/12536408/764b6ecc-c2a4-11e5-9a92-ab9a54ea0b8c.jpg)

```
                                                             Visual 3
```

![4 2016-01-24 12_52_54-2016-01-24 12_09_02-manufacturing orders - odoo-mo-total real cost](https://cloud.githubusercontent.com/assets/15608043/12536415/9a4ceb16-c2a4-11e5-9c03-e2ba6e73a337.jpg)

```
                                                              Visual 4
```

Expectde Behavior.

![5 2016-01-24 14_04_43-sin titulo - paint](https://cloud.githubusercontent.com/assets/15608043/12536420/c05353e0-c2a4-11e5-9ffe-31ec50fec3e7.jpg)

```
                                                               Visual 5
```

pedrobaeza:
Please follow expected schema for reporting bugs:

module: mrp_production_real_cost
version: 8.0

**Steps to reproduce**
[...]

PR #85, opened 2016-01-25
[FIX] mrp_production_real_cost - include timesheet in cost

ddico:
In the readme file of this module it states that:

_...thanks to project_timesheet modules, users time is also translated to costs in the linked analytic account._

My findings are that time sheet entries created for the correct analytic account (MO/Project) are **not** included in the cost calculation as the timesheet entry is missing the `mrp_production_id` value. 

A little more technical: [_compute_real_cost](https://github.com/OCA/manufacture/blob/8.0/mrp_production_real_cost/models/mrp_production.py#L15) method will only sum account_analytic_line where mrp_production_id was set (as per the field definition [here](https://github.com/OCA/manufacture/blob/8.0/mrp_production_real_cost/models/mrp_production.py#L23))

This PR will add mrp_production_id into the timesheet and now cost calculation will include the timesheet amounts.

### DIFF
--- a/mrp_production_real_cost/README.rst
+++ b/mrp_production_real_cost/README.rst
@@ -26,14 +26,15 @@ https://github.com/OCA/product-variant
 Usage
 =====
 
-Operating with a manufacture order, analytic entries adding costs will be
+Processing a manufacture order, analytic entries adding costs will be
 created when:
 
 * Some raw material is consumed.
 * A work order is finished or paused.
+* Also, together with *project_timesheet* module, users time is also translated
+  to costs in the linked analytic account.
 
-Also, thanks to *project_timesheet* modules, users time is also translated to
-costs in the linked analytic account.
+The sum of all these analytic entries is the real cost.
 
 
 
--- a/mrp_project/models/hr_analytic_timesheet.py
+++ b/mrp_project/models/hr_analytic_timesheet.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Daniel Dico <dd@oerp.ca>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, api
+
+
+class HrAnalyticTimesheet(models.Model):
+    _inherit = "hr.analytic.timesheet"
+
+    @api.model
+    def create(self, vals):
+        production = self._context.get('production', False)
+        vals['mrp_production_id'] = vals.get(
+            'mrp_production_id', False) or production and production.id
+        return super(HrAnalyticTimesheet, self).create(vals)
--- a/mrp_project/models/mrp_production.py
+++ b/mrp_project/models/mrp_production.py
@@ -91,3 +91,8 @@ class MrpProductionWorkcenterLine(models.Model):
     work_ids = fields.One2many(
         comodel_name="project.task.work", inverse_name="workorder",
         string="Task works")
+
+    @api.multi
+    def write(self, vals, update=True):
+        return super(MrpProductionWorkcenterLine, self.with_context(
+            production=self.production_id)).write(vals)
--- a/mrp_project/__openerp__.py
+++ b/mrp_project/__openerp__.py
@@ -12,6 +12,7 @@
         "project",
     ],
     'license': 'AGPL-3',
+    "images": [],
     "author": "OdooMRP team,"
               "AvanzOSC,"
               "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
@@ -22,7 +23,8 @@
         "views/mrp_production_view.xml",
         "views/project_project_view.xml",
         "views/account_analytic_line_view.xml",
-        "views/project_task_view.xml"
+        "views/project_task_view.xml",
+        "views/hr_analytic_timesheet.xml"
     ],
     'installable': True,
     'auto_install': False,
--- a/mrp_project/models/__init__.py
+++ b/mrp_project/models/__init__.py
@@ -8,3 +8,4 @@ from . import account_analytic_line
 from . import project_project
 from . import project_task
 from . import project_task_work
+from . import hr_analytic_timesheet
--- a/mrp_project/models/project_task.py
+++ b/mrp_project/models/project_task.py
@@ -29,3 +29,8 @@ class ProjectTask(models.Model):
                     (task.id, "[%s] %s" % (task.user_id.name, task.name)))
             return res
         return super(ProjectTask, self).name_get()
+
+    @api.multi
+    def write(self, vals):
+        return super(ProjectTask, self.with_context(
+            production=self.mrp_production_id)).write(vals)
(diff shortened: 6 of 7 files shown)


ddico:
@JMODaERP - This PR will probably help you with #81

pedrobaeza:
I don't like too much this method because you can have multiple production to the same analytic account. We should provide the production id via a dictionary context for filling the correct data.

JMODaERP:
This PR will probably help to # 81 . Yes indeed, this functionally solves PR PR # 81 . I tested on a testing framework and has added  the cost of the actual total cost operators .
Then, functionally it solves the # 81, so I approve .

Thank you :+1:

rafaelbn:
Is runbot related?
Tested locally :+1:

ddico:
Is coveralls an issue here?

pedrobaeza:
Well, the coverage fall down because you're not testing the new "feature". Anyway, I'm afraid for the other concern pointed in https://github.com/OCA/manufacture/pull/85#issuecomment-174457892

ddico:
Passing a dictionary while recording the "operators time" in a work order it's indeed a possibility, but it's not going to work if users use "HR/Timesheet Activities" for example.

In this case we could show all MOs related to the Account Analytic selected... but, is the user going to know the MO info? Most likely a user will know the task he was assigned to, so we need to show the task to the user, and use the task to get the MO#? what do you think? I believe some of the related fields were removed lately, does this takes us back there?

Do you see a different alternative?

pedrobaeza:
In the case users use "HR/Timesheet Activities", they should select directly in a many2one the involved MO/WO.

rafaelbn:
Do you know why travis fails?

ddico:
Hi @pedrobaeza, I've implemented the changes we have agreed above. While testing I started to think that we are creating some limitations forcing the users to use MO/WO in timesheet activities. They should be able to select the task instead, in my opinion.

For example we have one MO with 2 WO. One of the work orders has 1 operator and the other one has 2 operators. In this scenario we will end up with a total of 4 tasks (1 Generic and 3 for the WO)
- By going with MO/WO in Timesheet we are going to present the user just the 2 WO.
- By going with TASKS in Timesheet we show user the possibility to select one of the 4 tasks created. This way we can track time for the generic task and also the user can chose the task he was assigned to.

What do you think? 

Also @anajuaristi @rafaelbn @JMODaERP - do you have a preference?

oihane:
@ddico you have a PEP8 error:

```
mrp_production_real_cost/models/hr_analytic_timesheet.py:14:17: E126 continuation line over-indented for hanging indent
```

ddico:
@oihane - thank you for your feedback. 

@pedrobaeza - I've finally pushed the changes into mrp_project as it seems a better fit. I made the following changes:
- production id via dictionary context for filling the correct data
- "HR/Timesheet Activities" with many2one for MO

As per the original issue, timesheet entries will be now included in the cost calculation and Users/Operators can enter the time/labour using:
- Work Orders
- Project Tasks
- Timesheet Activities

Please note that using the Timesheet Activities is not the preferred way to capture the labour as `project_task_work` records will not be created. As a result this time will not be visible in Workorders or Project Tasks but it will be included in the real cost calculation (this might confuse some users hence is not the preferred way)

Hope this helps

mrp_project/models/hr_analytic_timesheet.py
```diff
@@ -0,0 +1,16 @@
...
+from openerp import models, api
+
+
+class HrAnalyticTimesheet(models.Model):
+    _inherit = "hr.analytic.timesheet"
+
+    @api.model
+    def create(self, vals):
+        vals['mrp_production_id'] = 'mrp_production_id' in vals and \
+                                    vals['mrp_production_id'] or \
```
oihane: if instead of using `vals['mrp_production_id']` you use `vals.get('mrp_production_id, False)` you won't require to check `'mrp_production_id' in vals`

mrp_production_real_cost/README.rst
```diff
@@ -26,14 +26,15 @@ https://github.com/OCA/product-variant
...
 =====
 
-Operating with a manufacture order, analytic entries adding costs will be
+Processing a manufacture order, analytic entries adding costs will be
 created when:
 
 * Some raw material is consumed.
 * A work order is finished or paused.
+* Also, together with *project_timesheet* module, users time is also translated
+to costs in the linked analytic account.
```
pedrobaeza: This comment should go now on the other module (and you should put 2 spaces indentation here for a proper display of the RST).

mrp_production_real_cost/README.rst
```diff
@@ -26,14 +26,15 @@ https://github.com/OCA/product-variant
...
 
-Also, thanks to *project_timesheet* modules, users time is also translated to
```
pedrobaeza, line 35: This text is still valuable I think.

mrp_project/models/mrp_production.py
```diff
@@ -91,3 +91,8 @@ class MrpProductionWorkcenterLine(models.Model):
     work_ids = fields.One2many(
         comodel_name="project.task.work", inverse_name="workorder",
         string="Task works")
+
+    @api.multi
+    def write(self, vals, update=True):
+        return super(MrpProductionWorkcenterLine, self.with_context(
+            production_id=self.production_id.id)).write(vals)
```
pedrobaeza: It's better practice pass the full recordset instead only the ID, in case you need any field for not making the browse again.

mrp_production_real_cost/README.rst
ddico, line 35: It's a few lines above...

pedrobaeza, line 35: Sorry, very bad review mine I should say... Please make the 2 spaces indentation thing.

pedrobaeza, line 35: Well, reading again, maybe my first comment is valid, because this refers to the _mrp_project_ module in general. What do you think?

mrp_production_real_cost/README.rst
ddico: When a user looks at the description of the `mrp_production_real_cost` module, it will be useful to see what is included in the cost. I think is more useful here. 
I'll add proper indentation.

pedrobaeza: OK, let's continue having here. I have to say that I love this collaboration we are having here in these modules!

ddico:
done

mrp_production_real_cost/README.rst
ddico, line 35: I chose mp_project because the relationship between `account_analytic_line` and `mrp_production` is created there. I was thinking to be close to that with these changes. 

But the costing issue I found was related to `mrp_production_real_cost` ... and that's where users are going to read the module description

ddico:
It finally passed the checks... I'll be the first to vote for the increase of PEP8 Maximum Line Length :-)

pedrobaeza:
:clap::clap: 

Merging!

PR #89, opened 2016-02-04
[FIX][REF] mrp_operations_extension: Fix WO raw stock move relation

zamberjo:
When a **MO** has subproducts, the resulting **WO** adds - wrongly - the
stock moves as raw materials, even if they aren't.

This fix, ensures that all the stock moves must be raw materials and
must belong to that **WO**.

In addition, I've refactored `_compute_is_material_ready` method to
use these calculated stock moves, and I have also moved the `_compute*`
methods position to follow Odoo's guidelines

### DIFF
--- a/mrp_operations_extension/__openerp__.py
+++ b/mrp_operations_extension/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     "name": "Manufacturing Operations Extension",
-    "version": "8.0.3.0.0",
+    "version": "8.0.2.1.0",
     "category": "Manufacturing",
     "license": "AGPL-3",
     "author": "OdooMRP team, "
--- a/mrp_operations_extension/__openerp__.py
+++ b/mrp_operations_extension/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     "name": "Manufacturing Operations Extension",
-    "version": "8.0.2.0.0",
+    "version": "8.0.3.0.0",
     "category": "Manufacturing",
     "license": "AGPL-3",
     "author": "OdooMRP team, "
--- a/mrp_operations_extension/models/mrp_production.py
+++ b/mrp_operations_extension/models/mrp_production.py
@@ -65,8 +65,8 @@ class MrpProductionWorkcenterLine(models.Model):
     time_start = fields.Float(string="Time Start")
     time_stop = fields.Float(string="Time Stop")
     move_lines = fields.Many2many(
-        'stock.move', compute="_compute_move_lines", string='Moves',
-        store=False)
+        comodel_name='stock.move', compute="_compute_move_lines",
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Are you talking about module _mrp_byproduct_?

zamberjo:
Yes, I've seen the error using the module `mrp_byproduct`.

pedrobaeza:
Please bump the version number as you change a field definition? Can you point me to the guideline where talks about the method order (just curious about that)?

zamberjo:
On [Symbol section](https://www.odoo.com/documentation/8.0/reference/guidelines.html#symbols) -> **In a Model attribute order should be**.

What format I use to increase the version?
1. 8.0.2.0.1
2. 8.0.2.1.0
3. 8.0.3.0.0
   ¿?

pedrobaeza:
Now I give you another link as payback :wink: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

zamberjo:
Cool! thx!

mrp_operations_extension/__openerp__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     "name": "Manufacturing Operations Extension",
-    "version": "8.0.2.0.0",
+    "version": "8.0.3.0.0",
```
pedrobaeza: In this case it should be 8.0.2.1.0, as is a non-breaking feature but requires to update the module.

pedrobaeza:
:+1:

mrp_operations_extension/models/mrp_production.py
```diff
@@ -81,8 +64,9 @@ def _compute_possible_workcenters(self):
     do_production = fields.Boolean(string='Produce here')
     time_start = fields.Float(string="Time Start")
     time_stop = fields.Float(string="Time Stop")
-    move_lines = fields.One2many(
-        comodel_name='stock.move', inverse_name='work_order', string='Moves')
+    move_lines = fields.Many2many(
+        'stock.move', compute="_compute_move_lines", string='Moves',
```
oihane: @zamberjo Why did you remove `comodel_name` text, although it is not required, it was there.

zamberjo: The parameter name can be omitted as we are declaring the parameters in the same order as the declaration, it isn't necessary at all... it saves line width

pedrobaeza: I prefer not to remove it indeed, because tomorrow, the signature can change. Using keyword arguments prevent the incompatibility in that case.

LoisRForgeFlow:
Closing as this is stale and it has conflicts not resolved. Please reopen if you think still is relevant.

PR #91, opened 2016-02-13
8.0 mrp production real cost uptime update

pedrobaeza:
Up to now, you have to rely only on play/pause/stop controls, and the time passed between them for having real costs reflected, but there are manufacturing processes that reflects their times not directly, but afterwards introducing the summaries. This commit allows to reflect the changes on costs when you edit up times.

This PR also fixes an incorrect dependency that is not needed (you can have it installed to get the extra feature, but it's not a requirement).

### DIFF
--- a/mrp_production_real_cost/__openerp__.py
+++ b/mrp_production_real_cost/__openerp__.py
@@ -13,7 +13,6 @@
         "mrp_operations_extension",
         "mrp_operations_time_control",
         "stock_account",
-        "product_variant_cost_price",
     ],
     'license': 'AGPL-3',
     "author": "OdooMRP team, "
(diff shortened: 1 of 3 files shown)


pedrobaeza:
cc @rafaelbn @antespi @ddico

ddico:
Hi @pedrobaeza 

This is a great idea. I tested and works as expected. :thumbsup:

I see a little performance issue as `_create_analytic_line` is called twice at every `action_done`. I think the problem is when `_write_end_date_operation_line` is called twice in `action_done` (mrp_production_workcenter_line.py) and also in the `action_done` (operation_time.py in mp_operations_time_control)

JMODaERP:
Thanks Peter for fixing fast the BUG you conveyed on Friday as well hours charged manually entered once the work order. OK.
Postscript: I guess it's not on this PR the average cost price of the product it is not right when the Cost Method is average? :+1:

pedrobaeza:
@ddico, you're right! I have removed one of the duplicated

rafaelbn:
Thanks! :+1:

PR #94, opened 2016-02-17
[8.0][IMP][mrp_project] Set workorder in analytic.line related with a task from a work order

antespi:
This PR improves @ddico work added here https://github.com/OCA/manufacture/commit/b82abb6ecea258586e34802d48e9a700d935118b
Setting also work order to `account.analytic.line` when creating task work from a task related with a work order

### DIFF
--- a/mrp_project/__openerp__.py
+++ b/mrp_project/__openerp__.py
@@ -6,7 +6,7 @@
 {
     "name": "MRP Project Link",
     "summary": "Link production with projects",
-    "version": "8.0.1.1.1",
+    "version": "8.0.1.2.0",
     "depends": [
         "mrp_analytic",
         "project",
--- a/mrp_project/__openerp__.py
+++ b/mrp_project/__openerp__.py
@@ -6,10 +6,11 @@
 {
     "name": "MRP Project Link",
     "summary": "Link production with projects",
-    "version": "8.0.1.1.0",
+    "version": "8.0.1.1.1",
     "depends": [
         "mrp_analytic",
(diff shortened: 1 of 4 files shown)


mrp_project/__openerp__.py
```diff
@@ -6,10 +6,11 @@
 {
     "name": "MRP Project Link",
     "summary": "Link production with projects",
-    "version": "8.0.1.1.0",
+    "version": "8.0.1.1.1",
     "depends": [
         "mrp_analytic",
         "project",
+        "project_timesheet",
```
pedrobaeza, line 13: Why this new dependency? This breaks current installations.

pedrobaeza, line 13: A new addon _mrp_project_timesheet_?

antespi, line 13: This dependency is necessary for enabling inheritance in this method: https://github.com/OCA/manufacture/pull/94/files#diff-8655ff19eb2fb4b321a2742bdcf95e8eR33 `_create_analytic_entries`

pedrobaeza, line 13: OK, go on with it. You're not going to find any installation that enables real costs without it, but please change version to 8.0.1.2.0

antespi, line 13: Done, thanks

pedrobaeza:
:+1:

rafaelbn:
:+1:

sergio-teruel:
:+1:

PR #98, opened 2016-02-21
[8.0][FIX][mrp_production_real_cost] Support several stock moves for the same product when calculating new price

antespi:
In a manufacture order you can produce a different product qty that initially provided. In that case, two stock moves are created.

Current `product_price_update_production_done` method only supports one stock move per product. This PR adds support for this use case.

### DIFF
--- a/mrp_production_real_cost/models/stock_move.py
+++ b/mrp_production_real_cost/models/stock_move.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # © 2014-2015 Avanzosc
 # © 2014-2015 Pedro M. Baeza
+# © 2016 Antiun Ingenieria S.L. - Antonio Espinosa
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from openerp import api, models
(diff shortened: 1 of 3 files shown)


mrp_production_real_cost/models/stock_move.py
```diff
@@ -37,30 +38,58 @@ def action_done(self):
...
+                product_data['available'] -= move.product_qty
+                product_data['moved'] += move.product_qty
+                tmpl_data['available'] -= move.product_qty
+                tmpl_data['moved'] += move.product_qty
+            products[product.id] = product_data
+            templates[tmpl.id] = tmpl_data
+        for product_id, product_data in products.iteritems():
+            new_price = self._new_average_price(product_data),
+            product_data['product'].sudo().standard_price = new_price
+        for tmpl_id, tmpl_data in templates.iteritems():
```
pedrobaeza: Don't update template prices. Module _product_variant_cost_price_ should do it (when template.product_variant_ids == 1).

mrp_production_real_cost/models/stock_move.py
```diff
@@ -37,30 +38,58 @@ def action_done(self):
...
+            tmpl = product.product_tmpl_id
+            product_data = products.get(product.id, False) or {}
+            if not product_data:
+                product_data['product'] = product
+                product_data['available'] = product.qty_available
+                product_data['price'] = product.standard_price
+                product_data['moved'] = 0.0
+                product_data['cost'] = move.production_id.real_cost
+            tmpl_data = templates.get(tmpl.id, False) or {}
+            if not tmpl_data:
```
pedrobaeza: Remove this part

mrp_production_real_cost/models/stock_move.py
antespi: But remember that this addon is not depending to _product_variant_cost_price_ anymore

pedrobaeza: And that's the key: if you don't install it, variant cost = template cost. Writing to product is the same as writing to template. If you have it installed, there are 2 cases:
- There are more than 1 variants: the cost price is only written to the variant.
- There are only 1 variant: the cost price in written on the variant and then written upwards to the template.

In any case, you only need to write to `product.product`. Inheritance and the other module do the rest.

antespi: Ok, cool, this simplify the code a lot

antespi:
Done. Squash and rebase too

pedrobaeza:
:+1:

pedrobaeza:
Please bump version number

rafaelbn:
Thanks! :+1:

antespi:
@pedrobaeza Fixes done, thanks

PR #99, opened 2016-02-23
[8.0][FIX][mrp_operations_time_control] Set end date only when not defined

antespi:
When pressing "Finished" button in a Work Order the last "Machine up time" line is overwritten at "End date" field with the today date, even if "End date" is assigned or not.

This PR adds a condition for setting "End date" only when not set.


pedrobaeza:
:+1: 

Simple patch. I'll fast-track when CIs finish

ddico:
:+1:

JMODaERP:
I´ve tested :+1:

Issue #102, opened 2016-03-01
[8.0][mrp_operations_extension] 'action_start_working' ignores the MO's workflow


zamberjo:
When you start a WO, it sets the MO state ignoring the MO workflow behaviour as you can see in this PR: (https://github.com/OCA/manufacture/commit/5e6ced1cd9b18855ce3f7ed2f1860d9f3ad4c1ae#diff-5ddfc0f2cd6e2a6524362393d297c9deR158)

This code isn't necessary at all  because the **modify_production_order_state** method (**mrp_operations**) changes the  state to **in_production** through the workflow: see https://github.com/odoo/odoo/blob/8.0/addons/mrp_operations/mrp_operations.py#L118

zamberjo:
We've fixed it in our fork https://github.com/diagramsoftware/manufacture/commit/01c75aa18ad7a6e1ffbb5c02f30513746e84ee85, we can do a PR if you agree.

pedrobaeza:
OK, make the PR and we discuss about it there (although I see at first sight as correct).

LoisRForgeFlow:
Related PR has been merged.

PR #103, opened 2016-03-01
[8.0] mrp_operations_extension: Remove write MO's state on 'action_start_working'

zamberjo:
Modify state on WO it isn't necessary at all because the **mrp_operations**
changes the state to **in_production** through the workflow.

Issue #102

### DIFF
--- a/mrp_operations_extension/models/mrp_production.py
+++ b/mrp_operations_extension/models/mrp_production.py
@@ -133,6 +133,4 @@ class MrpProductionWorkcenterLine(models.Model):
                     ['assigned', 'cancel', 'done']):
                 raise exceptions.Warning(
                     _("Missing materials to start the production"))
-            if workorder.production_id.state in ('confirmed', 'ready'):
-                workorder.production_id.state = 'in_production'
         return super(MrpProductionWorkcenterLine, self).action_start_working()


pedrobaeza:
@Daniel-CA, @anajuaristi, @agaldona do you remember why this condition was introduced?

anajuaristi:
I should have to see it better but I thik it's related with materials availability.
Odoo does not allow to start MO unless ALL raw materials are available. In odoomrp we are controlling raw materials by operations so having available raw materials for first operation, if user starts this one, system should start the MO independently of Mo workflow conditions.
There is several industries were this functionality is critical and I would not aprove this PR if it means to loose the functionality I'm mentioning.

pedrobaeza:
Wow, that's critical indeed. @anajuaristi, next time, this kind of things can be documented as comments, so that when we came to them, we have clear the intention of that piece of code.

zamberjo:
Manually modifying the state of the MO to **in_production** does not trigger the workflow signal to change to the next activity. Due to this, the MO is still treated as if it were on the **ready** workflow activity instead of **in_production**.

One possible solution will be to replace the removed lines with `...signal_workflow(...'button_produce')`.

Daniel-CA:
The proposed change by @zamberjo is right. In current code we are bypassing the `mrp.production.basic` workflow, but there is no need to add any code. When the WO starts `action_start_working` executes, and it calls to `modify_production_order_state` in `mrp_operations` that it is responsible to change MO state.

So :+1:

pedrobaeza:
OK then.

I merge.

Daniel-CA:
Ok, I have seen the problem and why we add those lines. It was because the bug of the force production, when we start the WO if the MO is in 'confirmed' state it forces the production for all moves.

pedrobaeza:
OK, I revert then the PR.

PR #104, opened 2016-03-07
first commit product versioning - wip

sergiocorato:
mostly cloned from mrp_bom_version


pedrobaeza:
This should go in OCA/product-attribute

PR #108, opened 2016-03-11
[IMP] mrp_operations_extension: configurable cycle compute

agaldona:
Added new config parameter "Calc Cycles by BoM Quantity".
If it is activated (Default option), the system will calculate the number of cycles like before.
If it is deactivated, the system will calculate the number of cycles without taking into account BoM product quantity.

### DIFF
--- a/mrp_operations_extension/README.rst
+++ b/mrp_operations_extension/README.rst
@@ -31,6 +31,8 @@ Configuration
 
 Go to *Settings > Manufacturing* and activate "Manage routings and work orders"
 for handling the features of this module.
+Go to *Settings > Manufacturing* and deactivate "Compute Cycles by BoM Quantity"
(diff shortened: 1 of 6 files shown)


pedrobaeza:
I prefer to be deactivated by default, but :+1:

anajuaristi:
@pedrobaeza.. deactivated? So... the way it was before? or the way is now in OCA (activated)?
Now, you made me doubt. 
From my side both are correct and depends on customer needs so... let the one you consider the best.
Could we merge?

agaldona:
Deactivated is the new mode, computes cycles without BoM quantity. I set activated as default to keep the above formula, but maybe the check concept is not correct.

gemmabr:
tested.  OK!  :+1:

pedrobaeza:
:+1:

PR #112, opened 2016-03-18
[8.0][FIX] quality_control_mrp: ignore cancelled moves when creating inspections

maurochip:
When an inspection was created, it wasn't ignoring the cancelled moves.

### DIFF
--- a/quality_control_mrp/models/mrp_production.py
+++ b/quality_control_mrp/models/mrp_production.py
@@ -29,7 +29,8 @@ class MrpProduction(models.Model):
         if production_mode == 'consume_produce':
             inspection_model = self.env['qc.inspection']
             production = self.browse(production_id)
-            for move in production.move_created_ids2:
+            for move in production.move_created_ids2.filtered(
+                    lambda r: r.state == 'done'):
                 qc_trigger = self.env.ref('quality_control_mrp.qc_trigger_mrp')
                 trigger_lines = set()
                 for model in ['qc.trigger.product_category_line',


quality_control_mrp/models/mrp_production.py
```diff
@@ -30,6 +30,8 @@ def action_produce(self, production_id, production_qty, production_mode,
             inspection_model = self.env['qc.inspection']
             production = self.browse(production_id)
             for move in production.move_created_ids2:
+                if move.state != 'done':
```
pedrobaeza: Better to put:

```
for move in production.move_created_ids2.filtered(
        lambda x: x.state == 'done'):
   ...
```

maurochip: With the current way you can avoid iterating twice the moves list.

pedrobaeza: Filtered acts adding a condition on the select clause, so no penalty in performance, and it's more readable

pedrobaeza: Even more, your option can have performance penalty because of the computation of computed fields on the discarded records.

pedrobaeza:
Please squash this little change in only 1 commit, not 2.

Thanks.

pedrobaeza:
Thanks :+1:

cyrilgdn:
:+1:  (but it would be perfect with a test :-))

@pedrobaeza I don't understand how qc.trigger works. Is there an explanation somewhere?

pedrobaeza:
qc.trigger is an abstract model. It's inherited in product template, product variant and product category, and it allows to set tests to create inspections at certain events (manufacture, picking type...). Nothing more than that.

pedrobaeza:
@maurochip, have you signed the CLA?

maurochip:
Here you can see that is signed https://github.com/OCA/stock-logistics-workflow/pull/185#issuecomment-175454862

pedrobaeza:
OK, thanks, merging

PR #113, opened 2016-03-24
[ADD] new module mrp_bom_dismantling

cyrilgdn:
Depends on https://github.com/OCA/stock-logistics-warehouse/pull/148 (stock_available_mrp V9 migration).

This module add the ability to create a dismantling BOM by reversing a BOM.

### DIFF
--- a/mrp_bom_dismantling/README.rst
+++ b/mrp_bom_dismantling/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+BOM Dismantling
+===============
+
+This module adds the ability to create a dismantling BOM by reversing a BOM.
+
+
+Usage
+=====
+
+
+* On BOM form view, click on "Create dismantling BOM" button and it will reverse your BOM.
+* In Manufacturing -> Products, there is a new menu "Dismantling".
+* In dismantling tree view, you can search by dismantled product.
+* On BOM form view, there is a new button "Create Manufacturing Order".
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/129/9.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+manufacture/issues/new?body=module:%20
+mrp_bom_dismantling%0Aversion:%20
+9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Camptocamp - Cyril Gaudin <cyril.gaudin@camptocamp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/mrp_bom_dismantling/__openerp__.py
+++ b/mrp_bom_dismantling/__openerp__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2016 Cyril Gaudin (Camptocamp)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "BOM Dismantling",
+    "summary": "Ability to create a dismantling BOM by reversing a BOM.",
+    "version": "9.0.1.0.0",
+    "category": "Manufacturing",
+    "website": "http://www.camptocamp.com/",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        'mrp_byproduct',
+        "stock_available_mrp",
+    ],
+    "data": [
+        "views/mrp_bom.xml",
+        "views/product_template.xml",
+    ],
+}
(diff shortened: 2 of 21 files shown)


anajuaristi:
As I tried to explain in twitter you could do exactly the same without needing to create reverse bom and not using subproducts

Just needed a check in bom to mark "inverse" +
Include same check on Mo and inizialize value from bom. Allow to check or uncheck Mo when needed.

Modify Mo confirm button. If inverse it's checked, stock  move will be created having inverse origin and destiny location.
That is.. in normal Mo consumption moves are created from stock to production and production move is from production to stock.
Inverse MO will create comsumption moves from production to stock and production move from stock to production.
That's all.
My2cents
No more code lines needed to make dissasembly.

elicoidal:
I tend to agree with Ana here: it would be much simpler to reverse the method at BoM level and it should work for simple cases.
Nevertheless I can see that this would not solve many cases where the dismantling is not symmetric to the original assembly:
- routing is different for dismantling
- sub-products are different from the origin (eg: one piece made of 2 cannot be broken apart etc.)

So all in all I am supporting this module (not yet reviewed)

mrp_bom_dismantling/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+BOM Dismantling
+===============
+
+This module add the ability to create a dismantling BOM by reversing a BOM.
```
elicoidal: This module adds

mrp_bom_dismantling/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2016 Cyril Gaudin (Camptocamp)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "BOM Dismantling",
+    "summary": "Manage BOM dismantling",
```
elicoidal: Ability to create a dismantling BOM by reversing a BOM.

mrp_bom_dismantling/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+    "version": "9.0.1.0.0",
+    "category": "Manufacturing",
+    "website": "https://odoo-community.org/",
```
elicoidal: camptocamp website no?

mrp_bom_dismantling/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf8 -*-
```
elicoidal: copyright (several)

mrp_bom_dismantling/models/mrp_bom.py
```diff
@@ -0,0 +1,137 @@
...
+    def _check_bom_validity(self, check_dismantling=False):
+        """ Ensure this BoM can be use for creating a dismantling BoM
+        or a manufacturing order.
+
+        :type check_dismantling: bool
+        :raise exceptions.UserError: If this BoM is not valid.
+        """
+        warning = None
+        if check_dismantling and self.dismantling:
+            warning = 'This BoM is already a dismantling Bom'
```
elicoidal: add a period

mrp_bom_dismantling/models/mrp_bom.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if not len(self.bom_line_ids):
+            warning = 'This BoM does not have components.'
+
+        if not self.product_id \
+                and len(self.product_tmpl_id.product_variant_ids) > 1:
+            warning = 'This product has several variants, ' \
```
elicoidal: This product has several variants:

elicoidal:
LGTM :+1: 
Small details.
Please have a look into runbot (maybe not related)

elicoidal:
runbot: dependency problem: 'mrp_bom_dismantling' that depends on module 'stock_available_mrp'

anajuaristi:
Sorry Eric but I don't understand your point.
As a product can have as many boms as you want, you can define different
routings for assembly and dissasembly.
About subproducts... you are dissasembling so you can create as many as you
want on any step.
Anyway, as my aproach is only a few code lines and very easy to do... I
will try to develop and also propose here if community agrees.

gurneyalex:
hello @anajuaristi @elicoidal 

we extracted this from a customer project, and it solves a simple problem for simple BOMs. Our budget constraints unfortunately do not allow us to change this design right now. I perfectly understand if the PSC does not agree with the design of this module. Just downvote the PR, and we will close it. 

Regarding runbot: this PR depends on an unmerged PR in stock-logistics-warehouse, so yes it will be red until https://github.com/OCA/stock-logistics-warehouse/pull/148 is merged.

cyrilgdn:
@elicoidal I think all your remarks have been addressed.

anajuaristi:
Hi @gurnelayex. I will not downvote any PR than can provide a missing functionality simply because I consider too complex design or I don't like it.
Don't like is a very subjective think, no real reason... so I let other people to decide.
Anyway...Thank you for your hard work!!
:)

elicoidal:
@gurneyalex no problem with the module, all the contrary
@cyrilgdn thanks for the contribution

:+1:

jgrandguillaume:
Hi,

The advantage with this module here is:
- Intuitive to use, because you start from the list of product you want to have / dismantle. 
- Take advantage of the scheduler. The scheduler can produce a MO to dismantle products, so you have the necessary component to build another.
- On a MO, you see what product you take to dismantle it and what product you're looking to have (produce). If you use reverse BoM "ionly" you don't have those infos.

This modules works fine for me. I'm :+1: 

Regards,

Joël

cyrilgdn:
I added the ability to fill lot number for each component when dismantling a product (Currently, if you select a lot number in produce wizard, he will be applied to all dismantled components)

Can someone review that (and merge if ok)?

Thanks.

mrp_bom_dismantling/wizards/mrp_product_produce.py
```diff
@@ -0,0 +1,51 @@
...
+
+from openerp import api, fields, models
+
+
+class MrpByProductLine(models.TransientModel):
+    _name = "mrp.product.produced.line"
+
+    produce_id = fields.Many2one('mrp.product.produce', required=True,
+                                 string="Produce")
+    move_id = fields.Many2one('stock.move', required=True, string='Product')
```
atchuthan: typo mistake for the string

mrp_bom_dismantling/models/mrp_bom.py
```diff
@@ -0,0 +1,139 @@
...
+# © 2016 Cyril Gaudin (Camptocamp)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import _, api, exceptions, fields, models
+
+
+class MrpBom(models.Model):
+    _inherit = 'mrp.bom'
+
+    dismantling = fields.Boolean(string='Dismantling', default=False)
```
atchuthan, line 11: field's optional parameter(string & default) does not have significance as the value is same as not provided

atchuthan:
:+1:

PR #120, opened 2016-04-22
[8.0][IMP][mrp_project] If production come from a sale order, set project parent

antespi:
This PR improves `mrp_project` for this case use:

A production order created by a sale order with an analytic account assigned.

With this PR the project created to manage task related with production order will be child of analytic account assigned to the sale order. This allows to have outcomes (from materials and human resources) and income (invoice generated when sale order is delivered) together below the analytic account assigned to sale order.


pedrobaeza:
:+1:

oihane:
:+1:

PR #121, opened 2016-04-28
[FIX] quality_control: deleting qc.inspection does not delete its lines fixed

oihane:


pedrobaeza:
:+1: Fast-tracking as it an obvious fix.

PR #127, opened 2016-05-23
[FIX] quality_control_mrp: expected singleton error

oihane:


pedrobaeza:
:+1: Fast-tracking as blocking point.

PR #128, opened 2016-05-25
[FIX] quality_control_mrp: partial fabrications...

oihane:
... where creating more inspections than required

### DIFF
--- a/quality_control_mrp/models/mrp_production.py
+++ b/quality_control_mrp/models/mrp_production.py
@@ -24,13 +24,15 @@ class MrpProduction(models.Model):
     @api.model
     def action_produce(self, production_id, production_qty, production_mode,
                        wiz=False):
+        production = self.browse(production_id)
+        done_moves = production.move_created_ids2.filtered(
+            lambda r: r.state == 'done')
         res = super(MrpProduction, self).action_produce(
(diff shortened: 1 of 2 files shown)


quality_control_mrp/models/mrp_production.py
```diff
@@ -24,13 +24,15 @@ def _count_inspections(self):
...
+        done_moves = production.move_created_ids2.filtered(
+            lambda r: r.state == 'done')
         res = super(MrpProduction, self).action_produce(
             production_id, production_qty, production_mode, wiz=wiz)
         if production_mode == 'consume_produce':
             inspection_model = self.env['qc.inspection']
-            production = self.browse(production_id)
             for move in production.move_created_ids2.filtered(
-                    lambda r: r.state == 'done'):
+                    lambda r: r.state == 'done' and r not in done_moves):
```
pedrobaeza: Better to use difference operator: `for move in production.move_created_ids2.filtered(lambda r: r.state == 'done') - done_moves:`

pedrobaeza:
:+1: Thanks for the extra test!

oihane:
After taking a look to the travis error, it is not an error caused by the module in this PR. So please more reviews...

CarlosMolinaP:
:+1: Good Job!!

PR #129, opened 2016-05-25
mrp_bom_dismantling: fix lot_id required bug.

cyrilgdn:
When products are configured with tracking serial number or lot, lot_id is mandatory in produce wizard but mrp_bom_dismantling hides this field.

I removed required on lot_id and implement the same required in the new wizard field.


jgrandguillaume:
Hi,

Thanks for the fix, it works like a charm. :+1: 

Regards,

Joël

OpenCode:
:+1:

gurneyalex:
:+1:

Issue #133, opened 2016-06-01
[quality_control_stock][8.0] No inspection created when transfer product using package


andhit-r:
module: quality_control_stock
version: 8.0

**Steps to reproduce**
1. Create a test
   ![selection_1](https://cloud.githubusercontent.com/assets/1947376/15702694/a9476d40-280b-11e6-80fb-57d1d7666d2d.png)
2. Assign QC trigger on product
   ![selection_2](https://cloud.githubusercontent.com/assets/1947376/15702738/e006d672-280b-11e6-829d-d15a9597b9c1.png)
3. Create a picking that will trigger QC
   ![selection_3](https://cloud.githubusercontent.com/assets/1947376/15702794/34b53b8c-280c-11e6-8ab2-474076b82017.png)
   ![selection_5](https://cloud.githubusercontent.com/assets/1947376/15702799/3c048230-280c-11e6-801f-4898aab9de43.png)
4. Transfered using package
   ![selection_4](https://cloud.githubusercontent.com/assets/1947376/15702820/4b0b7234-280c-11e6-9455-328cc869ecca.png)

**Current behavior**
Inspection did not created

**Expected behavior**
Inspection created

Is it by design? QC inspection won't trigger when transfer product using package

pedrobaeza:
It's a non covered use case.

andhit-r:
@pedrobaeza do you think it will useful for community if i add functionality to cover such use case? If you think it's useful should i add functionality on quality_control_stock or create new module?

Many thanks

pedrobaeza:
Yes, it can be useful in the same module, as packages are also from stock module.

andhit-r:
@pedrobaeza thanks for the confirmation. I will create PR for that functionality.

PR #134, opened 2016-06-01
[quality_control_stock][8.0][IMP] Package transfer trigger qc

andhit-r:
Now package transfer can trigger QC

### DIFF
--- a/quality_control_stock/models/stock_picking.py
+++ b/quality_control_stock/models/stock_picking.py
@@ -37,6 +37,8 @@ class StockPicking(models.Model):
     def do_transfer(self):
         res = super(StockPicking, self).do_transfer()
         inspection_model = self.env['qc.inspection']
+        package_obj = self.env['stock.quant.package']
+        product_obj = self.env['product.product']
         for operation in self.pack_operation_ids:
             qc_trigger = self.env['qc.trigger'].search(
                 [('picking_type', '=', self.picking_type_id.id)])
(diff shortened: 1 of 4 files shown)


quality_control_stock/models/stock_picking.py
```diff
@@ -46,9 +48,20 @@ def do_transfer(self):
                           'qc.trigger.product_line']:
                 partner = (self.partner_id
                            if qc_trigger.partner_selectable else False)
-                trigger_lines = trigger_lines.union(
-                    self.env[model].get_trigger_line_for_product(
-                        qc_trigger, operation.product_id, partner=partner))
+                if not operation.product_id:
+                    dict_product = package_obj._get_all_products_quantities(
```
pedrobaeza, line 52: Isn't better to just retrieve products with a `mapped` operation over the one2many field of the details of the package? You don't need quantities

anajuaristi:
Hi @andhit-r 
Could you please explain the functionality you want to add? It seems interesting but I can not fully understand it just reading the code. It would be great if you could explain the functional behavior change.

Thank you very much
Ana

andhit-r:
Hi @anajuaristi . 
You can look at this issue https://github.com/OCA/manufacture/issues/133. This PR will add functionality use case describe on that issue. Thanks.

anajuaristi:
Ok @andhit-r. I see. Thank you!!

PR #138, opened 2016-06-23
[8.0][FIX] mrp_project: Fix singleton exception in 'ProjectTask.write'

crimoniv:
The `write` method was not able to process multiple records, so they are now iterated to avoid singleton errors. Same fix has been applied in `MrpProductionWorkcenterLine.write` also.

### DIFF
--- a/mrp_project/models/mrp_production.py
+++ b/mrp_project/models/mrp_production.py
@@ -101,5 +101,7 @@ class MrpProductionWorkcenterLine(models.Model):
 
     @api.multi
     def write(self, vals, update=True):
-        return super(MrpProductionWorkcenterLine, self.with_context(
-            production=self.production_id)).write(vals)
+        for rec in self:
+            super(MrpProductionWorkcenterLine, rec.with_context(
+                production=rec.production_id)).write(vals)
+        return True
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@oihane, why this variable was needed to be passed to super?

oihane:
I don't know is something added by @ddico in https://github.com/OCA/manufacture/pull/85

pedrobaeza:
OK, I see the purpose. Then this patch is correct. :+1: 

We can fast-track it as it prevents the normal work.

crimoniv:
In case it helps, I've found that the value is referenced in a few different places:
- [OCA/manufacture/mrp_operations_extension/models/mrp_bom.py:18](https://github.com/OCA/manufacture/blob/8.0/mrp_operations_extension/models/mrp_bom.py#L18)
- [OCA/manufacture/mrp_project/models/hr_analytic_timesheet.py:14](https://github.com/OCA/manufacture/blob/8.0/mrp_project/models/hr_analytic_timesheet.py#L14)
- [odoomrp/odoomrp-wip/mrp_product_variants/models/mrp_bom.py:86,87](https://github.com/odoomrp/odoomrp-wip/blob/8.0/mrp_product_variants/models/mrp_bom.py#L86,L87)
- [odoomrp/odoomrp-wip/mrp_product_variants/models/mrp_bom.py:120](https://github.com/odoomrp/odoomrp-wip/blob/8.0/mrp_product_variants/models/mrp_bom.py#L120)

Edit: Thanks for merging!

PR #147, opened 2016-08-23
[8.0][ADD] mrp_production_manual_close

Daniel-CA:
This module modifies mrp production workflow to finish the production orders manually adding a new close button.

### DIFF
--- a/mrp_production_manual_close/README.rst
+++ b/mrp_production_manual_close/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+MRP Production manual close
+===========================
+
+This module modifies mrp production workflow to finish the production orders
+manually adding a new close button.
+
+Known issues / Roadmap
+======================
+
+When there are pending movements to finish production order, it will cancel all
+movements (including pending procurement orders) and will mark the production
+as done.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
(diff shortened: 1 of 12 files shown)


Daniel-CA:
@pedrobaeza Looks like that the test in mrp_production_real_cost is giving an error because this module changes the behavior when the production is done but not all the products to be produced has been produced.
I am having 2 errors, one with the amount of analytic lines created that doesn't match (remaining movements are cancelled) and the other one is with the amount when the `operation_time_line` start_time changes and the amount doesn't change.
Shoud I change the test on mrp_production_real_costs or I need to do a glue module?

pedrobaeza:
You have to inhibit this behavior when testing if a context key is not provided. Ask @oihane that has made something similar.

oihane:
Here it is an example
https://github.com/odoomrp/odoomrp-wip/blob/8.0/product_variants_no_automatic_creation/models/product_template.py#L56

Daniel-CA:
I tried to avoid behavior adding a condition in the workflow but I have an empty frozendict on context so I can't add the key to differentiate code execution.

andhit-r:
Hi @pedrobaeza @oihane @Daniel-CA , i tried to solve travis problem by using @oihane example. I could not see that example can be used for this PR. No behavior in mrp_production_manual_close that needed to be inhibit. Would you kindly give pointer which behavior that need to be inhibited?

Meanwhile i have tested to:

1. change test on mrp_production_real_cost (https://github.com/andhit-r/manufacture/tree/8.0-fix-issue-147). I only made sure that assert on analytic line happen when production state is done (see https://github.com/andhit-r/manufacture/blob/8.0-fix-issue-147/mrp_production_real_cost/tests/test_mrp_production_real_cost.py#L43, https://github.com/andhit-r/manufacture/blob/8.0-fix-issue-147/mrp_production_real_cost/tests/test_mrp_production_real_cost.py#L74, and https://github.com/andhit-r/manufacture/blob/8.0-fix-issue-147/mrp_production_real_cost/tests/test_mrp_production_real_cost.py#L90), and
2. Migrate **mrp_production_manual cost**

It passed all travis checks.

Then i tried to alter the test without mirgating mrp_production_manual_cost. It passed travis checks and same coverage score (OCA/8.0 https://coveralls.io/github/OCA/manufacture?branch=8.0 against https://coveralls.io/github/andhit-r/manufacture?branch=8.0-real-cost-alter-test).

Is it possible that we just go ahead with altering test on mrp_production_real_cost?

Issue #124, opened 2016-05-16
is:issue is:open [8.0][mrp_production_real_cost] Error when produce product


yastri:
Hi,

there's new error from mrp_production_real_cost, after I do git pull from the last commit 

```

ERROR demo1 openerp.sql_db: Programming error: can't adapt type 'mrp.production', in query  SELECT "mrp_production"."id" FROM "mrp_production"
                        WHERE "mrp_production".id IN %s  ORDER BY "mrp_production"."priority" DESC,"mrp_production"."date_planned" ASC         

  File "/opt/odoo/server/addons/mrp_production_real_cost/models/mrp_production.py", line 34, in action_production_end
    self.mapped('move_created_ids2').filtered(


  File "/usr/lib/python2.7/dist-packages/psycopg2/extensions.py", line 129, in getquoted
    pobjs = [adapt(o) for o in self._seq]
ValueError: "can't adapt type 'mrp.production'" while evaluating
u'action_production_end()'
```

regards

pedrobaeza:
You need to update your module and restart server.

yastri:
I've already update the module and all dependency and restart server, but still error.
I try to uninstall, after that manufacturing can do produce, but when I install again there's same error when I push the button produce.

pedrobaeza:
The log you have put proves that it wasn't updated and restarted, because the referenced line is not the same as the shown one. Has the error log changed?

sandrafig:
I just had the same error, I've updated all my repositories this morning and the database aswell

**ValueError: "can't adapt type 'mrp.production'" while evaluating
u'action_production_end()'**

pedrobaeza:
Do you have a module from odoomrp-wip called mrp_production_manual_close or similar?

sandrafig:
I don't have that module installed

pedrobaeza:
Then please give me an update error log. The first one is not correct

sandrafig:
[...]

PR #149, opened 2016-09-07
[8.0][FIX]mrp_production_real_cost: fix error in old api inherit

maurochip:
The function '_costs_generate' has an api.multi, expecting that the 3rd parameter in old API is 'ids', but it is a 'mrp.production' record instead. Convert to api.model resolves the error.
This could be related to this issue #124

### DIFF
--- a/mrp_production_real_cost/models/mrp_production.py
+++ b/mrp_production_real_cost/models/mrp_production.py
@@ -81,8 +81,8 @@ class MrpProduction(models.Model):
             'general_account_id': general_account.id,
         }
 
-    @api.multi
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:, but it's not related

maurochip:
It is related. It messes up the ORM cache as it is adding the "mrp.production" record to the "ids" list when browsing related records, giving the error `ValueError: "can't adapt type 'mrp.production'" while evaluating`.

pedrobaeza:
OK, I get it. I fast track it then.

PR #152, opened 2016-09-23
[IMP] mrp_operations_extension: allow to load operation in routing lines

agaldona:
@anajuaristi reported an issue in routing lines. When you select an operation it disappears. I don't know why was it like this, but i think it is a bug.


anajuaristi (approved):

anajuaristi:
Tested functionally, it works

oihane:
I found that @pedrobaeza changed it
https://github.com/OCA/manufacture/commit/3c6fd5cf4105f2bc80032e2673205f72e720bd51#diff-697494787cd211366fb2099bff990859R82

I don't remember why he did it that way.

anajuaristi:
We need this PR merged and we had no comment from september so I do it.

PR #156, opened 2016-10-04
Add missing french translations

yterrettaz:
To complete the traduction of mrp_sale_info in French. It's my first OCA PR, thank you for your indulgence :-)


yterrettaz:
So I realized that I have to use Transifex for this operation :-/

Issue #159, opened 2016-10-17
Number of cycles in mrp.routing.workcenter


rubencabrera:
In the tree view of mrp.operation.workcenter inside the mrp_routing_form_view the field "Number of Cycles" actually shows the value of "capacity_per_cycle" from the workcenter. 

Is the name of the column wrong or should it do something else?

pedrobaeza:
Yeah, the name is wrong. You can make a PR changing it.

PR #160, opened 2016-10-17
Set cycle_nbr to visible and delete onchange field assignation

rubencabrera:
As mentioned in #159 , the column in the tree view references number of cycles but the onchange method assigns the capacity per cycle, resulting in a confusing output. 

As the field definition in the odoo mrp module states clearly that cycle_nbr is meant for:
Number of iterations this work center has to do in the specified operation of the routing
...and being in need for some counter for cycles when a manufacturing requires repeating an operation multiple times, modified back the onchange_lines_default and also the form view to show the Counter for cycles again.

### DIFF
--- a/mrp_operations_extension/i18n/es.po
+++ b/mrp_operations_extension/i18n/es.po
@@ -1,25 +1,21 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-# * mrp_operations_extension
-# 
-# Translators:
-# Matjaž Mozetič <m.mozetic@matmoz.si>, 2015
-# oihane <oihanecruce@gmail.com>, 2016
-# oihane <oihanecruce@gmail.com>, 2016
-# oihane <oihanecruce@gmail.com>, 2016
+# 	* mrp_operations_extension
+#
 msgid ""
 msgstr ""
-"Project-Id-Version: manufacture (8.0)\n"
+"Project-Id-Version: Odoo Server 8.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-09-17 22:50+0000\n"
-"PO-Revision-Date: 2016-10-04 09:43+0000\n"
-"Last-Translator: oihane <oihanecruce@gmail.com>\n"
-"Language-Team: Spanish (http://www.transifex.com/oca/OCA-manufacture-8-0/language/es/)\n"
+"POT-Creation-Date: 2016-11-02 08:34+0000\n"
+"PO-Revision-Date: 2016-11-02 09:40+0100\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@anajuaristi, what's your opinion on this?

@rubencabrera check this graph about how costs work: https://github.com/OCA/manufacture/pull/83#issuecomment-176890577

rubencabrera:
Thank you for the link, @pedrobaeza , it is clearer now.

This PR tries to make use of the Number of Cycles field in the mrp.routing.workspace model, as it seems the most logical place to take it into account when the same workspace has to be repeated in a manufacturing process.

anajuaristi:
@rubencabrera @pedrobaeza
In my point of view is correct as it is now just because on routing it's imposible to knoe the exact number of cycles you will need to do as this value will be calculated on Mo never on routing.
If you need to repeat same operation on routing several times it will be enough just to duplicate it. But this is not the meaning of number of cycles.
IMHO unit of cicles it's related to MO as units per cycle is related to routing step. Both fields are required each one in their correct place.
Anyway if you find usefull to make the field visble do it but it will not be considered in any place if you change manually so i would recomend not to merge the PR. I'm sorry for that.
Thank you for contributing
Ana

rubencabrera:
Then what is the purpose of the cycle_nbr field inside the mrp.routing.workcenter model? 
In the whole repo it's only used in the onchange_lines_default method, where it is assigned the value of capacity_per_cycle. How is that right? Aren't they opposite concepts? If Capacity per Cycle gets bigger, number of cycles should be smaller.  
In the MO, cycle_nbr from the routing object is not used by action_compute when computing the mrp.production.workcenter.line. All the routing elements are imported but their cycle_nbr is ignored.

I accept your point about routing not being the place to set the number of cycles, but why having a field there for it and then making it invisible? Where should it go to be useful?

anajuaristi:
There is no purpose for it, but it's a field included on Odoo core, so we can not erase but reuse and try to make our developments as compatible as is possible.

:)

rubencabrera:
@anajuaristi , you are right about the work order being the best place to set the number of cycles a manufacturing operation might actually take. 

There is the field number of cycles.

mrp.production.workcenter.line has a Number of cycles field. 
Also, mrp.routing is linked to mrp.routing.workcenter, where the field cycle_nbr I am insisting on is defined. On the tree view of mrp.routing, the value of cycle_nbr is shown, but it is WRONGLY taken from the production efficiency (I am positive about this being wrong because the field is not corresponding its label).

If MyProductionRoute always makes me use 2 cycles of a precise workcenter, why do I have to set it manually when there is a field for that in the mrp.routing.workcenter, linked and visible in mrp.routing, but not taken into account in the MO?

rubencabrera:
Sorry, the latest commits are not related to this pull request. As the argument seems stale now, I'll close it. Thanks for your input.

PR #162, opened 2016-10-28
[MOD] mrp_operations_extension

esthermm:
There is an issue about adding an element to the workcenter lines, there is an onchange that fills the fields with the values, but as they are readonly when you save the changes of those fields are not saved. So custom_data field is not necessary.


pedrobaeza:
:-1: You shouldn't remove this option. They are not readonly when you mark custom_data, so no issue at all.

anajuaristi:
Could be it's not a technical issue but functional issue.
In my opinion check is not necesary at all. It makes it more complex to
understand the concept and at the end the valid data can be always taken
from routing step.

Check is making necesary more code to control and it's not improving
anything.
This change was asked by one of hour customers and i aproved it
functionally.
Please don't stop the merge.

Manufacturing experts... could you give you opinion on this?
@jordieficent

pedrobaeza:
@anajuaristi, the only to control if the route has specific data is through that check, and you were one of the defenders of that option for your use cases, so I don't understand why you change your mind now.

anajuaristi:
No. I wasn't
I always thougth it was no necesary but you included yourself so I let it stay. 
But now I think it's causing usability problems to my customers and it's posible to simplify code so I said Esther to erase it.

anajuaristi:
Thinking about this, maybe make check active by default (so editable line ) would be enough
I mean... it's much more usual taking cost data from routing line than from machine so just active by default this option but still we would have the option to take from machine.

This way we improve usability withouth loosing functionality.

pedrobaeza:
OK, change the default and repair Travis.

anajuaristi:
Ok @pedrobaeza 
@esthermm please... change the default to set check true and repair travis

pedrobaeza:
Thanks for the changes. Travis is still red I'm afraid.

esthermm:
Yes, I know, mrp_production_estimated_cost test is failing. I still don't know why

LoisRForgeFlow:
@esthermm Is this still relevant? are you going to finish the work? Thanks.

anajuaristi:
@lreficent it's solved. I'm closing.
Thank you

PR #164, opened 2016-11-11
[IMP] mrp_production_estimated_cost: 

agaldona:
When changing the production qty to 0, a traceback error raised "Float Division by Zero".


oihane (approved):
:+1: 
LGTM, it is a good solution to avoid division by zero error, although it is not common to create a MO with quantity zero (system does not allow to save a zero quantity MO)

Daniel-CA:
:+1:

esthermm:
:+1:

pedrobaeza (approved):

PR #169, opened 2016-11-30
8.0 mrp_project: settings to deactivate automatic task generation

rubencabrera:
This changes add 2 settings in Manufacturing Settings to set default value for 2 Boolean fields in MO.

These fields are the condition checked before creating a related project for the current MO and the Automatically generated tasks when changing states to "In production". 

The default behavior doesn't change the current one.

### DIFF
--- a/mrp_project/models/mrp_production.py
+++ b/mrp_project/models/mrp_production.py
@@ -16,10 +16,10 @@ class MrpProduction(models.Model):
         related="project_id.analytic_account_id", store=True)
     create_tasks = fields.Boolean(
         string="Automatically create tasks",
-        help=_(""" Automatically create tasks linked to manufacturing orders in
+        help=""" Automatically create tasks linked to manufacturing orders in
                assigned project. The tasks will be created when the production
                is started.
-               """),
+               """,
         readonly=True,
         states={'draft': [('readonly', False)],
                 'confirmed': [('readonly', False)]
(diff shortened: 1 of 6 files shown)


anajuaristi (commented):
I can not aprove this changes since the meaning about creating a task or creating a project on mrp_project is a different one.
The current focus is creating a task related to workers assigned in a routing step. You just need to set 0 on the field to avoid creating tasks.
About the project related to an MO, is necesary in any case there is workers assigned to one routing step on the MO, and it's also necesary on the case stimated or real cost is installed, so... If you need another functionality or you need to change the meaning of module, I'm sorry but, it will be necesary you build another module.

The idea is creating a task for each worker. It's not needed creating a task for each WO, since you have the wo by now.

Sorry but, from my side the PR should be rejected.

Thank you:

Ana

mrp_project/models/mrp_production.py
```diff
@@ -63,21 +79,35 @@ def _prepare_production_task(self, production):
...
-        task_obj = self.env['project.task']
-        for record in self:
-            task_domain = [('mrp_production_id', '=', record.id),
-                           ('workorder', '=', False)]
-            tasks = task_obj.search(task_domain)
-            if not tasks:
-                task_obj.create(self._prepare_production_task(record))
+        if not self.project_id:
+            raise exceptions.ValidationError(_(
+                "To create a task per work order, a project must be selected. "
```
anajuaristi: This is the cause of my rejection.
You don't need to create a task per work order. 
One single task will be created for the MO.
If you need creating one task by WO, you just need to install the module mrp_operations_project and set, one single worker for each routing step.

To create a task related to each WO, is not needed if there is no workers on it.

PR #175, opened 2017-02-07
[FIX] mrp_operations_extension: calendar view error because non stored field

agaldona:


oihane (approved):
:+1: 

LGTM

oihane:
@pedrobaeza @gurneyalex
I think is a trivial change as `date_planned_end` is a computed field that is computed by adding `hour` to `date_planned`
Could you please take a look?

pedrobaeza (approved):

PR #181, opened 2017-02-21
[FIX] mrp_operations_extension: Don't check option "Calc Cycles by Bo…

Daniel-CA:
Don't check option "Calc Cycles by BoM Quantity" even if not desired.

### DIFF
--- a/mrp_operations_extension/models/res_config.py
+++ b/mrp_operations_extension/models/res_config.py
@@ -22,12 +22,9 @@ class MrpConfigSettings(models.TransientModel):
         param_obj = self.env['ir.config_parameter']
         rec = self._get_parameter(key)
         if rec:
-            if not value:
-                rec.unlink()
-            else:
-                rec.value = value
-        elif value:
-            param_obj.create({'key': key, 'value': value})
+            rec.value = str(value)
+        else:
+            param_obj.create({'key': key, 'value': str(value)})
 
     @api.multi
     def get_default_parameter_cycle_bom(self):
--- a/mrp_operations_extension/tests/test_mrp_operations_extension.py
+++ b/mrp_operations_extension/tests/test_mrp_operations_extension.py
@@ -230,3 +230,18 @@ class TestMrpOperationsExtension(common.TransactionCase):
                              'Error work order moves quantity do not match')
             self.assertEqual(workorder2.state, 'done',
                              'Error work center line not in done state')
+
+    def test_param_config(self):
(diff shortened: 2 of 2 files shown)


pedrobaeza:
The option is correct. What you need to do is to put to False the parameter.

Daniel-CA:
Hi @pedrobaeza, the problem comes when you update the module or due to an update=all, it always checks it again, even if you have marked to False.

pedrobaeza:
No, the noupdate="1" prevents that. The only problem is that you remove the record, then that attribute makes that to recreate it. But it's put on purpose for always seeing the value (False or True). Try again.

Daniel-CA:
Hi @pedrobaeza,
You are right, but when I try to update the attribute value I get an integrity error because the record is already accessed, don't know why it leaves unlinking it but not updating.
```
    if rec:
            if not value:      
                rec.unlink()
            else:
                rec.value = value
```
My change:
```
    if rec:
        rec.value = value
```
Any suggestions?

pedrobaeza:
I don't get you. What is the relation between the PR and your code?

Daniel-CA:
The PR is a "fix" for the problem I was talking, when the mark is uncheked the param key is erased, so if the module is updated the key is recreated again with the True value even if it is in a noupdate section.

pedrobaeza:
But where is that code? The problem is in your code, not here.

Daniel-CA:
That code is in the `res_config.py` file.

pedrobaeza:
OK, the error is there then. You have to put:
```
    def _write_or_create_param(self, key, value):
        param_obj = self.env['ir.config_parameter']
        rec = self._get_parameter(key)
        if rec:
            rec.value = str(value)
        else:
            param_obj.create({'key': key, 'value': str(value)})
```

Daniel-CA:
That's the change that I was talking about in my previous message.

pedrobaeza:
OK, we have agreement then!

Daniel-CA:
And do you have any suggestions to avoid the integrity error?

pedrobaeza:
With the new code do you still have that problem?

Daniel-CA:
Yes, and don't know why.

Daniel-CA:
Changes done! Thanks @pedrobaeza

pedrobaeza (approved):

elicoidal (approved):

agaldona (approved):
:+1:

PR #185, opened 2017-03-28
[8.0][FIX] mrp_operations_extension: Check option "Calc Cycles by BoM…

Daniel-CA:
Check option "Calc Cycles by Bom Quantity" does not update correctly.

Comes from here:  #181 - Now 'False' is like True, so set function is not correctly updated.


pedrobaeza (approved):

Daniel-CA:
Test updated

elicoidal (approved):

agaldona (approved):
:+1:

PR #188, opened 2017-04-20
[FIX][8.0] mrp_operations_extension: update "Calc Cycles by Bom Quant…

Daniel-CA:
Update "Calc Cycles by Bom Quantity" condition to new logic
Comes from here: #181 - Now 'False' is like True, so condition is not correctly defined.


Daniel-CA:
Can someone review this? 
@pedrobaeza  @elicoidal

agaldona (approved):

elicoidal (approved):

anajuaristi (approved):

Issue #189, opened 2017-04-24
[9.0] mrp_bom_dismantling: Dependency not used?


LoisRForgeFlow:
I'm wondering why this dependency is needed here: https://github.com/OCA/manufacture/blob/9.0/mrp_bom_dismantling/__openerp__.py#L16

cc @cyrilgdn

cyrilgdn:
@lreficent He uses components computation method from this module.

This method https://github.com/OCA/stock-logistics-warehouse/blob/9.0/stock_available_mrp/models/product_product.py#L92
Is called here: https://github.com/OCA/manufacture/blob/9.0/mrp_bom_dismantling/models/mrp_bom.py#L153

JordiBForgeFlow:
@cyrilgdn I propose to remove the dependency, and define a new method that does the same in this module. Because that method is generic. We are really not instested in what 'stock_available_mrp' does, and it is adding unneded complexity to us. Better decouple, IMHO. Can we propose a PR?

LoisRForgeFlow:
Related PR has been merged.

PR #190, opened 2017-04-26
[9.0] mrp_bom_dismantling: remove dependency

LoisRForgeFlow:
Little change to remove the dependency in stock_available_mrp of mrp_bom_dismantling as response to discussion in https://github.com/OCA/manufacture/issues/189

### DIFF
--- a/mrp_bom_dismantling/__openerp__.py
+++ b/mrp_bom_dismantling/__openerp__.py
@@ -4,7 +4,7 @@
 {
     "name": "BOM Dismantling",
     "summary": "Ability to create a dismantling BOM by reversing a BOM.",
-    "version": "9.0.1.0.0",
+    "version": "9.0.1.0.1",
     "category": "Manufacturing",
     "website": "http://www.camptocamp.com/",
     "author": "Camptocamp, Odoo Community Association (OCA)",
--- a/mrp_bom_dismantling/__openerp__.py
+++ b/mrp_bom_dismantling/__openerp__.py
@@ -13,7 +13,6 @@
     "installable": True,
     "depends": [
         'mrp_byproduct',
-        "stock_available_mrp",
     ],
     "data": [
         "views/mrp_bom.xml",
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow:
I pushed another commit changing the method name to avoid problems in databases with both modules installed `mrp_bom_dismantling` and `stock_available_mrp`. Not sure whether this is the best option.

JordiBForgeFlow (approved):

LoisRForgeFlow:
In my local the testd are running fine, but travis tests are failing [here](https://github.com/OCA/manufacture/blob/10.0/mrp_bom_dismantling/tests/test_template.py#L20). I checked another PR in the repository and they are failling also there, however I cannot find the reason, it seems a simple test:
1. Create a product.template
2. Create a BOM related
3. Check the bom_count expecting it to have changed from 0 to 1.

The compute method for bom_count is overrited in this module. But neither the original method (`_bom_orders_count`) or the new compute method (`_bom_count`) are inherited nor modified in any other module in  this repository.

JordiBForgeFlow:
@cyrilgdn Do you agree?

serpentcs-dev1:
@lreficent I checked in my local the [computation method](https://github.com/OCA/manufacture/blob/10.0/mrp_bom_dismantling/models/product_template.py#L15) is triggered when the template is created but not when the bom for that is created so, maybe you have to call it manually before asserting the value.
`tmpl1._bom_count()`

LoisRForgeFlow:
@darshan-serpent yeah, let's give it a try!

LoisRForgeFlow:
@darshan-serpent Fixed now, can you give your review here please?

OSevangelist (approved):
Changes seem reasonable to me

pedrobaeza (changes requested):
Please change version number

serpentcs-dev1 (approved):
Code Review LGTM :+1:

LoisRForgeFlow:
@pedrobaeza version bumped

LoisRForgeFlow:
ping @pedrobaeza your review is blocking, could you check?

pedrobaeza (approved):

PR #191, opened 2017-04-26
[9.0][ADD] mrp_mto_with_stock

LoisRForgeFlow:
The author of this module is @michaeljohn32. I just adapt it to OCA standards, document it and do minor fixes.

MRP MTO with Stock
==================

This module extends the functionality of Manufacturing to support the creation
of procurements when there is no stock available. This allow you to pull from
stock until the quantity on hand is zero, and then create a procurement
for fulfill the MO requirements.

Configuration
=============

To configure this module, you need to:

1. Go to the products you want to follow this behaviour.
2. In the view form go to the tab *Inventory* and set the *Manufacturing
    MTO/MTS Locations*. Any other location not specified here will have the
    standard behavior.

Usage
=====

To use this module, you need to:

#. Go to *Manufacturing* and create a Manufacturing Order.
#. Click on *Confirm Production*.


LoisRForgeFlow:
@michaeljohn32 Do you think this is fine?

Also if you provide me your mail I would add it to the contributors section.

AlainVanUtterbeeck:
I did some testing and found a problem:
- we use a production zone, which is defined in the bom, so the system creates new moves on confirmation (a picklist to move the goods from stock to production). These moves are the ones that need to be verified for available and misssing stock instead of the original ones (which remain in state waiting till the picklist is issued), now the code does not handle this way of working.

AlainVanUtterbeeck:
@lreficent 
in mrp.py, adding these 3 lines fix the problem:

    @api.one
    def action_assign(self):
        """Reserves available products to the production order but also reates
        procurements for more items if we cannot reserve enough (MTO with
        stock).
        @returns list of ids"""
        # reserve all that is available (standard behaviour):
        res = super(MrpProduction, self).action_assign()
        # try to create procurements:
        for move in self.move_lines:

+            new_move = self.env['stock.move'].search([('move_dest_id','=',move.id)])
+            if new_move:
+                move = new_move

            if (move.state == 'confirmed' and move.location_id in
                    move.product_id.mrp_mts_mto_location_ids):
                domain = [('product_id', '=', move.product_id.id),
                          ('state', '=', 'running'),
                          ('move_dest_id', '=', move.id)]

AlainVanUtterbeeck:
@lreficent
Forget my previous remark please, it is not the solution.
I will continue testing...

LoisRForgeFlow:
@AlainVanUtterbeeck Ok, let me know whenever you find some issue :wink:

MiquelRForgeFlow:
I think this module should have tests.

JordiBForgeFlow:
@mreficent @lreficent added test cases, and minor fixes, including:
- search for any procurement associated to the component and MO, regardless of the state of the existing procurement, before creating a new one.
- Do not auto-reserve the MO upon confirmation

mrp_mto_with_stock/models/__init__.py
```diff
@@ -3,5 +3,5 @@
 # Copyright 2015 John Walsh
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from . import mrp
```
MiquelRForgeFlow: then the `mrp.py` file should be deleted

MiquelRForgeFlow (changes requested):

mrp_mto_with_stock/models/__init__.py
LoisRForgeFlow: fixed @mreficent

mrp_mto_with_stock/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* John Walsh <?>
```
MiquelRForgeFlow: `<John.Walsh@interclean.com>`

LoisRForgeFlow: Thanks!

MiquelRForgeFlow (approved):
LGTM :+1:

florian-dacosta (approved):

wangyueming:
in mrp_production.py

I think 

`                        if qty_to_procure < move.product_uom_qty:
`

should be

`                        if qty_to_procure < move.product_qty:
`

LoisRForgeFlow:
This got merged in v10, can we move forward here too?

LoisRForgeFlow:
@OCA/manufacturing-maintainers can we get an update on this?

jgrandguillaume (approved):
LGTM

PR #193, opened 2017-05-04
[FIX] quality_control: Fix TypeError on new test category

zamberjo:
`TestQualityControl` is failing on `QcTestTemplateCatergory._get_complete_name`
when `name` is set as `False`.

### DIFF
--- a/quality_control/models/qc_test_category.py
+++ b/quality_control/models/qc_test_category.py
@@ -9,16 +9,16 @@ class QcTestTemplateCategory(models.Model):
     _name = 'qc.test.category'
     _description = 'Test category'
 
-    @api.one
+    @api.multi
     @api.depends('name', 'parent_id')
     def _get_complete_name(self):
-        if self.name:
-            names = [self.name]
-            parent = self.parent_id
+        for record in self:
+            names = [record.name or '']
+            parent = record.parent_id
             while parent:
                 names.append(parent.name)
                 parent = parent.parent_id
-            self.complete_name = " / ".join(reversed(names))
+            record.complete_name = " / ".join(reversed(names))
 
     @api.constrains('parent_id')
     def _check_recursion(self):
--- a/quality_control/models/qc_test_category.py
+++ b/quality_control/models/qc_test_category.py
@@ -12,7 +12,9 @@ class QcTestTemplateCategory(models.Model):
     @api.one
     @api.depends('name', 'parent_id')
     def _get_complete_name(self):
-        names = [self.name or '']
+        if not self.name:
+            continue
+        names = [self.name]
         parent = self.parent_id
         while parent:
             names.append(parent.name)
--- a/quality_control/models/qc_test_category.py
+++ b/quality_control/models/qc_test_category.py
@@ -12,14 +12,13 @@ class QcTestTemplateCategory(models.Model):
     @api.one
     @api.depends('name', 'parent_id')
     def _get_complete_name(self):
-        if not self.name:
-            continue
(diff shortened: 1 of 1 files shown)


oihane:
If `name` is required how can it be `False`?

zamberjo:
Is failling on line:20 (`self.complete_name = " / ".join(reversed(names))`) when computing values for new records (not yet persisted) as `join` is expeacting a list of strings and receives a list of booleans (`[False]`)

oihane:
Won't be more correct not to compute the `complete_name` if there is not `name`?

zamberjo:
LGTM

oihane (approved):
LGTM

agaldona (approved):

quality_control/models/qc_test_category.py
```diff
@@ -12,12 +12,13 @@ class QcTestTemplateCategory(models.Model):
     @api.one
     @api.depends('name', 'parent_id')
     def _get_complete_name(self):
-        names = [self.name]
-        parent = self.parent_id
-        while parent:
-            names.append(parent.name)
-            parent = parent.parent_id
-        self.complete_name = " / ".join(reversed(names))
+        if self.name:
```
pedrobaeza: Please don't make the diff bigger and increase cyclomatic complexity. Make instead:
```
if not self.name:
    return
...
```

oihane: Wouldn't be more correct to convert it to `@api.multi`
```python
    for record in self.filtered('name'):
        names = [record.name]
        parent = record.parent_id
        while parent:
            names.append(parent.name)
            parent = parent.parent_id
        record.complete_name = " / ".join(reversed(names))     
```

LoisRForgeFlow:
Just a favor, can you squash to only one commit when merging?  :grin:, that way is very fast for me to move this changes to v9 PR.

MiquelRForgeFlow:
Please @zamberjo , will you update the PR?

damendieta (approved):

quality_control/models/qc_test_category.py
damendieta: I think it's better to just add a or '' to the code, like this:

names = [record.name or '']

But it works, so, approved.

pedrobaeza:
I think one of the suggestions should be applied at least before merging

damendieta:
@zamberjo are you going to apply the suggestions.

As @zamberjo hasn't commented or made any commit from may 10, maybe it's better to close this PR and open a new one with a fix.

It's a one liner 3 characters fix and it's hanging for almost 6 months.

Please @zamberjo, could you be so kind to help us with this modifications?

Thanks.

LoisRForgeFlow:
I agree with @damendieta , this is a very small fix and it has to move forward... 

EDIT: Please, @damendieta open a PR with your proposal and we can supersede this PR.

zamberjo:
Sorry, I plan to review all outstanding PRs -including this one- over the week.

LoisRForgeFlow:
@zamberjo ok, let's keep this one then.

damendieta:
@zamberjo thanks a lot.
If possible, it's better to fix it just with this:

names = [record.name or '']

Adding the empty string alternative will solve the problem with the least code and no need to add a new indent to the formula, also, only one line will change.

Could you please consider this fix instead.

Thanks a lot.

zamberjo:
I should change the method name `_get_complete_name` to `_compute_complete_name`?

damendieta (approved):

PR #194, opened 2017-05-09
[8.0][IMP] mrp_operations_extesion: fix warning in log

oihane:


mrp_operations_extension/models/mrp_bom.py
```diff
@@ -13,8 +13,8 @@ class MrpBom(models.Model):
     def _prepare_wc_line(self, wc_use, level=0, factor=1):
         res = super(MrpBom, self)._prepare_wc_line(
             wc_use, level=level, factor=factor)
-        if not self.env['mrp.config.settings']._get_parameter('cycle.by.bom',
-                                                              False) == 'True':
+        if not self.env['mrp.config.settings']._get_parameter(
+                'cycle.by.bom').value:
```
pedrobaeza: What happens if the parameter doesn't exist?

oihane: If the parameter exists it will have value = 'True', if it shall be False, parameter won't exist.

pedrobaeza: But then `False.value` will throw an error...

Daniel-CA:
:+1:

pedrobaeza (approved):

esthermm:
:+1:

oihane:
I've taken a look to the tests in branch 8.0 it takes [6.399s](https://travis-ci.org/OCA/manufacture/jobs/227919465#L1233) to pass all the test in this module, and in this PR [4.961s](https://travis-ci.org/OCA/manufacture/jobs/230294285#L1225)
it is not a big difference but it counts.

cc @jbeficent @tafaRU @andhit-r

tafaRU (approved):

anajuaristi:
:+1:

PR #93, opened 2016-02-16
[IMP] mrp_sale_info: Add Sale Line in "Other information" Manufacturi…

victor-ditay:
…ng (MO and WO) also add Customer and Sale Order in search view "Group by" for Manufacturing (MO and WO)

cc @elicoidal

PR #201, opened 2017-06-01
[10.0] Migrate mrp_sale_info

bimajatiwijaya:
- Change workcenter.line to mrp.workorder
- move_prod_id not exist. search sale information from move_finished_ids and changed related field to compute field
- add unit test


mamcode:
Hi @bimajatiwijaya, I have added changes to this pull request locally, are the same changes proposed in the pr #93 but for version 10.0. Except that i don't add the field sale_line because I don't see useful as already the MO/WO has relation with sale.order through field sale_id.

Do you think it's convenient to add my commit to this pr or should I create a new pull request based on this pr (is this possible?)? 

@pedrobaeza

pedrobaeza:
You can make a PR over the original Bima's branch.

mamcode:
Thanks @pedrobaeza I will do that :+1:

mamcode:
@bimajatiwijaya @pedrobaeza PR done https://github.com/bimajatiwijaya/manufacture/pull/1

humanytek:
@bimajatiwijaya Fields sale_id and partner_id must be stored because the filters (group by) do not work ... the fields to be grouped must be fields stored in the database, it does not work with computed or related fields o/

mamcode:
@bimajatiwijaya Fields sale_id and partner_id must be stored because the filters (group by) do not work ... the fields to be grouped must be fields stored in the database, it does not work with computed or related fields o/

bimajatiwijaya:
Hi @mamcode ,
Before both field are computed field. 
If using stored sale_id and partner_id will empty. Unit test always failed.
So i changed way to get sale info. Now all field are not computed but stored value.
I extend make_mo to fulfill sale information.

mamcode:
Now I understand @bimajatiwijaya :-) Good!

congdpt (approved):

bimajatiwijaya:
Up :)

LoisRForgeFlow (approved):

PR #205, opened 2017-06-07
[8.0][FIX] quality_control: not take into account inactivated tests trigger lines

agaldona:


pedrobaeza:
This shouldn't be needed, as inactive records (with `active=False`) are automatically discarded by ORM...

agaldona:
Bat there is no active field in trigger lines, it is in the test, and you are filtering trigger lines. We have received an issue about it.

pedrobaeza (approved):
OK, I see.

agaldona:
Other option would be creating an active field related to the test's active field in trigger lines. And it would be filtered by orm.

pedrobaeza:
OK, seems correct this way.

andhit-r (approved):

oihane (approved):
LGTM

agaldona:
@pedrobaeza Can we merge this PR?

PR #208, opened 2017-06-08
[9.0][ADD] mrp_split (WIP)

LoisRForgeFlow:
Responds to issue https://github.com/OCA/manufacture/issues/206

This is still a WIP and there are some points to improve/discuss (maybe more):
* [ ] Take into account consumable products.
* [ ] Picking direct validation lead to error. 

To reproduce this, follow these steps for instance:
1. In the manufacture route add a procurement rule to feed a location A with location Stock.
2. Create a procurement for location A of a manufactured product (MP) this will create a MO1 for MP and a picking from stock to location A.
3. Split MO1 and produce both, MO1 and new MO2.
4. Go to the picking created previously.
5. A) If you modify the the "Done" qty with all the "To Do" qty and click on validate you will be asked to create a backorder (with the qty in MO2).
5. B) If you click in validate and accept the immediate transfer you won't create a backorder and the MO2 is stuck. 

* [ ] Add security group related to splitting MOs?
* [ ] Cancelled stock moves are shown as consumed raw materials. This will require to modify a standard field, not sure if it is the good call.
image

### DIFF
--- a/mrp_split/README.rst
+++ b/mrp_split/README.rst
@@ -1,0 +1,74 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========
+Mrp Split
+=========
+
+This module extends the functionality of Manufacturing to allow you to split
+existing Manufacturing Order (MO) in *New* or *Awaiting Raw Materials* state.
+The proposed quantity to split will ensure that the originating MO will be
+ready to produce (if possible).
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Manufacturing > Manufacturing Orders*.
+#. Open a Manufacturing Order in state *New* or *Awaiting Raw Materials*.
+#. Click the button *Split MO*.
+#. In the wizard, click on *Compute lines* to get a quantity recommendation to
+   split.
+#. Change the *Quantity to split* if desired and click on *Split* button at
+   the bottom of the form.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/129/9.0
+
+Known issues / Roadmap
+======================
+
+* Take into account consumable products.
+* Picking direct validation lead to error.
+* Add security group related to splitting MOs?
+* Cancelled stock moves are shown as consumed raw materials.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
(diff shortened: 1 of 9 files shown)


mrp_split/README.rst
```diff
@@ -0,0 +1,74 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/129/9.0
+
+Known issues / Roadmap
+======================
+
+* Take into account consumable products.
+* Picking direct validation lead to error.
+* Add security group related to splitting MOs?
+* Cancelled stock moves are shown as consumed raw materials.
```
andhit-r, line 37: @lreficent IMO you should not worry about cancelled stock moves. Just leave it there. Changing original field to hide cancelled move is not a wise choice. Cancelled stock move information still needed on some case.

andhit-r:
**Functional Test 1**

1. Create procurement order manually for 5 unit of *[LAP-CUS] Laptop Customized*. A MO will be created automatically.
2. Confirm MO from step 1
3. Split into 2 unit

Original MO will have Product to Produce

mrp-split-1

This case also happen if you create MO manually (without procurement order)

mrp_split/models/mrp_production.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = 'mrp.production'
+
+    backorder_id = fields.Many2one(
+        comodel_name='mrp.production', string="Backorder of", copy=False)
```
andhit-r: Make this field readonly

mrp_split/models/mrp_production.py
```diff
@@ -0,0 +1,14 @@
...
+    backorder_ids = fields.One2many(
```
andhit-r: Make this field readonly

andhit-r (changes requested):

LoisRForgeFlow:
Hi @andhit-r and thanks for the review!

Yes, we had noticed that issue you mentioned. But, eventually we moved to a different approach to solve the issue related to this, see [mrp_production_request](https://github.com/OCA/manufacture/pull/210) which I think is ready to review (even with test added and all checks in green).

We don't have plans to works on this for the sort term, we have solved the issue with the module above, and this approach (as you suggested in the issue :wink: ) is quite tricky. Also, I guess this module would be difficult to keep alive in v10 due to big changes in standard mrp. However, if you are interested and you think this module is still a valid approach you can take it, I would love to review it and support!

mrp_split/models/mrp_production.py
LoisRForgeFlow: I will attend this.

mrp_split/models/mrp_production.py
LoisRForgeFlow: I will attend this.

LoisRForgeFlow:
This can be closed, as this approach was abandoned in favor of mrp_manufacturing_request.

PR #210, opened 2017-06-13
[9.0][ADD] mrp_production_request

LoisRForgeFlow:
MRP Production Request
===========

This module extends the functionality of Manufacturing to allow you to use
Manufacturing Request (MR) as a previous step to Manufacturing Orders (MO).

Configuration
=============

To configure this module to automatically generate Manufacturing Requests
from procurement orders instead of directly create manufacturing orders yo
need to:

#. Go to the products that you want them to trigger manufacturing orders.
#. Go to the *Inventory* tab.
#. Check the route *manufacture* and the box *Manufacturing Request*.

Usage
=====

To use this module, you need to:

#. Go to *Manufacturing > Manufacturing Requests*.
#. Create a manufacturing request or open a existing one (assigned to you or
   created from a procurement).
#. If you click on *Request approval* button the user assigned as approver
   will added to the thread.
#. If you are the approver you can either click on *Approve* or *Reject*
   buttons.
#. Rejecting a MR will cancel associated procurements and propagate this
   cancellation.
#. Approving a MR will allow to create manufacturing orders.
#. You can manually set to done a request by clicking in the button *Done*.

To create MOs from MRs you have to:

#. Go to approved manufacturing request.
#. Click on the button *Create Manufacturing Order*.
#. In the opened wizard, click on *Compute lines* so you will have a
   quantity proposed for creating a MO. This quantity is the maximum quantity
   you can produce with the current stock available in the source location.
#. Use the proposed quantity or change it and click on *Create MO* at the
   bottom of the wizard.


andhit-r:
Hi @lreficent ,

No mechanism to limit MO? *Create MO* button still appear before MR set done. Even if all wanted qty are produced?

LoisRForgeFlow:
@andhit-r Thanks for reviewing! Regarding your point, I think that as a base module it is ok like this. For instance, if you have to attend a manufacturing request for 19 units and you work in batches of a minimum of 30, what do you do?

I understand your point, but it may be specific to some scenario and I think that customization could be easy to implement. Even a module mrp_production_request_restrict_qty or whatever can be proposed to OCA if that scenario becomes popular.

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,207 @@
...
+    def _onchange_product_id(self):
+        self.product_uom = self.product_id.uom_id
+        self.bom_id = self.env['mrp.bom']._bom_find(
+            product_id=self.product_id.id, properties=[])
+
+    @api.model
+    def _get_mo_valid_states(self):
+        return ['draft', 'confirmed', 'ready', 'in_production', 'done']
+
+    @api.one
```
andhit-r: Use @api.multi

andhit-r:
@lreficent thanks for your explanation.

**Functional Test 1**

1. Create procurement order manually for a product that need MR
2. Run procurement order. Odoo will create MR
3. Create MO for all qty
4. Process MO till done
5. Set MR as done

The procurement order still on running state. I expect done MR would finish the procurement order. Did i miss some step?

andhit-r (changes requested):

andhit-r:
**Functional Test 2**

1. Create/Edit product. Make sure the product (1) has manufacture route active, (2) has manufacture request active, and (2) does not have BoM
mrp-request-2

2. Create procurement order manually

**Result**
Validation error raise
mrp-request-1

**Expected Result:**
Procurement order change into exception state
mrp-request-3

andhit-r:
@lreficent regarding your case:
> For instance, if you have to attend a manufacturing request for 19 units and you work in batches of a minimum of 30

I will:
1. Disable **Make to Order** config for that product
2. Set a reordering rule that has quantity multiple == 30. The MR creation will be base on reordering rule. And it will has 30 multiplier.

Or i just tell user to cancel MR and modify the qty so it will match the minimum batch req in case of manual MR creation.

mrp_production_request/models/mrp_production_request.py
LoisRForgeFlow: api.depends is widely use in conjunction with api.one in odoo standard. Do you think is really needed to avoid this usage of api.one?

LoisRForgeFlow:
@andhit-r 

**Functional test 1**: You are right, need to fix that.

**Functional test 2**: Good point, I'll fix!

**batches scenario**: Yes, that's a good solution, but still not sure about restricting the user capabilities by default. That's why I still prefer to add that restriction in a different module.

Thanks your your advices and review, really appreciate them!

mrp_production_request/models/mrp_production_request.py
andhit-r: Citing from CONTRIBUTING.md

> @api.one method: For v8 is recommended use @api.multi and avoid use @api.one, for compatibility with v9 where is deprecated @api.one.

andhit-r:
@lreficent let's go ahead with unrestricted way. But if you don't mind, put information on README regarding this matter.

mrp_production_request/models/mrp_production_request.py
LoisRForgeFlow: fixed!

LoisRForgeFlow:
@andhit-r Sure, everything is attended now I think.

andhit-r:
Hi @lreficent, thanks for attending all reviews. I created PR https://github.com/Eficent/manufacture/pull/1 that contain few change that i hope make the module easier to extend. Feel free to use/not.

LoisRForgeFlow:
@andhit-r They look good to me, added.

Thanks!

andhit-r (changes requested):
@lreficent functional test 1 & 2 passed. Thank you.

Just one more thing, Codecov report show important part of code is not covered by test https://codecov.io/gh/OCA/manufacture/compare/967c9c743f2941f882f6fd7dbe9476f1b557c9e0...c8b6b79213a7763e8b48aad551dd8094169fd01a/diff. If it's not too much, would you add unittest that covered that part of code?

LoisRForgeFlow:
@andhit-r Yes, it is in my to do list, I'll do ASAP. :wink:

mrp_production_request/README.rst
```diff
@@ -0,0 +1,99 @@
...
+
+#. Go to approved manufacturing request.
+#. Click on the button *Create Manufacturing Order*.
+#. In the opened wizard, click on *Compute lines* so you will have a
+   quantity proposed for creating a MO. This quantity is the maximum quantity
+   you can produce with the current stock available in the source location.
+#. Use the proposed quantity or change it and click on *Create MO* at the
+   bottom of the wizard.
+
+**NOTE:** This module does not restrict the quantity that can be converted
```
elicoidal, line 50: you could use the standard rst note marker:
```.. note::```
(definitely non blocking ;))

File: mrp_production_request/__init__.py [hunk 3]
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
elicoidal, line 3: shouldnot the target be: http://www.gnu.org/licenses/agpl-3.0-standalone.html

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,219 @@
...
+            self.mapped('procurement_id').write({'state': 'done'})
+        return True
+
+    @api.multi
+    def _check_reset_allowed(self):
+        if any([s in self._get_mo_valid_states() for s in self.mapped(
+                'mrp_production_ids.state')]):
+            raise UserError(
+                _("You cannot reset a manufacturing request with "
+                  "manufacturing orders not cancelled."))
```
elicoidal: You cannot reset a manufacturing request if the related manufacturing orders are not canceled.

mrp_production_request/models/product.py
```diff
@@ -0,0 +1,14 @@
...
+
+from openerp import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    mrp_production_request = fields.Boolean(
+        string='Manufacturing Request', help="Check this box to generate "
+        "manufacturing request instead of generating requests manufacturing "
```
elicoidal: instead of generating ~requests~ manufacturing

mrp_production_request/security/mrp_production_request_security.xml
```diff
@@ -0,0 +1,63 @@
...
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="False"/>
+        <field name="perm_create" eval="False"/>
+        <field name="perm_unlink" eval="False"/>
+        <field name="domain_force">['|',('requested_by','=',user.id),
+                                        ('message_partner_ids', 'in', [user.partner_id.id])]</field>
+    </record>
+
+    <record id="mrp_production_request_rule" model="ir.rule">
+        <field name="name">Manufacturing Request User</field>
```
elicoidal, line 52: This role is ambiguous with the one defined on line https://github.com/OCA/manufacture/pull/210/files#diff-61116aedf96b6ea5e449fdf80415ec10R15
Maybe a better naming?

mrp_production_request/wizards/mrp_production_request_create_mo.py
```diff
@@ -0,0 +1,144 @@
...
+    @api.multi
+    def compute_product_line_ids(self):
+        self.product_line_ids.unlink()
+        res = self._prepare_lines()
+        product_lines = res[0]
+        # workcenter_lines = res[1]  # TODO: expand with workcenter_lines
+        for line in product_lines:
+            self.env['mrp.production.request.create.mo.line'].create(
+                self._prepare_product_line(line))
+        # reset workcenter_lines in production order
```
elicoidal: remove dead code

mrp_production_request/wizards/mrp_production_request_create_mo.py
```diff
@@ -0,0 +1,144 @@
...
+class MrpProductionRequestCreateMo(models.TransientModel):
+    _name = "mrp.production.request.create.mo"
+    _description = "Wizard to create Manufacturing Orders"
+
+    @api.multi
+    def compute_product_line_ids(self):
+        self.product_line_ids.unlink()
+        res = self._prepare_lines()
+        product_lines = res[0]
+        # workcenter_lines = res[1]  # TODO: expand with workcenter_lines
```
elicoidal: Remove dead code

elicoidal (changes requested):
Excellent contribution!

mrp_production_request/README.rst
LoisRForgeFlow: didn't know that, let's see how it looks :wink:

mrp_production_request/__init__.py
LoisRForgeFlow: I just followed the OCA's template module

mrp_production_request/security/mrp_production_request_security.xml
LoisRForgeFlow: This is a rule not a group, and this rule refers to that group you pointed.

mrp_production_request/wizards/mrp_production_request_create_mo.py
LoisRForgeFlow: I prefer to keep it to not forget about this in future improvements, or to give some hint to anyone interested to implement this feature.

mrp_production_request/wizards/mrp_production_request_create_mo.py
LoisRForgeFlow: same as above

mrp_production_request/README.rst
LoisRForgeFlow: don't see a significative change in the layout of the README, am I missing something?? @elicoidal

elicoidal: Normally it boxes the note but I do not know whether Odoo interprets it properly (http://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#colored-boxes-note-seealso-todo-and-warnings)

mrp_production_request/__init__.py
elicoidal: Yep: we might think about updating the template indeed

mrp_production_request/security/mrp_production_request_security.xml
elicoidal: oups :+1:

mrp_production_request/wizards/mrp_production_request_create_mo.py
elicoidal: still: keep the TODO and remove the dead code

mrp_production_request/wizards/mrp_production_request_create_mo.py
elicoidal: please remove dead code and add a FIXME or TODO

andhit-r (approved):
LGTM

mrp_production_request/__init__.py
LoisRForgeFlow: Yes, I agree. You can propose a PR, put me in copy please!

mrp_production_request/wizards/mrp_production_request_create_mo.py
LoisRForgeFlow: OK, Done!

mrp_production_request/wizards/mrp_production_request_create_mo.py
LoisRForgeFlow: it's removed now.

mrp_production_request/README.rst
LoisRForgeFlow: Seems that not, I move back to the initial version.

elicoidal: :+1:

mrp_production_request/__init__.py
elicoidal: https://github.com/OCA/maintainer-tools/issues/278

elicoidal (approved):

LoisRForgeFlow:
Thanks for your reviews and help @elicoidal and @andhit-r 

I am squashing a little bit the commits now. - Do not squash more on merge please

LoisRForgeFlow:
Issue https://github.com/OCA/purchase-workflow/issues/440 attended here: 3376a0756692026ebd1760c26a3df04e25858fce

PR #213, opened 2017-07-03
[IMP] quality_control: show action_cancel button in inspections even if it is autogenerated

agaldona:


anajuaristi:
:+1: 
Please merge ASAP. It's causing several issues, when people wants to go back an inspection and they can not because the cancel button is not shown.
It should work exactly same way when inspection is manual or it's created automatically. From a functional point of view there is no reason to work on 2 different ways.

Thank you:
Ana

elicoidal (approved):

LoisRForgeFlow:
Cherry-picked to v9 migration

PR #215, opened 2017-07-11
[9.0][ADD] mrp_production_putaway_strategy

JordiBForgeFlow:
MRP Production Putaway Strategy
===============================

This module allows to apply putaway strategies to the products resulting from
 the manufacturing orders.

The finished products will be placed in the location designated by the putaway
strategy, based on the finished products location that was defined in the
manufacturing order.

Usage
=====

#. Go to `Inventory / Settings' and activate the option 'Advanced routing of
   products using rules'.

#. Define a putaway strategy in the location zone where the finished products
   are supposed to be placed, and indicate the specific sub-location/bin
   where the products should be placed.

#. Create a manufacturing order and indicate the product and the finished
   products location zone. When you confirm the manufacturing order you will
   notice that the finished products location has changed to the putway
   location, and the chatter shows a message indicating that the putaway
   strategy was applied.


JordiBForgeFlow:
@andhit-r you might be interested in this module.
cc @lreficent

LoisRForgeFlow (approved):
Code + functional review :+1:

mrp_production_putaway_strategy/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+The finished products will be placed in the location designated by the putaway
+strategy, based on the finished products location that was defined in the
+manufacturing order.
+
+
+Usage
+=====
+
+#. Go to `Inventory / Settings' and activate the option 'Advanced routing of
```
MiquelRForgeFlow: Change the <code>``</code> by a `'`

MiquelRForgeFlow (approved):
code review :+1:

JordiBForgeFlow:
We found that this is causing issues with MO's that are connected with a subsequent picking.

LoisRForgeFlow:
@jbeficent Tests are updated. This is ready to review again, could you change the label?

PR #216, opened 2017-07-27
[10] migrate mrp mto with stock

florian-dacosta:
Migration of mrp_mto_with_stock module from version 9 (not merged yet)
I did add some improvments, mainly, you can check an option on warehouse so the procurements created are not directly linked to the main MO raw material.

Indeed, we may want to automatically take raw material from stock or from another MO so the forecast stock is always 0, without making a direct link between a raw material MO and the main MO.

For example I create a MO-001 for product A, there is no stock at all for the raw material product B
=> Odoo create a MO-002 for product B.
Then I create another MO-003 for product A, but more urgent than Mconstraint-001 => I want to take the product B from MO-002 instead of waiting the stock from MO-004.
Anyway, this mode offer a lot more flexibility.

I let the classic MTO funcionality by default.
@jbeficent @michaeljohn32 @lreficent


mrp_mto_with_stock/models/mrp_production.py
```diff
@@ -0,0 +1,115 @@
...
+                    procurement = self.env['procurement.order'].search(domain)
+                    if not procurement:
+                        # We have to split the move because we can't have
+                        # a part of the move that have ancestors and not the
+                        # other else it won't ever be reserved.
+                        qty_to_procure = (move.remaining_qty -
+                                          move.reserved_availability)
+                        if qty_to_procure < move.product_uom_qty:
+                            move.do_unreserve()
+                            new_move_id = move.split(
```
LoisRForgeFlow, line 49: I have added this fix to v9, thanks @florian-dacosta

LoisRForgeFlow (commented):
The improvements LGTM. I added a pair of commits to v9 one is an action to mass reserve MOs and the other is a fix based in your improvements here.

Thanks for the work!

florian-dacosta:
Thanks for your comment @lreficent 
For the mass assign commit, I wonder if it should be in this module, I guess it is more generic than that.
For my use case, I needed to assign automatically, that's why I made this module https://github.com/OCA/manufacture/pull/217 at the same time. Maybe the mass assign action should also be in a separated module?
Well, I am not sure, it may become complicated if we have a lot a micro module.

LoisRForgeFlow:
@florian-dacosta Yes, I was not completely sure about adding it there. As you noticed, to use correctly this module is important to assign the earliest possible the MOs in order to allow them to create needed procurements. Yet, in our case it was conflicting with some other features that we don't want to apply immediately in every case. So the mass assign action comes up as a good intermediate solution as for the needs of this module. This is why I finally added the action in the module, not 100% sure of this as said so feel free to not include it here.

LoisRForgeFlow (approved):
LGTM :+1: 

Thanks

wangyueming:
in mrp_production.py

I think 

`                        if qty_to_procure < move.product_uom_qty:
`

should be

`                        if qty_to_procure < move.product_qty:
`

omar7r (approved):

PR #217, opened 2017-07-27
[ADD] mrp_auto_assign

florian-dacosta:
Add a small module to automatically assign the MO once they are created.
One use case is to be used with https://github.com/OCA/manufacture/pull/216
So all subassemblies MO are created on main MO creation

### DIFF
--- a/mrp_auto_assign/README.rst
+++ b/mrp_auto_assign/README.rst
@@ -1,0 +1,57 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+MRP Auto Assign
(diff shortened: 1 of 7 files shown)


LoisRForgeFlow (approved):
Code review LGTM :+1:

omar7r (approved):

pedrobaeza:
Shouldn't this be at MO confirmation instead?

florian-dacosta:
@pedrobaeza From what I saw there is no confirmation step anymore.
There is no more draft state, after creation, it is already confirmed.

There is a `button_plan` action used to create workorders, but it is not used in every case, only if a routing is configured. And the `action_assign` button is available before and after the creation of workorders.

pedrobaeza:
OK, thanks for the explanation. I haven't checked mrp v10 yet, so that's why I didn't know that.

Merging then.

PR #218, opened 2017-07-27
[10.0] Add mrp_progress_button

florian-dacosta:
Small module that ad a 'Mark As started' button on MOs.
The button is only visible if there are no routing linked to the MO, because if you use routing, then you can plan the MO and start workorders, which already do the same.


rami-wafaie (approved):
Code review: no Comment
Testing: it works as it should be
:+1:

rami-wafaie:
@pedrobaeza can you please have a look at it ?

hurrinico:
Code LGTM :+1:

LoisRForgeFlow (approved):
Code review, LGTM

PR #228, opened 2017-10-11
10.0 mig procurement_service

SimoRubi:
Hi,
I've migrated this module from [odoomrp-wip 8.0](https://github.com/odoomrp/odoomrp-wip/tree/8.0/procurement_service) to 10.0, please check


JordiBForgeFlow:
@SimoRubi We refactored for v9 here: https://github.com/Eficent/ddmrp/tree/9.0/procurement_service
Would you mind testing from your side? It looks simpler to me.

JordiBForgeFlow:
@SimoRubi We can then propose to OCA. Or perhaps a mix between the two.

SimoRubi:
@jbeficent Why refactoring from original version? Did you improve any behaviour?

I've looked at your implementation and actually it seems more invasive than the original version: for instance why do you you override the confirm action of sale.order? You could benefit of the built-in logic that in v10 is placed in sale.order.line (and in v8 was in sale.order).

eLBati (approved):

procurement_service/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Alfredo de la Fuente - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+{
+    "name": "Procurement Service",
+    "version": "10.0.1.0.1",
```
LoisRForgeFlow: Why not `10.0.1.0.0` since this is the first version of the module for this repo.

procurement_service/models/procurement_order.py
```diff
@@ -0,0 +1,37 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import models, api
+
+
+class ProcurementOrder(models.Model):
+    _inherit = 'procurement.order'
+
+    @api.multi
+    def _is_procurement_service(self):
+        return self.product_id._is_service_buy_make_to_order() or False
```
LoisRForgeFlow: Why is this method needed? `self.product_id._is_service_buy_make_to_order()` already gives the desired boolean value.

procurement_service/models/procurement_order.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.model
+    def _assign(self):
+        res = super(ProcurementOrder, self)._assign()
+        if not res and self._is_procurement_service():
```
LoisRForgeFlow: see comment above

procurement_service/models/product_product.py
```diff
@@ -0,0 +1,26 @@
...
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def _is_service_buy_make_to_order(self):
+        for product in self:
+            if (product.type == 'service' and len(product.route_ids) == 2 and
+                self.env.ref('stock.route_warehouse0_mto').id in
+                product.route_ids.ids and
+                self.env.ref('purchase.route_warehouse0_buy').id in
```
LoisRForgeFlow, line 18: I'm not sure about this, is there a reason to require both routes in the service definition? I tend to think that just checking that the service has at least one route will be enough here and it will allow other implementations to reuse this module.

procurement_service/models/procurement_order.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class ProcurementOrder(models.Model):
+    _inherit = 'procurement.order'
+
+    @api.multi
+    def _is_procurement_service(self):
+        return self.product_id._is_service_buy_make_to_order() or False
+
+    @api.model
```
LoisRForgeFlow: `_assign` is `@api.multi`. Please respect the method declaration.

LoisRForgeFlow (changes requested):
See my comments below.

procurement_service/__manifest__.py
SimoRubi: done

procurement_service/models/procurement_order.py
SimoRubi: It is compliant with the module as it was in 8.0, so it doesn't break inheriting modules.
Moreover, this allows overriding the behaviour for the order without the necessity to override product's behaviour.

procurement_service/models/procurement_order.py
SimoRubi: You are right, done

procurement_service/models/product_product.py
SimoRubi: This is compliant with the module as it was in 8.0, so it doesn't break inheriting modules.

procurement_service/models/procurement_order.py
LoisRForgeFlow: Could you point me some module which depends on this method? I'm not capable to figure out an use case. Thanks!

SimoRubi: I don't know any module inheriting from this, but since this method already existed in v8 it is possible that some module exists, I prefer to keep this method for the reasons I said above.
Isn't it preferrable to keep inheritance intact as long as it is possible?

LoisRForgeFlow: Well, this is a module from a repo outside the OCA and 2 versions back, now you are proposing it to OCA so we have the chance to refactor as needed to make it useful and generic enough to make it valid for many usecases. Plus, no one has ported the module (not even to 9) and you (the one porting it now) don't know of any module depending on it... the chances are that no one is using it in v10.

summarizing :sweat_smile: : now is the moment to discuss about how to properly implement the module

procurement_service/models/product_product.py
LoisRForgeFlow: again, forget about v8, which are your thought about this condition? I'm not saying it is not a good one, I'm just not sure about it.

procurement_service/models/procurement_order.py
SimoRubi: Looking at things from that perspective yes, this method is useful just because it is a shortcut.
I have found that it is used [here](https://github.com/OCA/manufacture/pull/232/files#diff-2c2d8c588716d0c2dabeaae23db8f46eR14) (almost forgot about it), but since that is my PR too I can change them both.

procurement_service/models/product_product.py
SimoRubi: In this module we are trying to make services mimic the behaviour of stockable products, and as far as I know (and please correct me if I'm wrong, I'm still quite new to OdooWorld):
1. when a stockable product is sold and it has 'buy' route, a purchase order is made unless you have enough in inventory
2. when a stockable product is sold and it has 'mto' route, a purchase order is created

Since for services has no sense talking about inventory, both 1. and 2. are the same.
So yes I can change this and make it return true if any of the two routes is checked.

procurement_service/models/procurement_order.py
LoisRForgeFlow: ok, if you find it useful let's keep it.

procurement_service/models/product_product.py
LoisRForgeFlow: Yes, that's how it works but the thing is this is the expression that defines which services will need a procurement. I was trying to see this module as start or base for others that use services in procurements. In this sense you could find interesting to have a look at: https://github.com/OCA/purchase-workflow/tree/10.0/subcontracted_service and https://github.com/Eficent/ao-odoo/tree/9.0/mrp_production_service. 

For the last one I was considerating to propose it to OCA and depend on this module. But maybe, I'm just over complicating it...

SimoRubi: I think those modules might depend on this, but they should override method `_need_procurement` and not `_is_service_buy_make_to_order`, because the latter defines the way _this_ module decides when services have to be managed differently. 
Actually I'm realizing that since this method is called also in other models it is not very inheriting-friendly, I'll make a new commit later that might help it.

SimoRubi: @lreficent I've committed, now the new method name is more 'anonymous' so it can be easily overridden

LoisRForgeFlow: Great, Thanks @SimoRubi

LoisRForgeFlow (approved):
LGTM :+1: 

Thanks

SimoRubi:
@jbeficent is this ok for you after the changes?
Thanks

SimoRubi:
@jbeficent could you please merge this? Are the changes we and @lreficent performed ok with you?
Thanks

florian-dacosta:
Sorry I am kind of late on this topic, but, the odoomrp procurement_service module was not already migration to oca in version 9, then 10?
The module in version 10 : https://github.com/OCA/purchase-workflow/tree/10.0/subcontracted_service
The v9 PR to let it in the OCA : https://github.com/OCA/purchase-workflow/pull/351

The name changed, but AFAIK, it does the same thing this modules does, just in a slightly différent way.

From What I see, the goal is the same. Also, I don't really see why this would be in manufacture repo?
@jbeficent @lreficent @damdam-s

procurement_service/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
```
florian-dacosta, line 28: You should add a category 'Usage' and explain how to configure the service product so it generates a PO.
Indeed, I don't think it is obvious that you have to fill the mto and buy route, it deserve an explanation.
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst#usage

SimoRubi:
@florian-dacosta I didn't know about https://github.com/OCA/purchase-workflow/tree/10.0/subcontracted_service, I think I'll use that module instead of this one and we might close this PR unless @jbeficent has a different opinion

SimoRubi:
Closing

PR #235, opened 2017-11-07
ADD Group by project_id in mrp.productions

jcarlosmontoya:
ADD Group by project_id in mrp.productions tree view


LoisRForgeFlow (approved):
Code review - it isn't the best option to use string in xpath expression but odoo didn't give you many options... so I think is fine in v8 :+1:

rubencabrera:
Works as expected, surprising it wasn't there before.

damendieta:
It will fail in v9, maybe it's better to think in migrations and use 

`<group position="inside">`
or
`//group/filter[2]`

 instead.

bodedra (approved):

PR #198, opened 2017-05-24
[9.0][MIG] quality_control

LoisRForgeFlow:
Continue the works started in https://github.com/OCA/manufacture/pull/100.

PR #199, opened 2017-05-25
[9.0][ADD] quality_control_issue

LoisRForgeFlow:
This PR respond to the RFC described in https://github.com/OCA/manufacture/issues/196.

Quality Control Issue
=====================

See README file.

------
Depends on:
* [x] : https://github.com/OCA/manufacture/pull/198 
* [x] : https://github.com/OCA/manufacture/pull/214
* [x] : https://github.com/OCA/stock-logistics-workflow/pull/350

PR #236, opened 2017-11-13
9.0 [WIP] [MIG] quality_control_samples

damendieta:
Initial commit of migration, it's working and functional, but no test had been added.

Can someone help whit the test?

Thanks.

------

Depends on:

* [ ] : https://github.com/OCA/manufacture/pull/198


LoisRForgeFlow:
@damendieta I hope you don't mind, I've updated your PR message with the dependency.

damendieta:
@lreficent hi, not at all, thanks for the update.

Runbot and travis are failing because quatlity_control is a dependency but it's on another branch.

How can I make the module available here without making it part of the PR as will happen if I include the module in this branch?

LoisRForgeFlow:
@damendieta Here you have a example of another module depending on quality_control (among others) https://github.com/OCA/manufacture/pull/199/commits/d824263ca36909b72d2bbff5ab23b235ed3b79d2. Note that this is temporary and with testing purposes, in order to be in merging shape we need to merge all the dependencies first.

PR #238, opened 2017-11-15
[9.0] [WIP] quality_control_stock

damendieta:
This is a work in progress, the module is functional but no test had been added.

------

Depends on:

* [ ] : https://github.com/OCA/manufacture/pull/198


LoisRForgeFlow:
@damendieta please squash all the consecutive transifex commits together to smash a little bit the history.

damendieta:
Hi

@lreficent I was thinking about to squash all commits before merge to have only one big commit for the module, that's why I havent's done that in this few commits.

Do you think it's necessary to do this in this state of the PR? can we wait to do it before the merge?

LoisRForgeFlow:
@damendieta yes, it is preferable, because we want to always keep the history and the previous contributions to a module, if we squash everything into one big commit we'll lose all previous contributions which wouldn't be fair for them. In the other hand, it is also good to keep history as small as possible, so that's why transifex commits can be squashed (as they are automatic translation commits) and (right before merging and after all possible reviews are attended) to squash all your commit of the migration.

damendieta:
@lreficent of course I was referring to squash all my commits, not all on the PR, that's why I took the work of keeping the history, even in the ones merged from odoo-mrp, it's important to respect everyones job.

I squashed the commits and I'll squash subsequents commit to have only one commit for all required changes for this PR. 

Daniel.

damendieta:
@lreficent wich is the correct way to squash the changes for the OCA, I've tried several alternatives but all the time I get an undesired result.

You can see this:
https://github.com/damendieta/manufacture/commits/9.0-mig-quality_control_samples

It has added my name to all previous commits, I think this is because it was on another repo, but I don't think it's ok.

Could you please instruct me with the standard OCA way to squash commits.

Thanks.

LoisRForgeFlow:
Hi  @damendieta 

I'll point you [this comment](https://github.com/OCA/partner-contact/pull/507#issuecomment-344597006) of another member of the OCA which is perfectly detailed and explained. Let me know if you have any issue.

PR #242, opened 2017-11-30
Quality control fix

SimoRubi:
Little modifications to the module, please check.
Thanks

### DIFF
--- a/quality_control/views/qc_inspection_view.xml
+++ b/quality_control/views/qc_inspection_view.xml
@@ -86,6 +86,8 @@
                                            groups="product.group_uom"
                                            attrs="{'readonly': [('question_type', '=', 'qualitative')]}" />
                                     <field name="test_uom_category" invisible="1"/>
+                                    <field name="min_value" invisible="1"/>
+                                    <field name="max_value" invisible="1"/>
                                     <field name="valid_values" />
                                     <field name="success" />
                                 </tree>
--- a/quality_control/__manifest__.py
+++ b/quality_control/__manifest__.py
@@ -8,14 +8,10 @@
 
 {
     "name": "Quality control",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Quality control",
     "license": "AGPL-3",
     "author": "OdooMRP team, "
-              "AvanzOSC, "
-              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
-              "Eficent, "
-              "Agile Business Group, "
               "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/manufacture/tree/10.0/quality_control",
     "depends": [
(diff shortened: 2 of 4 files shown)


eLBati (approved):
thanks

quality_control/views/qc_inspection_view.xml
```diff
@@ -86,6 +86,8 @@
                                            groups="product.group_uom"
                                            attrs="{'readonly': [('question_type', '=', 'qualitative')]}" />
                                     <field name="test_uom_category" invisible="1"/>
+                                    <field name="min_value" invisible="1"/>
```
pedrobaeza, line 89: Why adding this?

pedrobaeza (changes requested):
Please change module version number

quality_control/views/qc_inspection_view.xml
SimoRubi: I noticed that [here](https://github.com/OCA/manufacture/pull/242/files#diff-2235da0e3e8b1a36e273264aa3309e16R243) max_value and min_value were always 0, so I figured out that Odoo couldn't retrieve their value because they were not present in the grid.
That's why I added them in the grid

pedrobaeza: That's not the solution: view is independent from ORM code.

SimoRubi: Then I'm sorry but I don't know another way to do this.. could you suggest one?
I'll try to explain myself more clearly:
The inspection record has the correct values in min_value and max_value.
During the editing of the inspection, the depends is triggered but [here](https://github.com/OCA/manufacture/pull/242/files#diff-2235da0e3e8b1a36e273264aa3309e16R243) min_value and max_value are 0, I thought that was because their values are fetched from the view (for the little I know, during editing a temporary record is created gathering values from fields in the view).
Then when I save, the calculations [again here](https://github.com/OCA/manufacture/pull/242/files#diff-2235da0e3e8b1a36e273264aa3309e16R243) are made with correct values, fetched from the inspection record.

SimoRubi: @pedrobaeza in my first answer I didn't mention that the issue is during the edit, is it more clear now why I did this? Is this still the wrong way? 
Thanks for the review

SimoRubi: @pedrobaeza after the clarifications, is this still the wrong way to do it?
Thanks

pedrobaeza: I see something strange in this, but I don't have time to check the inner problem, so if this solves your problem, let's continue with it.

pedrobaeza (approved):

SimoRubi:
@sbidoul how often are packages compiled to https://pypi.python.org/pypi?
I still see [the old version of this module](https://pypi.python.org/pypi/odoo10-addon-quality-control/10.0.1.0.0.99.dev1).
Thanks

sbidoul:
@SimoRubi packages are uploaded nightly. Last night run crashed unfortunately. It should be up tomorrow.

PR #244, opened 2017-12-19
[8.0][FIX] mrp_production_real_cost: calculate of unit cost taking into account real produced quantity

agaldona:


LoisRForgeFlow (approved):
Code review :+1:

StefanRijnhart (approved):

PR #246, opened 2018-01-11
[10.0][MIG] mrp_production_putaway_strategy

LoisRForgeFlow:
Migration to v10.

MRP Production Putaway Strategy
===============================

This module allows to apply putaway strategies to the products resulting from
the manufacturing orders.

The finished products will be placed in the location designated by the putaway
strategy (if they do not have another destination move), based on the
finished products location that was defined in the manufacturing order.

Configuration
=============

To configure a putaway strategy follow the next steps:

#. Go to 'Inventory / Settings' and activate the option 'Advanced routing of
   products using rules'.
#. Define a putaway strategy in the location zone where the finished products
   are supposed to be placed, and indicate the specific sub-location/bin
   where the products should be placed.

Usage
=====

To use this module proceed as follows:

#. Create a manufacturing order and indicate the product and the finished
   products location zone.
#. Confirm the manufacturing order.
#. You will notice that the finished products location has changed to the
   putaway location, and the chatter shows a message indicating that the
   putaway strategy was applied.


StefanRijnhart:
Looks like you need to update the test

LoisRForgeFlow:
@StefanRijnhart Thanks! Fixed, rebased and ready to review.

StefanRijnhart (approved):
Thanks! I checked the differences in the code. Agreed that the second test had become redundant because of the refactoring of the MRP module in Odoo 10.0.

mrp_production_putaway_strategy/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
MiquelRForgeFlow: `.png`

mrp_production_putaway_strategy/README.rst
```diff
@@ -0,0 +1,80 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
```
MiquelRForgeFlow: https

mrp_production_putaway_strategy/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "MRP Production Putaway Strategy",
+    "summary": "Applies putaway strategies to manufacturing orders for "
+               "finished products.",
+    "version": "10.0.1.0.0",
+    "author": "Eficent, "
+              "Odoo Community Association (OCA)",
+    "website": "http://www.eficent.com",
```
MiquelRForgeFlow: /OCA/manufacture

MiquelRForgeFlow (approved):

mrp_production_putaway_strategy/__manifest__.py
LoisRForgeFlow: everything fixed, thanks!

Issue #248, opened 2018-01-30
[11.0] mrp_bom_equivalent


max3903:
This module considers all the the products in the same product category to be equivalent and interchangeable in a Manufacturing Order, using the product priority. The user can specify non-equivalent products to a part on a BOM.

# Usage
- Go to Manufacturing > Master Data > Bill of Materials
- Create or update a BOM to check the box "Use Equivalences"
- Set the non-equivalent products among the products of the same category

# Dependencies
- https://github.com/OCA/product-attribute/issues/322

bodedra:
@max3903 

PR #250 is ready for review.

PR #250, opened 2018-02-02
[ADD] mrp_bom_equivalent

bodedra:
#248 

Depends on https://github.com/OCA/product-attribute/pull/327


mrp_bom_equivalent/models/mrp.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2017 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MRPBoMLine(models.Model):
+    _inherit = 'mrp.bom.line'
+
+    is_equivalences = fields.Boolean(
```
max3903: @bodedra `use_equivalences`

mrp_bom_equivalent/models/mrp.py
```diff
@@ -0,0 +1,19 @@
...
+        string='Use equivalences'
+    )
+    equivalences_product_ids = fields.Many2many(
```
max3903: `nonequivalent_product_ids`

max3903 (changes requested):

mrp_bom_equivalent/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2017 - TODAY, Open Source Integrators
```
max3903: @bodedra `2018` ;)

max3903 (changes requested):

max3903:
@bodedra You need to add `product-attribute` in the `oca_requirements.txt` file

mrp_bom_equivalent/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+
+{
+    "name": "MRP BoM Equivalences",
+    "summary": "Specify non-equivalent products to a part on a BOM",
+    "version": "11.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "category": "MRP",
+    "website": "http://www.opensourceintegrators.com",
+    "depends": ["mrp"],
```
max3903: @bodedra It should depend on `product_priority` also.

max3903 (approved):

mrp_bom_equivalent/models/product.py
```diff
@@ -0,0 +1,38 @@
...
+            category_id = self.browse(
+                self._context.get('nonequivalent_product_id')).categ_id
+            domain +=\
+                [('categ_id', '=', category_id.id),
+                 ('id', '<>', self._context.get('nonequivalent_product_id'))]
+        return super(ProductProduct, self).search_read(domain=domain,
+                                                       fields=fields,
+                                                       offset=offset,
+                                                       limit=limit,
+                                                       order=self.priority)
```
guewen: Why is the order overriden in any case?

mrp_bom_equivalent/models/product.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.model
+    def search_read(self, domain=None, fields=None, offset=0, limit=None,
+                    order=None):
+        if 'nonequivalent_product_id' in self._context:
+            category_id = self.browse(
+                self._context.get('nonequivalent_product_id')).categ_id
+            domain +=\
+                [('categ_id', '=', category_id.id),
+                 ('id', '<>', self._context.get('nonequivalent_product_id'))]
```
guewen: `!=`

mrp_bom_equivalent/models/product.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        if 'nonequivalent_product_id' in self._context:
+            category_id = self.browse(
+                self._context.get('nonequivalent_product_id')).categ_id
+            recs = self.browse()
+            recs = self.search(
+                [('categ_id', '=', category_id.id),
+                 ('id', '<>', self._context.get('nonequivalent_product_id')),
```
guewen: `!=`

mrp_bom_equivalent/models/product.py
max3903: @guewen To have the top priority ones first

guewen: @max3903 it means if a developer / rpc client explicitely calls this method with a different order, it will be ignored, why not `order or self.priority`?

max3903: @guewen ok

guewen (approved):
Thanks!

LoisRForgeFlow:
Hi @bodedra @max3903 @guewen 

I am trying to understand how to use this module. But I'm not able to figure it out. How the substitution for a equivalent product happens on the Manufacturing Order?

Imagine the following scenario:
* My BoM to built Product A uses a only component, "Component 1", and I set "use equivalences" for that line.
* I have a product category "components" in which I have component 1, component 2 and component 3.
* I create a MO for product A, the consumed material stock move is for component 1.

How can I use the component 2 to build product A?

Thanks in advance and sorry for being late here :smile: 

EDIT: cc @jgrandguillaume @jbeficent

max3903:
@lreficent The idea is to have another module to use the equivalence defined here.

In our case, we plan to check DDMRP buffer level when the MO is created:
* If the component 1 buffer is red, then we check the buffer for component 2. 
* If the component 2 buffer is green, the MO will consume component 2.

mrp_bom_equivalent/models/product.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.model
+    def search_read(self, domain=None, fields=None, offset=0, limit=None,
+                    order=None):
+        if 'nonequivalent_product_id' in self._context:
+            category_id = self.browse(
+                self._context.get('nonequivalent_product_id')).categ_id
+            domain +=\
+                [('categ_id', '=', category_id.id),
+                 ('id', '!=', self._context.get('nonequivalent_product_id'))]
+        order = order or self.priority
```
LoisRForgeFlow, line 35: self.priority on an api.model will return 0. Whats the purpose of this?

LoisRForgeFlow:
@max3903 Thanks for the clarification, maybe it would be good to think on an implementation to use this without ddmrp also. Maybe to give the option to use a different product when manufacturing? Use the product priority you already added?

max3903:
@lreficent That's why we splitted the feature in 2 modules (1 module for the definition and 1 for each use case).

Our customer chose the one described here https://github.com/OCA/ddmrp/issues/10

LoisRForgeFlow:
@max3903 Good, thanks for pointing that out!

PR #223, opened 2017-09-15
[9.0][ADD] mrp_bom_location

LoisRForgeFlow:
MRP BOM Location
================

This module adds the location field to the Bill of Materials and its
components. This may be useful to distinguish between different BoMs for the
same product or to highlight the preferred locations to fetch the
components from.

PR #254, opened 2018-02-12
[MIG] 11.0 mig mrp bom location

mpanarin:
Port to Odoo 11 of this module:
https://github.com/OCA/manufacture/pull/223

depends on
- [x] odoo/odoo#23986


LoisRForgeFlow:
Maybe in v11 we might use the new picking_id field, which aims to cover the same need. Thus, maybe there is no need to migrate this one.

mpanarin:
@lreficent @jbeficent 
there is only picking_type_id, which won't be of much help

max3903 (approved):
LGTM

LoisRForgeFlow:
@mpanarin @jbeficent yes, sorry I meant picking_type_id, from there you have access to the at least the WH where the MO will be executed, so you can distinguish which BoM should apply regarding the picking_type_id (also new in v11) of the MO. Also, you could potentially have more picking types for the same WH with different default locations if needed.

yvaucher:
Just for the records if someone is tempted to port this module to v10.0, the picking_type_id field is already there in v10.0.

https://github.com/odoo/odoo/blob/de62e33618211fd754ad6e372ac93805cce58770/addons/mrp/models/mrp_bom.py#L54

yvaucher:
@lreficent not sure about the needs of this module however this one has the granularity on the BOM line. Thus picking_type_id on BOM might be not enough. What could be done though is to get the default location from the picking_type_id. What do you think?

jgrandguillaume:
Hi,

My understanding here is that we need the location in BoM to compute several things in DDMRP. It'll be really useful when working in a multi-warehouse env of course.

Does it make sens that a component in that case would be produce in WH 1 where the final product will be produce in WH 2 ? I guess this case can happen, but is more likely to be an exception.

If think I agree with @lreficent , sounds more reasonable to base everything on the warehouse stock location of the picking type linked to a bom location_id. Easier info to maintain the infos as well.

@jbeficent Can you confirm this ? 

If this is confirmed, can I suggest that this modules mrp_bom_location makes the fields "location_id" on bom and bom.line a stored computed/related field based on the picking_type_id.warehouse_id.lot_stock_id 

Doing that will ease the migration path for DDMRP's related modules, what do you say ?

Regards,

Joël

JordiBForgeFlow:
We discussed with @lreficent and we agree that this would be a good approach! I propose:
mrp.bom.line
location_id = fields.Many2one('stock.location', relation='bom_id.picking_type_id.default_location_src_id', store=True)

mrp.bom
location_id = fields.Many2one('stock.location', relation='picking_type_id.default_location_dest_id', store=True)

This would follow a simimar approach to how the locations for raw materials and finished products are being proposed in the manufacturing order:
https://github.com/odoo/odoo/blob/11.0/addons/mrp/models/mrp_production.py#L299

mrp_bom_location/models/mrp_bom.py
```diff
@@ -0,0 +1,22 @@
...
+    location_id = fields.Many2one(
+        related='picking_type_id.default_location_dest_id',
+        store=True,
+    )
+
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    location_id = fields.Many2one(
```
grindtildeath, line 25: If these fields are related, shouldn't these be readonly here ?

mrp_bom_location/models/mrp_bom.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    location_id = fields.Many2one(
```
grindtildeath, line 10: If these fields are related, shouldn't these be readonly here ?

grindtildeath (commented):
Just a question. 
cc @lreficent @jbeficent

mrp_bom_location/models/mrp_bom.py
jgrandguillaume: Yes..

mrp_bom_location/models/mrp_bom.py
jgrandguillaume: Yes again :)

mrp_bom_location/README.rst
```diff
@@ -0,0 +1,65 @@
...
+
+The location appears in the BOM Structure Report.
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Manufacturing > Bill of Materials*.
```
pedrobaeza, line 22: Add 
```
#. Pick or create one of them.
#. You will see a new field to fill called "Location".
```

mrp_bom_location/README.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Lois Rilo <lois.rilo@eficent.com>
```
pedrobaeza, line 50: Add yourself as contributor

jgrandguillaume:
Hi,

This has been fixed and could now be merged IMO.

Regards,

jgrandguillaume (approved):
LGTM

mrp_bom_location/README.rst
pedrobaeza: Add also instructions about the bom list report

mrp_bom_location/models/mrp_bom.py
pedrobaeza: I see confusing for the user to use this related field as location. Maybe it's better to decouple from the picking type? And due to picking type is not used for any other thing. What about if you want to use a location that hasn't got any picking type related to?

mrp_bom_location/report/bom_structure.py
```diff
@@ -0,0 +1,64 @@
...
+        if uom:
+            if uom != record.bom_id.product_uom_id:
+                qty = uom._compute_quantity(qty, record.bom_id.product_uom_id)
+            child['pqty'] = (record.product_qty * qty) / qty_per_bom
+        else:
+            # for the first case, the ponderation is right
+            child['pqty'] = (record.product_qty * qty)
+        return child
+
+    def get_children(self, records, level=0):
```
pedrobaeza, line 32: You are completely rewritten this method from upstream, so you are destroying inheritability. Please don't do this and try to reimplemented calling super and adding later data.

pedrobaeza (changes requested):

mrp_bom_location/models/mrp_bom.py
grindtildeath: After doing some tests I tend to agree with @pedrobaeza that it's confusing.

Especially with such an help message.

image

mrp_bom_location/report/bom_structure.py
mpanarin: the problem here is that I really need to inherit the inner recursive method. But as the original author didn't think about extensibility of his awesome code there -> the only way I see is to reimplement it to add a single field.

mpanarin:
@jgrandguillaume @grindtildeath @pedrobaeza i updated location for bom to be simple Many2one, but also added onchange for picking_type_id to set it.

pedrobaeza:
Thanks for the changes. Good balance between both.

About the inheritance, please try to make your best for keeping inheritability.If not, at least put a common module to be used as hook one for other OCA modules. Example: https://github.com/OCA/manufacture/tree/8.0/mrp_hook

pedrobaeza:
And check Travis error :wink:

jgrandguillaume:
Hi @pedrobaeza ,

Would you update your review so we may be able to merge ?

Thanks,

Joël

pedrobaeza:
But the method is still totally overwritten...

LoisRForgeFlow:
@pedrobaeza see  @mpanarin's [comment](https://github.com/OCA/manufacture/pull/254#discussion_r175063997) and give him some insight on how to do it please.

pedrobaeza:
@lreficent, I would go for one of these options:

1. Make a PR to Odoo adding a hook method for preparing values for the report. As this is for v11 and with an OPW, it will probably be merged.
2. If not, add a hook module that makes the same, and inherit from that one all the modules you want to modify the list. This way, we have a common ground to be based on.

mpanarin:
pr to odoo is open
https://github.com/odoo/odoo/pull/23986
when it gets merged, I'll update this one

pedrobaeza:
Thanks

LoisRForgeFlow:
@mpanarin Finally the Odoo PR got merged, could you update this one? Thanks!

LoisRForgeFlow:
@mpanarin ping

LoisRForgeFlow:
@mpanarin can you please merge this PR adapting to the new hook added by odoo: https://github.com/mpanarin/manufacture/pull/1?

Thanks!

LoisRForgeFlow:
No answer. Superseded by https://github.com/OCA/manufacture/pull/285.

PR #260, opened 2018-03-29
[IMP] Products in subcategories are equivalents and can be excluded

max3903:


bodedra (approved):
LGTM. :+1:

agyamuta (approved):
LGTM...

mrp_bom_equivalent/models/product.py
```diff
@@ -30,7 +30,7 @@ def search_read(self, domain=None, fields=None, offset=0, limit=None,
             category_id = self.browse(
                 self._context.get('nonequivalent_product_id')).categ_id
             domain +=\
-                [('categ_id', '=', category_id.id),
+                [('categ_id', 'child_of', category_id.id),
                  ('id', '!=', self._context.get('nonequivalent_product_id'))]
         order = order or self.priority
```
LoisRForgeFlow: Can you check this line? it doesn't make sense for me.

LoisRForgeFlow: self.priority on an api.model will return 0. should it be `"priority"`?

PR #267, opened 2018-04-26
Update string in smart buttons

bealdav:
smart

Better display with 'Component' only in string to looks like other smart buttons

@tonygalmiche @hveficent


flotho (approved):
Thanks for your proposal.
LGTM

bealdav:
ping @hveficent

JordiBForgeFlow (approved):

PR #272, opened 2018-05-10
[11.0] [FIX] mrp_production_service

HviorForgeFlow:
Production services couldn't be used as unitary prices per bom quantities.

CC @jbeficent @lreficent

### DIFF
--- a/mrp_production_service/models/mrp_production.py
+++ b/mrp_production_service/models/mrp_production.py
@@ -8,31 +8,27 @@ from odoo import api, models
 class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
-    @api.model
-    def _prepare_service_procurement_values(self, production):
-        location = production.location_src_id
+    @api.multi
+    def _prepare_service_procurement_values(self):
+        self.ensure_one()
+        location = self.location_src_id
         return {
-            'company_id': production.company_id,
-            'date_planned': production.date_planned_start,
+            'company_id': self.company_id,
+            'date_planned': self.date_planned_start,
             'warehouse_id': location.get_warehouse(),
-            'group_id': production.procurement_group_id,
+            'group_id': self.procurement_group_id,
         }
 
     @api.model
-    def _create_service_procurement(self, line):
-        data = self._prepare_service_procurement(line)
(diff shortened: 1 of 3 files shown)


mrp_production_service/models/mrp_production.py
```diff
@@ -9,13 +9,13 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     @api.model
-    def _prepare_service_procurement_values(self, production):
-        location = production.location_src_id
+    def _prepare_service_procurement_values(self):
+        location = self.location_src_id
```
LoisRForgeFlow: then this is not anymore api.model but api.multi

mrp_production_service/models/mrp_production.py
```diff
@@ -24,15 +24,15 @@ def _create_service_procurement(self, line):
...
-        values = self._prepare_service_procurement_values(production)
+    def _action_launch_procurement_rule(self, line):
+        values = self._prepare_service_procurement_values()
 
         name = '%s for %s' % (line.product_id.name,
-                              production.name)
+                              self.name)
         self.env['procurement.group'].sudo().run(
-            line.product_id, line.product_qty,
+            line.product_id, self.product_qty,
```
LoisRForgeFlow: if the MO is for 3 and the parent/child ratio for product/service is 1:2 in the BoM, this line will procure for 3 (as you use `self.product_qty`) units of service instead of 6, am I wrong?

LoisRForgeFlow (commented):
is `_create_service_procurement` dead code? (mrp_production.py @ L22)

mrp_production_service/models/mrp_production.py
HviorForgeFlow: Done, is it right this way?

mrp_production_service/models/mrp_production.py
HviorForgeFlow: It seems to be fixed, I also update test

JordiBForgeFlow (approved):

JordiBForgeFlow:
@lreficent is also good. Now merging.

PR #278, opened 2018-05-29
[11.0][REW] mrp_mto_with_stock

LoisRForgeFlow:
Changes: 
* logic for reservable stock moved to _adjust_procure_method. This way it will procure as needed on creation without needing to assign the MO.
 * hook for mto_with_stock condition.
 * fix bug with unit_factor and add tests for it.
 * update module dependencies.

cc @jbeficent @florian-dacosta @bodedra

### DIFF
--- a/mrp_mto_with_stock/__manifest__.py
+++ b/mrp_mto_with_stock/__manifest__.py
@@ -9,11 +9,15 @@
     "author": "John Walsh, Eficent, Odoo Community Association (OCA)",
     "website": "https://odoo-community.org/",
     "category": "Manufacturing",
-    "version": "11.0.1.0.0",
+    "version": "11.0.2.0.0",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["stock", "sale", "purchase", "mrp"],
+    "depends": [
+        "stock",
+        "mrp",
+        "stock_available_unreserved",
+    ],
     "data": [
         'views/product_template_view.xml',
         'views/stock_warehouse.xml',
--- a/mrp_mto_with_stock/models/mrp_production.py
+++ b/mrp_mto_with_stock/models/mrp_production.py
@@ -1,20 +1,19 @@
-# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017-18 Eficent Business and IT Consulting Services S.L.
 # Copyright 2015 John Walsh
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
 from odoo.exceptions import UserError
 import copy
-import logging
-_logger = logging.getLogger(__name__)
 
 
 class MrpProduction(models.Model):
     _inherit = 'mrp.production'
 
-    def _get_procurement_group_data(self, move):
-        return {'partner_id': move.partner_id.id,
-                'name': '{0}:{1}'.format(self.name, move.product_id.name)}
+    def _mto_with_stock_condition(self, move):
+        """Extensibility-enhancer method for modifying the scenarios when
+        MTO/MTS method should apply."""
+        return move.location_id in move.product_id.mrp_mts_mto_location_ids
 
     @api.multi
     def action_assign(self):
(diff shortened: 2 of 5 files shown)


LoisRForgeFlow:
The bug I refer to here happened after producing a MO with splitted moves. See picture attached:
image

This bug makes the module nonfunctional so I think this is major fix.

florian-dacosta:
Hello @lreficent 
Thanks for your work.
I did a functional review, and it is ok for me. I just wonder why changing the logic to _adjust_procure_method
Now, depending on the 'mode' (`mto_with_no_move_dest_id`) the procurement will be created at creation or at assignation, it is a bit strange.
And I notice that in the method `_adjust_procure_method`, you actually reserve the raw material automatically, so the user don't have a choice but to reserve automatically its MO with this mode.

In this case, why not use mrp_auto_assign and leave the logic in `check_assign` method?
This way, the user of the module can still choose if he wants to auto-assign the MOs, no?

LoisRForgeFlow:
Hi @florian-dacosta

Thanks for your comments:

> In this case, why not use mrp_auto_assign and leave the logic in check_assign method?

My problem is that I'm not comfortable with needing to auto-reserve to make this module work as I expect (auto-generete a MTO order chain but using residual stock). I don't want to auto-reseve every MOs, MOs that are MTS would also be reserved, and that's not what I want.

> And I notice that in the method _adjust_procure_method, you actually reserve the raw material automatically

Yes, because I still need to reserve the residual quantities, if we don't, a new order for the same product would reduce again its MTO part by the residual quantity (thus you would considere the residual qty more than once leading to inconsistencies on planning).

> Now, depending on the 'mode' (mto_with_no_move_dest_id) the procurement will be created at creation or at assignation, it is a bit strange.

That's true, but I didn't want to move the other mode because I know you are using it, and you may not agree with the change.

florian-dacosta:
> Yes, because I still need to reserve the residual quantities, if we don't, a new order for the same product would reduce again its MTO part by the residual quantity (thus you would considere the residual qty more than once leading to inconsistencies on planning).

Wouldn't it be possible to check if a linked RFQ/MO exists already?
I am asking, because, the system seems a bit fragile. If I understand correctly, if you have 2 products in stock, and create a MO needing 3 products, it will create a sub MO for 1 product and reserve the 2 others.
If for any reason, you un-reserve the main MO, and re-reserve it later, it will create another sub MO?

> That's true, but I didn't want to move the other mode because I know you are using it, and you may not agree with the change.

If we go with this refactore, I think it would be better to move the second logic too for consistency. It is not a problem for me.
(but for this second logic, we don't need to reserve anything as we just look at the forecast stock)

LoisRForgeFlow:
> if you have 2 products in stock, and create a MO needing 3 products, it will create a sub MO for 1 product and reserve the 2 others.
If for any reason, you un-reserve the main MO, and re-reserve it later, it will create another sub MO?

Yes, that's correct with this approach and also with the auto assign approach. But any fix that comes to my mind seems like a lot of overhead or adding unneeded complexity. At the end, this is no more than a patch due tof the lack of a real MRP implementation in Odoo (we are actually working on that now). However, I think this module can still be useful in simple environments.

> (but for this second logic, we don't need to reserve anything as we just look at the forecast stock)

That a good point, then I'll do it asap if you agree.

florian-dacosta:
It would be great!
Thanks!

LoisRForgeFlow:
@florian-dacosta I just realized that this [relies on the virtual qty](https://github.com/OCA/manufacture/pull/278/files#diff-aea23d736e53e9d08e5df75726470609R131) and the virtual quantity do not consider draft moves and in the method `_adjust_procure_method` they are still in draft. Maybe that part should remain in `action_assign` after all.

florian-dacosta (approved):

florian-dacosta:
Let leave it this way for now then.

oerp-odoo:
So was it decided to leave stock reservation modification in `action_assign`, or use this new approach (as on PR) where reservation happens when MO is just created?

LoisRForgeFlow:
@oerp-odoo as it is now, it will happens on creation for the default mode.

oerp-odoo:
@lreficent but I did not understand the reasoning, why it should be done on creation? To me it looks like this module should respect standard way of reservation. Maybe extra module could change that by calling `action_assign` on create?

Is it done, because of MTO involvement?

P.S. I also noticed another thing (not related with this PR, but maybe someone could answer me that. If not, ignore it:)). If you use standard MTO and do this workflow -> create SO for manufactured product (with manufacture route) and then have materials with buy and MTO routes. When procurement rules trigger, materials (POs in this case), it will specify `Origin` (or source) as sale order number. But using this module, it will specify related MO as origin, instead of SO. 
Shouldn't it follow same approach and specify origin as SO if SO is related with MO and materials' POs?

LoisRForgeFlow:
@oerp-odoo 
> but I did not understand the reasoning, why it should be done on creation?

My problem is that I'm not comfortable with needing to auto-reserve to make this module work as I expect (auto-generete a MTO order chain but using residual stock). I don't want to auto-reseve every MOs, MOs that are MTS would also be reserved, and that's not what I want.

Explained in other words: without this fix, I have to use mrp_auto_assign module to trigger the automatic creation of the subsequent MOs. I don't want that because I could have products that follow this MTO/MTS rule and other with pure MTS (case in which I wouldn't want auto-reservation).

So, without mrp_auto_assing in the equation, the module works exactly the same as before but now the creation of subsequent MOs is automatic.

Does this sound good to you?

> (...) But using this module, it will specify related MO as origin, instead of SO.

You are right, I'll look into that.

LoisRForgeFlow:
@oerp-odoo I fixed the origin issue, if the MO has an origin it uses it. If not, defaults to the one in the move (usually the initial MO number). I remove the previous origin logic as I actually don't like it too much (I think using the MO number is enough information), but I can readd it if someone think it is useful.

EDIT: words

oerp-odoo:
Yeah it makes sense now. Origin looks good now too.

oerp-odoo (approved):

LoisRForgeFlow:
Squashed all commits and ready to merge.

Issue #279, opened 2018-06-15
[11.0] mrp_mto_with_stock consumed materials are doubled.


oerp-odoo:
I don't know If I understand configuration of this module or intention, but it looks like it messes up manufacturing order consumed materials when used.

I'm using default mode (where nothing is enabled on warehouse to have forecast quantity mode). My product that is consumed in MO has `wh/stock` location set as `MTO/MTS` location. And uses `Buy` route only.

It seems it works correctly till materials are needed to be consumed (taking quantity from stock, ordering missing quantity, reserving needed quantity). When I start producing MO, then consumed materials stock moves duplicate.

For example If I manufacture a product where I need 5 units of material A and I have 2 units in stock and 3 are ordered. Then when I receive remaining materials, after pressing produce button, I see such stock moves (instead of 2 moves for 2 units and 3 ordered units):

```
06/15/2018 14:48:57	MO/00015	T3	WH/Stock	Virtual Locations/Production	3.000	Done

06/15/2018 14:48:57	MO/00015	T4 MF	Virtual Locations/Production	WH/Stock	5.000	Done

06/15/2018 14:48:57	MO/00015	T3	WH/Stock	Virtual Locations/Production	2.000	Done

06/15/2018 14:48:57	MO/00015	T3	WH/Stock	Virtual Locations/Production	3.000	Done

06/15/2018 14:48:57	MO/00015	T3	WH/Stock	Virtual Locations/Production	2.000	Done
```
Which means you use more materials than were required for an MO.

P.S. You can see my attached video how I reproduced this behavior.

[mrp-mto-with-stock-bug.ogv.zip](https://github.com/OCA/manufacture/files/2105959/mrp-mto-with-stock-bug.ogv.zip)

oerp-odoo:
I have created a fix for this issue: https://github.com/OCA/manufacture/pull/280

LoisRForgeFlow:
This can be closed too I think as it is solved.

Thanks!

oerp-odoo:
Yes. Forgot to close it:)

PR #280, opened 2018-06-15
[FIX] mrp_mto_with_stock: action_assign with split

oerp-odoo:
When stock move is split in manufacturing order, its `unit_factor` must
be updated to reflect quantity change. Otherwise, it would consume
materials (stock moves) as before split, meaning it will consume either
more or less than was actually required to consume to produce MO.

To update `unit_factor`, we use formula
`new_stock_move_quantity / mo_quantity` (
updated quantity divided by quantity of goods to produce on MO).

Should solve this issue: https://github.com/OCA/manufacture/issues/279

### DIFF
--- a/mrp_mto_with_stock/README.rst
+++ b/mrp_mto_with_stock/README.rst
@@ -31,7 +31,7 @@ To configure this module, you need to:
 If you want to use the second mode, based on forecast quantity
 
 #. Go to the warehouse you want to follow this behaviour.
-#. In the view form go to the tab *Warehouse Configuration* and set the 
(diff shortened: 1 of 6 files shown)


LoisRForgeFlow:
Hi @oerp-odoo 

There is already an open PR solving this: https://github.com/OCA/manufacture/pull/278

Could you please review it?

oerp-odoo:
@lreficent Oh, somehow I didn't notice that there was PR already.. I'll take a look.

oerp-odoo:
PR: https://github.com/OCA/manufacture/pull/278 already fixed this issue.

PR #288, opened 2018-07-13
10.0 fix mrp production request date planned

SimoRubi:
PR made to clear runbot's warnings.
In module mrp_production_request, field `date_planned` should be named `date_planned_start` in order to be the counterpart of `mrp.production.date_planned_start`.
Plus other minor fixes to `mrp_production_putaway_strategy` and `mrp_production_service`


SimoRubi:
I've done this basically to fix runbot's warnings, @OCA/manufacturing-maintainers could you please check?
Thanks

mrp_production_request/models/mrp_production_request.py
```diff
@@ -108,6 +113,19 @@ def _compute_manufactured_qty(self):
...
         required=True, copy=False, default='draft')
+    procurement_group_id = fields.Many2one(
+        string='Procurement Group',
+        comodel_name='procurement.group',
+        copy=False)
+    procurement_ids = fields.One2many(
+        string='Related Procurements',
+        comodel_name='procurement.order',
+        inverse_name='mrp_production_request_id')
+    propagate = fields.Boolean(
```
LoisRForgeFlow, line 124: Why do you add these three fields? They aren't used anywhere, are they?

mrp_production_request/models/mrp_production_request.py
```diff
@@ -91,9 +91,14 @@ def _compute_manufactured_qty(self):
...
         readonly=True, states={'draft': [('readonly', False)]})
     description = fields.Text('Description')
-    date_planned = fields.Date(
-        string='Planned Date', required=True, track_visibility='onchange',
-        default=fields.Date.context_today)
+    date_planned_start = fields.Datetime(
+        'Deadline Start', copy=False, default=fields.Datetime.now,
+        index=True, required=True,
+        states={'confirmed': [('readonly', False)]}, oldname="date_planned")
+    date_planned_finished = fields.Datetime(
```
LoisRForgeFlow, line 98: This is making me think :thinking: Maybe the wizard should be extended with a date field that is defaulted to this but can be changed. In such case I think that makes more sense to keep the `date_planned` as the only date field in the manufacture request, i.e. the date in which all the quantity must have been made. 

What do you think?

mrp_production_request/models/mrp_production_request.py
SimoRubi: Hi, Thanks for your review!
I did this PR to fix warnings in runbot, the absence of these 3 fields generated [this warning](https://travis-ci.org/OCA/manufacture/jobs/402942342#L1340) because the procurement order creates the `mrp_production_request` in [this way](https://github.com/SimoRubi/manufacture/blob/2c743e97f1774a4d994ae71d6fad3f5a4493af65/mrp_production_request/models/procurement.py#L18): among others it gives a value to these 3 of fields that, before this PR, were in `mrp.production` but not in `mrp.production.request`.

mrp_production_request/models/mrp_production_request.py
SimoRubi: I know quite little of this module, but I'm starting to think that the `mrp.production.request` should be as similar as possible to `mrp.production` (also in order to solve the *issue* I told you about [here](https://github.com/OCA/manufacture/pull/288/files/2c743e97f1774a4d994ae71d6fad3f5a4493af65#r203273104)), so I'd leave the fields of `mrp.production.request` as they are int this PR.
Anyway your suggestion is a functionality that we can still add: the only difference is that `date_planned` of the wizard will default to `date_planned_start` of the request, would this be ok with you?

mrp_production_request/models/mrp_production_request.py
LoisRForgeFlow: Ok, I got you. Then, please add those new fields in [`_prepare_manufacturing_order`](https://github.com/SimoRubi/manufacture/blob/2c743e97f1774a4d994ae71d6fad3f5a4493af65/mrp_production_request/wizards/mrp_production_request_create_mo.py#L84) method.

mrp_production_request/models/mrp_production_request.py
LoisRForgeFlow: That's fine, don't worry about the new functionality I note it as a possible improvement. 

Regarding the renamed field (date_planned -> date_planned_start), could you please add a migration script?

SimoRubi: Done in https://github.com/OCA/manufacture/pull/288/commits/f1273cba85ae97c0f11e742a964c942e2796a273, I don't know if I did it correctly for the migration, I never did that before, could you please check?

mrp_production_request/models/mrp_production_request.py
SimoRubi: Done in https://github.com/OCA/manufacture/pull/288/commits/f1273cba85ae97c0f11e742a964c942e2796a273

mrp_production_request/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "summary": "Allows you to use Manufacturing Request as a previous "
                "step to Manufacturing Orders for better manufacture "
                "planification.",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
```
LoisRForgeFlow: First of all I think you should bump the version to `10.0.1.1.0` not `10.0.1.0.1`. See the [contributing reference](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers).

mrp_production_request/migrations/10.0.1.0.0/pre-migration.py
```diff
@@ -6,7 +6,7 @@
 
 field_renames = [
     ('mrp.production.request', 'mrp_production_request', 'product_uom',
-     'product_uom_id'),
+     'product_uom_id', 'date_planned', 'date_planned_start'),
```
LoisRForgeFlow: Then, this should be in a separate migration folder for version `10.0.1.1.0` (as migration need to happen for people that has already the module installed in v10 and is just upgrading).

LoisRForgeFlow (commented):
Thanks! See my remarks below.

SimoRubi:
@lreficent done your remarks in https://github.com/OCA/manufacture/pull/288/commits/9f5aa9a3a1bdabf1ec6a353e5a2d53fa92e051bf, hope it's ok now; migrations logic is still a bit obscure to me

LoisRForgeFlow (approved):
Looks great now! :+1:

Thanks!

SimoRubi:
@OCA/manufacturing-maintainers can someone please check?
Thanks

SimoRubi:
@OCA/manufacturing-maintainers can someone please check?
Thanks

JordiBForgeFlow (approved):

PR #289, opened 2018-07-30
[FIX][quality_control_stock] get_lot and get_picking should be able to manage recordsets

SimoRubi:
Since those methods are used in computed fields, they should be able to manage recordsets


LoisRForgeFlow (approved):
:+1:

pedrobaeza (approved):

PR #292, opened 2018-08-07
[8.0][FIX] mrp_sale_info: Sale Order with multiple procurement orders…

ljsalvatierra-binovo:
… loses reference on MRP production order.

### DIFF
--- a/mrp_sale_info/models/mrp_production.py
+++ b/mrp_sale_info/models/mrp_production.py
@@ -1,21 +1,38 @@
 # -*- coding: utf-8 -*-
 # © 2016 Antiun Ingenieria S.L. - Javier Iniesta
+# Copyright 2018 Binovo IT Human Project SL
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields
+from openerp import models, fields, api
 
 
 class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
+    sale_line_id = fields.Many2one(
+        comodel_name='sale.order.line', string='Sale Line',
+        compute='_compute_sale_line_id')
     sale_id = fields.Many2one(
         comodel_name='sale.order', string='Sale order',
         readonly=True, store=True,
-        related='move_prod_id.procurement_id.sale_line_id.order_id')
-    sale_line_id = fields.Many2one(
-        comodel_name='sale.order.line', string='Sale Line',
-        related='move_prod_id.procurement_id.sale_line_id')
-    partner_id = fields.Many2one(related='sale_id.partner_id',
-                                 string='Customer', store=True)
-    commitment_date = fields.Datetime(related='sale_id.commitment_date',
-                                      string='Commitment Date', store=True)
+        related='sale_line_id.order_id')
+    partner_id = fields.Many2one(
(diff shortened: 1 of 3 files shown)


anajuaristi:
I don't understand the point, nor the solution you give unless you explain the use case a little bit more. An example will be also welcome.
When you say "Sale Order with multiple procurement orders… " 
There is a single procurement order for each sale line. Not needed complex calculations here.
From my side would be :-1:

mrp_sale_info/models/mrp_production.py
```diff
@@ -1,21 +1,38 @@
...
+        string='Commitment Date', store=True)
+
+    @api.multi
+    def _compute_sale_line_id(self):
+        def get_parent_move(move_obj):
+            if move_obj.move_dest_id:
+                return get_parent_move(move_obj.move_dest_id)
+            return move_obj
+
+        for record in self:
```
oihane, line 34: You can directly filter, and avoid adding `if record.move_prod_id` by changing the for
```python
for record in self.filtered('move_prod_id'):
```

ljsalvatierra-binovo:
Hi, sorry for not explaining my case.

I've faced the problem with the following configuration:

* Multiple warehouses and locations.
* Raw materials are transferred from one warehouse to another, because the production is for example in another location.

In my case I have:

* 1 Sale order.
* 3 Stock pickings (generated from procurements).

  * Stock -> Customer (outgoing)
  * Stock -> production location (internal)
  * Production location -> Stock (internal)

* 1 Manufacture order (generated from procurement).

Kind regards,
Luis

mrp_sale_info/models/mrp_production.py
```diff
@@ -1,21 +1,38 @@
...
+        comodel_name='res.partner', string='Customer',
+        store=True,
+        related='sale_id.partner_id')
+    commitment_date = fields.Datetime(
+        related='sale_id.commitment_date',
+        string='Commitment Date', store=True)
+
+    @api.multi
+    def _compute_sale_line_id(self):
+        def get_parent_move(move_obj):
```
oihane, line 29: usually move_obj will be used to refer to self.env['stock.move'] so change it to only move

mrp_sale_info/models/mrp_production.py
```diff
@@ -1,21 +1,38 @@
...
-        related='move_prod_id.procurement_id.sale_line_id.order_id')
-    sale_line_id = fields.Many2one(
-        comodel_name='sale.order.line', string='Sale Line',
-        related='move_prod_id.procurement_id.sale_line_id')
-    partner_id = fields.Many2one(related='sale_id.partner_id',
-                                 string='Customer', store=True)
-    commitment_date = fields.Datetime(related='sale_id.commitment_date',
-                                      string='Commitment Date', store=True)
+        related='sale_line_id.order_id')
+    partner_id = fields.Many2one(
```
oihane, line 19: You are not doing any real change to this definition so avoid formatting code

oihane:
Please add tests in order to show that it works as it was working (with only a procurement) and with your case of multiple procurements. I don't understand what are you trying to do.

LoisRForgeFlow:
Unresolved comments for a long time. I am now closing, please reopen if you think this is still relevant.

PR #294, opened 2018-08-08
[11.0][MIG] mrp_sale_info: Migration to 11.0

oihane:


oihane:
@OCA/manufacturing-maintainers can anyone check this please, thanks in advance

bodedra (approved):
Functional review. :+1:

mrp_sale_info/__manifest__.py
```diff
@@ -1,28 +1,28 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "MRP Sale Info",
     "summary": "Adds sale information to Manufacturing models",
-    "version": "9.0.1.0.0",
+    "version": "11.0.1.0.0",
     "category": "Manufacturing",
-    "website": "http://www.antiun.com",
+    "website": "https://github.com/oca/manufacture",
     "author": "Antiun Ingeniería S.L., "
```
pedrobaeza: Please remove Antiun, OdooMRP team, Serv. Tecnol... and add Tecnativa

pedrobaeza (changes requested):

mrp_sale_info/__manifest__.py
oihane: Change done

pedrobaeza (approved):

gustavovalverde:
This used to have a relation to sale order lines. Was it removed on purpose or based on an actual limitation?

anajuaristi:
I think the correct behavior should be having a link with sale.order.line and so ... related to sale_id 
I don't think it's complex to change on the module itselft but... hook needed to fix old data updating the module.
@oihane @pedrobaeza what do you think? should we afford this on V11 or change directly when migrating to V12?

pedrobaeza:
Yes, I agree

PR #301, opened 2018-08-28
[10.0][mrp_mto_with_stock] Take draft PO quantities into account for forecast

florian-dacosta:
Hello,
In the module mrp_mto_with_stock, there is an option to llok at the forecast quantity, when checking the availability of a Manufacture Order. If the forecast quantity is negative, then Odoo will try to procure the missing products.
But there is a big issue, the forecast quantity only take validated stock move into account. So if there is draft purchase order on the system, it won't count in the forecast stock.

The main issue here, is if a Manufacture Order generate a draft PO when clicking on 'check availability' button, it will increment the draft PO each time we click on the button until the draft PO is validated.

So, the aim of this PR is to add an auto-install module to take care of this problem when mrp_mto_with_stock and purchase module are both installed.

@jbeficent @lreficent @bodedra


bodedra:
@florian-dacosta Thanks for PR. As you mentioned that 

> The main issue here, is if a Manufacture Order generate a draft PO when clicking on 'check availability' button, it will increment the draft PO each time we click on the button until the draft PO is validated.

What's your suggestion with following:

Can we simply override 'check availability'  button method in MO ? 

Like check draft PO is already generated from MO. If we found, then skip to create new PO.

bodedra:
I knew that, you have already planned it with above approach, but want to the reason behind with your approach.

florian-dacosta:
@bodedra Actually, with the mode "with forecast stock" (option at warehouse level) the manufacture order generate a procurement which has no link with MO.
So we can't override 'check availability'' and check if there is already a procurement for this MO, as there is no link at all.

Also, if a PO has been created for a particular MO, but is still in draft, we want to take it into account when another MO is checked.
Let's take an example : 
We have 0 product in stock, and 0 in forecast quantity.
I create MO1, that need 2 products => It will generate a procurement and then a draft PO for 2 products.
Then I do an inventory, because I see I have 1 product available in my stock. My forecast stock is -1. (I have 1 product I have 2 outgoing products from MO1 et and the PO is still in draft, so no incoming products)
Then I create MO2, that need one product. Odoo will create a PO for 1 product because the forecast stock is negative. This is because the draft PO is not taken into account. Else, the forecast stock would be 0, and no PO would be generated (which is what we want)

mrp_mto_with_stock/models/mrp_production.py
```diff
@@ -99,6 +99,13 @@ def _prepare_mto_procurement(self, move, qty, mto_with_no_move_dest_id):
             vals['move_dest_id'] = move.id
         return vals
 
+    # This method should be overriden in submodule to manage cases where
+    # we need to add quantities to the forecast quantity. Like draft purchase
+    # order, purchase request, etc...
+    @api.model
+    def _get_incoming_qty_waiting_validation(self, move):
+        return 0.0
```
LoisRForgeFlow: Put the comments of the method in a triple quote comment after the method declaration
```python
def _get_incoming_qty_waiting_validation(self, move):
    """comment"""
    return
```

mrp_mto_with_stock_purchase/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "MRP MTO with Stock Purchase",
+    "summary": "Module needed to make mrp_mto_with_stock module compatible "
+               "with purchase module.",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://odoo-community.org/",
```
LoisRForgeFlow: https://github.com/OCA/manufacture

LoisRForgeFlow (approved):
Small comments inline. Other than that LGTM :+1:

bodedra (approved):
LGTM :+1:

florian-dacosta:
Comments attended!

LoisRForgeFlow:
The runbot error is due to the lack of README in the new module. Could you check this [Pedro's comment](https://github.com/OCA/account-payment/pull/199#discussion_r213626781) and try and go for the 3rd option? if you have issues let me know.

Thanks a lot!

florian-dacosta:
@lreficent Thanks. I thought the main README file was not requiered anymore since we split it in smaller files....
I tried the 3rd solution explained by pedro, I hope it does the trick!

LoisRForgeFlow (approved):
Great!

PR #302, opened 2018-08-28
[11.0][mrp_mto_with_stock] Fix forecast qty with purchase module

florian-dacosta:
Same PR than https://github.com/OCA/manufacture/pull/301 but for version 11.

@jbeficent @lreficent @bodedra


bodedra (approved):
Just need to address same point here as commented by @lreficent at https://github.com/OCA/manufacture/pull/301#pullrequestreview-150446850 Other than that LGTM :+1:

LoisRForgeFlow (approved):
Seems good :+1:

PR #320, opened 2018-11-14
[10.0][FIX] quality_control_stock: singleton error on write

MiquelRForgeFlow:
You can have more than one qc_trigger with the same picking_type.


AaronHForgeFlow (approved):
:+1: LGTM

pedrobaeza:
I think there's a constraint for that. If not, please rename variable to plural.

MiquelRForgeFlow:
Renamed to plural.

JordiBForgeFlow (approved):

PR #225, opened 2017-09-19
[9.0][ADD][mrp_repair_refurbish]

JordiBForgeFlow:
This module adds the ability to obtain refurbished product as a consequence
of the repair of a product that was defective.

Configuration
=============

#. For each product that can potentially be refurbished, define the
   refurbished version in the product form.

#. A default 'Refurbished' virtual location is created during module install,
   and proposed by default on each product.

Usage
=====

#. Create an MRP repair.
#. Indicate a product to repair, If the product has a proposed refurbished
   version, it will be proposed to be used. The destination location
#. Add operations if needed.
#. Complete the repair.

The initial product will be moved to the 'Scrap' location, and the
refurbished product will be moved from a 'Refurbish' location to the desired
destination location.

The components that were added to the repair order will be moved to the
'Refurbish' location.

PR #321, opened 2018-11-19
[11.0][MIG] mrp_repair_refurbish

LoisRForgeFlow:
Migration based on https://github.com/OCA/manufacture/pull/225. Added tests


HviorForgeFlow (approved):
LGTM :+1: Code and functional review

mrp_repair_refurbish/models/mrp_repair.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2017-18 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class MrpRepair(models.Model):
+    _inherit = 'mrp.repair'
+
+    to_refurbish = fields.Boolean('To refurbish')
```
rousseldenis: string not needed

mrp_repair_refurbish/models/mrp_repair.py
```diff
@@ -0,0 +1,76 @@
...
+            self.location_dest_id = self.refurbish_location_dest_id
+            self.refurbish_product_id = False
+            self.refurbish_location_dest_id = False
+
+    @api.multi
+    def action_repair_done(self):
+        res = super(MrpRepair, self).action_repair_done()
+        for repair in self:
+            if repair.to_refurbish:
+                move = self.env['stock.move'].create({
```
rousseldenis: @lreficent Maybe externalizing the dict preparation in another function to help inheritance ?

mrp_repair_refurbish/models/product_template.py
```diff
@@ -0,0 +1,44 @@
...
+    @api.depends('product_variant_ids',
+                 'product_variant_ids.refurbish_product_id')
+    def _compute_refurbish_product(self):
+        unique_variants = self.filtered(lambda template:
+                                        len(template.product_variant_ids) == 1)
+        for template in unique_variants:
+            template.refurbish_product_id = \
+                template.product_variant_ids.refurbish_product_id
+        for template in (self - unique_variants):
+            template.refurbish_product_id = False
```
rousseldenis: I think this is not needed as not stored (False is default value before compute)

mrp_repair_refurbish/models/product_template.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.one
```
rousseldenis: This is deprecated

mrp_repair_refurbish/models/product_template.py
```diff
@@ -0,0 +1,44 @@
...
+    def _set_refurbish_product(self):
+        if len(self.product_variant_ids) == 1:
+            self.product_variant_ids.refurbish_product_id = \
+                self.refurbish_product_id
+
+    def _search_refurbish_product(self, operator, value):
+        products = self.env['product.product'].search([
```
rousseldenis, line 41: Just return
```suggestion
        return [('refurbish_product_id', operator, value)]
```
You avoid a search.

rousseldenis (changes requested):
Code review. Some comments

mrp_repair_refurbish/models/product_template.py
LoisRForgeFlow: it is compute non-stored field, how would that work?

mrp_repair_refurbish/models/mrp_repair.py
LoisRForgeFlow: sure, good point.

mrp_repair_refurbish/models/product_template.py
LoisRForgeFlow: Indeed

mrp_repair_refurbish/models/product_template.py
rousseldenis: Right! My bad

LoisRForgeFlow:
@rousseldenis could you update your review?

rousseldenis (approved):

PR #323, opened 2018-12-04
[12.0][MIG] Migrated mrp_progress_button to v12.

sudhir-erpharbor:
Please review and give feedback.


mrp_progress_button/views/production.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="mrp_production_progress_button_form_view" model="ir.ui.view">
```
tbaden, line 4: cosmetical change: use original id

> prefer naming the inherited view after the original over using a name which follows the guidelines

not relevant for approvement

tbaden (changes requested):
thanks for the work 👍 
code review: please squash your commits

tested in runbot as well

LoisRForgeFlow:
@sudhir-erpharbor could you answer the open review? also, please squash your migration commits.

Thanks!

mrp_progress_button/views/production.xml
sudhir-erpharbor: @tbaden Sorry did not get you. Would you please let me know the exact change I need to do?

mymage (approved):

mrp_progress_button/views/production.xml
tbaden: @sudhir-erpharbor the id should be 'mrp_production_form_view' instead of 'mrp_production_progress_button_form_view', but it's not required for the approve itself.
Will approve it, but would be a nice change to have, as it's the oca naming convention.

tbaden (approved):

gurneyalex (approved):

marcelsavegnago (approved):
LGTM

PR #333, opened 2019-01-24
[11.0][mrp_multi_level] convert read_groups into sql queries due to issue with locale

JordiBForgeFlow:
We had a long standing issue using read_group where the group is done by day.

Finally we decide to use plain sql queries for this case, until we find a better approach.

### DIFF
--- a/mrp_multi_level/wizards/mrp_multi_level.py
+++ b/mrp_multi_level/wizards/mrp_multi_level.py
@@ -5,14 +5,10 @@
 
 from odoo import api, fields, models, exceptions, _
 from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
-from odoo.tools import DEFAULT_SERVER_DATE_FORMAT
 from datetime import date, datetime, timedelta
-import locale
 import logging
 from odoo.tools.float_utils import float_round
 logger = logging.getLogger(__name__)
-
-ODOO_READ_GROUP_DAY_FORMAT = '%d %b %Y'
 
 
 class MultiLevelMrp(models.TransientModel):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think this is a good idea. SQL should be avoided always. What is the issue you have? Please explain to see if I can tell you other solution.

JordiBForgeFlow:
@pedrobaeza The problem is that we have to use this https://github.com/OCA/manufacture/pull/333/files#diff-c5d846f2591493501b5fd8e13123adecL643

to be able to group by date. But we found that some environment, such as Windows, do not recognize  'en_GB.UTF-8' as locale.

`  File "C:\Program Files (x86)\Odoo 11.0e\server\addons\odoo-addons\mrp_multi_level\wizards\mrp_multi_level.py", line 773, in run_mrp_multi_level
    self._mrp_final_process()
  File "C:\Program Files (x86)\Odoo 11.0e\server\addons\odoo-addons\mrp_multi_level\wizards\mrp_multi_level.py", line 741, in _mrp_final_process
    self._init_mrp_inventory(mrp_product)
  File "C:\Program Files (x86)\Odoo 11.0e\server\addons\odoo-addons\mrp_multi_level\wizards\mrp_multi_level.py", line 669, in _init_mrp_inventory
    group['mrp_date:day'])
  File "C:\Program Files (x86)\Odoo 11.0e\server\addons\odoo-addons\mrp_multi_level\wizards\mrp_multi_level.py", line 649, in _convert_group_date_to_default
    locale.setlocale(locale.LC_TIME, 'en_GB.UTF-8')
  File "C:\Program Files (x86)\Odoo 11.0e\python\lib\locale.py", line 598, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting`

JordiBForgeFlow:
@pedrobaeza we have tried to find good solutions to do the read_group by date, but none seems to work fine because Odoo outputs a locale-specific format, which we don't want.

HviorForgeFlow (commented):
Not related with this PR, but [here](https://github.com/OCA/manufacture/pull/333/files#diff-c5d846f2591493501b5fd8e13123adecR574) should use mrp_product object instead of id

pedrobaeza:
Hey, you haven't let me time to think about this, and no reviews have been done...

JordiBForgeFlow:
@pedrobaeza we had to move on with this, as it was causing issues. When we figure out a better mechanism we can propose a new PR. @hveficent already reviewed, and we also have extended test coverage.

pedrobaeza:
You can use PR in your deployments without being merge. I don't think this is the proper way, but I'm not personally using the module, so I'm not going to be nitpicking.

PR #342, opened 2019-03-12
[12.0][MIG] mrp_repair_refurbish > repair_refurbish

grindtildeath:
I renamed the module according to the rename of mrp_repair to repair in odoo v12.0.

I'm not sure if this module should stay here or if we should target another repo as suggested here : https://github.com/OCA/manufacture/issues/310#issuecomment-458166839

cc @pedrobaeza @anajuaristi @jbeficent @lreficent


gurneyalex (approved):
code review

LoisRForgeFlow (approved):
Code review LGTM, thank for migration.

I'm not sure about the new repo though (I posted a msg in the migration issue with my opinion).

LoisRForgeFlow:
No answer in the issue, I think we can move forward with this.

Thanks for migrating!

Issue #347, opened 2019-04-12
[12.0] repair_refurbish, product with tracking show error when end repair order


bodedra:
Steps to reproduce:

- Create a sale order with product tracking either lot or serial.
- Confirm it. Assign serial number and validate delivery order.
- Create a Customer RMA and pick delivered product.
- Create a repair order and confirm it.
- Start repair and end repair.

Current behaviour:

![Screenshot from 2019-04-12 22-43-22](https://user-images.githubusercontent.com/14229503/56055202-9347ef80-5d76-11e9-8c77-1b8f514452be.png)

Excepted behaviour:

- It should end repair.

Additionally,  I have given following proper values.

![Screenshot from 2019-04-12 23-00-12](https://user-images.githubusercontent.com/14229503/56055351-ec178800-5d76-11e9-9609-4e2f70e1e4e4.png)

bodedra:
Ref: #348  We may close these issue.

PR #348, opened 2019-04-12
[12.0][FIX] repair_refurbish, product with tracking show error when end repair order

bodedra:
Issue reference: https://github.com/OCA/manufacture/issues/347

Analysis:

- Create a stock move (https://github.com/OCA/manufacture/blob/12.0/repair_refurbish/models/repair.py#L62)
- When Odoo creates stock move line, we receive following values.
Screenshot from 2019-04-12 21-22-53
- It fails at https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move_line.py#L408
- So instead of Odoo creates stock move line, we create it and pass lot number.

### DIFF
--- a/repair_refurbish/models/repair.py
+++ b/repair_refurbish/models/repair.py
@@ -49,8 +49,17 @@ class RepairOrder(models.Model):
             self.address_id.id or False,
             'location_id': self.location_dest_id.id,
             'location_dest_id': self.refurbish_location_dest_id.id,
-            'restrict_lot_id': self.refurbish_lot_id.id,
-        }
+            'move_line_ids': [(0, 0, {
+                'product_id': self.refurbish_product_id.id,
+                'lot_id': self.refurbish_lot_id.id,
+                'product_uom_qty': self.product_qty,
+                'product_uom_id': self.product_uom.id or
+                self.refurbish_product_id.uom_id.id,
+                'qty_done': self.product_qty,
+                'package_id': False,
+                'result_package_id': False,
+                'location_id': self.location_dest_id.id,
+                'location_dest_id': self.refurbish_location_dest_id.id})]}
 
     @api.multi
     def action_repair_done(self):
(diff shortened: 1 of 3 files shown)


bodedra:
@jbeficent @lreficent @grindtildeath @pedrobaeza

bodedra:
@aheficent  @hveficent

LoisRForgeFlow (approved):
Code review, LGTM :+1:

bodedra:
@OCA/manufacturing-maintainers

HviorForgeFlow (approved):

repair_refurbish/models/repair.py
```diff
@@ -50,7 +50,17 @@ def _get_refurbish_stock_move_dict(self):
             'location_id': self.location_dest_id.id,
             'location_dest_id': self.refurbish_location_dest_id.id,
             'restrict_lot_id': self.refurbish_lot_id.id,
```
grindtildeath: I guess this line can be deleted.

grindtildeath (approved):
LGTM :+1:

repair_refurbish/models/repair.py
bodedra: @grindtildeath these lines haven't disturb anything at all. so we are good to leave as it is. Please correct me. `restrict_lot_id` field is remove from Odoo 12, so we can remove that.

grindtildeath: @bodedra Yes, I only meant to remove the `restrict_lot_id` one, the other lines must of course stay there.
Actually this line gives a warning in server logs:
`WARNING oca_manufacture odoo.models: stock.move.create() with unknown fields: restrict_lot_id `

bodedra:
Thanks for your review @lreficent @hveficent and @grindtildeath 

@OCA/manufacturing-maintainers Would you please merge these ?

PR #350, opened 2019-04-14
[MIG] mrp_sale_info: Migration to 12.0

skukered:


skukered:
a question, why in version 12 when one makes a sale of a product (which is configured under order and to create a production order) the generated production order does not have the same procurement.group associated with it that was generated from the sale but creates a new one?

pedrobaeza:
Check procurement group propagation option in the stock rule.

mrp_sale_info/README.rst
```diff
@@ -0,0 +1,68 @@
...
+    * Commitment Date
+
+Usage
+=====
+
+New fields are displayed in tree and in form views (Extra information tab).
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/129/11.0
```
suniagajose: change link  to branch `12.0`

mrp_sale_info/README.rst
```diff
@@ -0,0 +1,68 @@
...
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/manufacture/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+manufacture/issues/new?body=module:%20
+mrp_sale_info%0Aversion:%20
+11.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
suniagajose: same here.... replace `11.0` to `12.0`

suniagajose (changes requested):
Still references to `11.0`

suniagajose:
:+1: functional review

@skukered 
@pedrobaeza

mrp_sale_info/README.rst
skukered: OK, thank you.

CasVissers-360ERP (approved):
LGTM, Functional testing.

alyngsoe:
Can somebody with the rights please merge this pull request to the v12 branch. As far as i can see, the module is ready to migrate to v12 ?

sergiocorato (approved):

rizagokay:
This module can't find the reference when there is a reordering rule defined on product. Because in this scenario the mrp.production is being created with orderpoint instead of sale.order.

This causes the procurement_group_id.sale_id field to be empty when the origin is reference to an orderpoint

mrp_sale_info/models/mrp_production.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = "mrp.production"
+
+    sale_id = fields.Many2one(
+        comodel_name='sale.order', string='Sale order', readonly=True,
+        store=True, related='procurement_group_id.sale_id')
+    partner_id = fields.Many2one(
+        comodel_name='res.partner', related='sale_id.partner_id',
+        string='Customer', store=True)
+    commitment_date = fields.Datetime(
+        related='sale_id.commitment_date', string='Commitment Date',
```
marianhtuyet, line 18: please remove `string`. The system will automatically handle the name of the field so we don't need define the name of field again

mrp_sale_info/models/mrp_workorder.py
```diff
@@ -0,0 +1,17 @@
...
+
+class MrpWorkorder(models.Model):
+    _inherit = "mrp.workorder"
+
+    sale_id = fields.Many2one(related='production_id.sale_id',
+                              string='Sale order', readonly=True, store=True)
+    partner_id = fields.Many2one(related='sale_id.partner_id', readonly=True,
+                                 string='Customer', store=True)
+    commitment_date = fields.Datetime(related='sale_id.commitment_date',
+                                      string='Commitment Date', store=True,
```
marianhtuyet, line 16: please remove the string

mrp_sale_info/models/stock_rule.py
```diff
@@ -0,0 +1,36 @@
...
+            product_uom,
+            location_id,
+            name,
+            origin,
+            values,
+            bom
+        )
+        res.update({
+            'procurement_group_id': values.get(
+                'group_id').id if values.get('group_id', False) else False,
```
daramousk, line 34: Where is this `group_id` coming from?

sergiocorato:
> This module can't find the reference when there is a reordering rule defined on product. Because in this scenario the mrp.production is being created with orderpoint instead of sale.order.
> 
> This causes the procurement_group_id.sale_id field to be empty when the origin is reference to an orderpoint

If there is no reference to find, it's a normal behaviour that it doesn't find a reference.

JuaniFreedoo:
functionally it has an error, when you cancel a production order all the delivery note lines are canceled

JuaniFreedoo:
My solution is in my repository, it seems valid to you?

https://github.com/JuaniFreedoo/manufacture/tree/12.0-mig-mrp_sale_info

sergiocorato:
Can some maintainers give a look or merge this one?

mrp_sale_info/models/mrp_production.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2019 Rubén Bravo <rubenred18@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    sale_id = fields.Many2one(
```
sergiocorato, line 11: `sale_id` is a duplication of `sale_order_id` [here](https://github.com/OCA/sale-workflow/blob/12.0/sale_mrp_link/models/mrp_production.py#L11) in module `sale_mrp_link`.
I think it's better to avoid duplication and preserve only one of these, the latter as is defined in a module in use.

pedrobaeza: @sergiocorato the link in the other module is very weak (through a search of a string in origin). This one has been here for a long and IMO this is the definitive one to be preserved. Also the typical name for this field in other places is `sale_id`, and not `sale_order_id`.

pedrobaeza:
I'm taking over this for finishing it (rebasing + squashing) + taking into account @JuaniFreedoo's comment, which is true, and can be avoidable through the extra field.

mrp_sale_info/models/mrp_production.py
sergiocorato: @pedrobaeza Ok, perhaps coordinate https://github.com/OCA/sale-workflow/pull/1109?

pedrobaeza:
Superseded by #521, where I have added the extra field for not overlapping existing one, and added tests.

JuaniFreedoo:
Thanks and regards

PR #354, opened 2019-04-25
[12.0][ADD] New module mrp_subcontract_mo

remi-filament:
This module allows to subcontract Manufacturing Orders

When a manufacturing order is subcontracted, a Subcontractor is selected (from Suppliers) and optionnally a service to be subcontracted for the manufacturing (e.g. Painting or Assembly).
Raw materials from BOM and finished materials pickings and stock moves are automatically created, as well as purchase order in case a service to be subcontracted has been selected.


rvalyi:
Hello @remi-filament thanks for your work.
I didn't look at your module yet. However, are you aware that Odoo SA is open sourcing its MRP subcontracting module on master for v13.
https://github.com/odoo/documentation-user/blob/bbc3b224e201b436685dee21a6a83644561a8ace/manufacturing/management/subcontracting.rst
https://github.com/odoo/odoo/pull/32390/commits/b1d6801bba12f9e7b2e644633b5be625c8fd5443

So in order to maximize comptability, I think that it would be interresting to be the closest possible to the Odoo SA master module or at least justify when you made a different decision. If you could make that analysis, that would certainly help getting OCA traction on your module...

remi-filament:
Thank you @rvalyi, I heard that they published a module the day after we completed the coding of our module...
Yet I did not know how to check the functionality they put in place, thank you for the link.

The way of handling subcontracting is quite different from what I can see:

- On Odoo module : you decide a priori of a product which is subcontracted, and you create a BoM of type subcontracting
- On the module I propose : you decide at the time of starting the manufacturing order whether you do it internally or you subcontract it (after the MO is created)

From my point of view these are 2 different use cases (in particular for complex products which are made out of intermediate products for which MO would be automatically created upon Sale or creation of MO for the final product)

The underlying mechanisms seem to comply (in particular the fact of moving the components from stock to subcontractor to production then have finished product at subcontractor moved to stock).

Please let me know your point of view ?

rvalyi (changes requested):
A first problem I can spot is that both modules are named mrp_subcontracting (see https://github.com/odoo/odoo/tree/master/addons/mrp_subcontracting ) so this is going to create migration problems in the future. So may be rename this module in a more explicit non conflicting way?

hparfr:
When choosing subcontract order, I get the following error:
```py 
  File "/home/odoo/build/OCA/manufacture/mrp_subcontracting/wizard/mrp_subcontract.py", line 78, in do_subcontract
    self.production_id._generate_service_purchase()
  File "/home/odoo/build/OCA/manufacture/mrp_subcontracting/models/mrp_production.py", line 356, in _generate_service_purchase
    'price_unit': service_supplier_infos[0].price,
  File "/.repo_requirements/odoo/odoo/models.py", line 5092, in __getitem__
    return self._browse((self._ids[key],), self.env)
IndexError: tuple index out of range
```
The supplierinfos only contains the seller (no variant, no lead time, no prices).

remi-filament:
Thanks for your feedbacks,
@rvalyi  I renamed the module mrp_subcontract_mo
@hparfr Indeed I did not consider that price might be empty in supplier info. it should be fixed with the latest commit

hparfr:
Same test case, I now have : 
```
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: location_dest_id - location.dest.id] 
```

remi-filament:
@hparfr, should be fixed now thanks

alang0808:
@remi-filament Thanks for the works on this module, I fully agree to your mind, it is quite flexible. 
In addition, in the case that you mentioned above-- ("in particular for complex products which are made out of intermediate products for which MO would be automatically created upon Sale or creation of MO for the final product"), I have some proposals below:
   (1) In most practice is the subcontracting could be set automatically by default while could change to 
 manual mode if needed;
   (2) The subcontracting service could be set as default as well(maybe according to product template);
   (3) The receipts of subcontracting service should not be one lot only, if we subcontacted 100 PCS final products, while the subcontractor only deliver back 90 PCS (the remaining maybe scrapped...), we should not pay full amount of that purchase order, instead we should make reception according to the delivered quantity only, the invoicing should follow the delivered quantity as well as the delivery  period(case of cross months

alang0808:
@remi-filament Thanks for the works on this module, I fully agree to your mind, it is quite flexible.
In addition, in the case that you mentioned above-- ("in particular for complex products which are made out of intermediate products for which MO would be automatically created upon Sale or creation of MO for the final product"), I have some proposals below:
(1) In most practice is the subcontracting could be set automatically by default while could change to
manual mode if needed;
(2) The subcontracting service could be set as default as well(maybe according to product template);
(3) The receipts of subcontracting service should not be one lot only, if we subcontacted 100 PCS final products, while the subcontractor only deliver back 90 PCS (the remaining maybe scrapped...), we should not pay full amount of that purchase order, instead we should make reception according to the delivered quantity only, the invoicing should follow the delivered quantity as well as the delivery period(case of cross months).

   Please kindly consider these cases for future commit, thank you!

remi-filament:
Closing as not maintained anymore since not merged

PR #358, opened 2019-05-01
[MIG] mrp_production_service: Migration to 12.0

skukered:
depends https://github.com/OCA/purchase-workflow/pull/677


sergiocorato:
@skukered purchase-workflow repo is missing in oca_dependencies.txt for module subcontracted_service, this will solve start issue in runbot/travis
Thanks for the contribution!

mrp_production_service/models/mrp_production.py
```diff
@@ -0,0 +1,50 @@
...
+        res = super()._generate_moves()
+        for production in self:
+            factor = production.product_uom_id._compute_quantity(
+                production.product_qty,
+                production.bom_id.product_uom_id
+            ) / production.bom_id.product_qty
+            boms, lines = production.bom_id.explode(
+                production.product_id, factor,
+                picking_type=production.bom_id.picking_type_id)
+            for line in lines:
```
daramousk, line 46: You could just do a `filtered()` here in order to avoid looping through everything and remove the line below.

daramousk: Also, it seems that the products of type `service` are removed when you create a new Manufacturing Order here https://github.com/OCA/OCB/blob/12.0/addons/mrp/models/mrp_production.py#L446

daramousk (changes requested):

HviorForgeFlow:
@skukered could you attend comments to move this forward?

mrp_production_service/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import test_mrp_production_service
```
HviorForgeFlow: Add blank line

mrp_production_service/models/mrp_production.py
HviorForgeFlow: Could you attend this suggestion, I think it is a good improvement.

HviorForgeFlow (approved):
LGTM ~ Minor comments.

Tardo:
Is it expected to give the last push?

elvise:
Hi, any good news for merge...?
If I can help please let me know...

sergiocorato (approved):

sergiocorato:
@OCA/manufacturing-maintainers merge?

marcelsavegnago (approved):
LGTM

LoisRForgeFlow (changes requested):
@skukered Can you clean commit history and do a proper rebase so you don't include merge commits (https://github.com/OCA/manufacture/pull/358/commits/286d5cc101ab6cfda141c49f458a8e09865a03fc)?

LoisRForgeFlow:
@daramousk Is your suggestion covered? can you update your review?

pedrobaeza:
Isn't this similar to `mrp_subcontracting`?

ypapouin (approved):
Functional testing is Ok. LGTM
Note that there is a dependency on `subcontracted_service` module in order to have a buy route.

sergiocorato:
> Isn't this similar to `mrp_subcontracting`?

No, this one adds ability to subcontract a simple service (without need of stock moves).

mrp_production_service/models/mrp_production.py
sergiocorato: > You could just do a `filtered()` here in order to avoid looping through everything and remove the line below.

lines is `[(record), {...}]` so filtered() is not an option

sergiocorato: > Also, it seems that the products of type `service` are removed when you create a new Manufacturing Order here https://github.com/OCA/OCB/blob/12.0/addons/mrp/models/mrp_production.py#L446

This is true, service is ignored in mrp production. This port to a production order incomplete, as the service is ordered but missing in production order. 

This logic is better implemented in mrp_subcontracting, so I think this PR should be closed.

sergiocorato (changes requested):
This PR is superseeded by mrp_subcontracting

MiquelRForgeFlow:
@sergiocorato first you said no, now you say yes. why did your opinion change?

sergiocorato:
> @sergiocorato first you said no, now you say yes. why did your opinion change?

Because as I wrote [here](https://github.com/OCA/manufacture/pull/358/files#r654309733) the logic of this module is a duplication of an existing one (with the little difference that this one add ability to use services, but in a way they are not traced in production as noticed [here](https://github.com/OCA/manufacture/pull/358/files#r325206431)), which is already implemented in future version, so in my opinion it is better to concentrate there.
My2cents

MiquelRForgeFlow:
> which is already implemented in future version

In which version exactly?

sergiocorato:
mrp_subcontracting module from 12 and so on

PR #359, opened 2019-05-02
[FIX] quality_control_stock: Fix error when transferring pickings.

alfredoavanzosc:
When transferring pickings can come more than one object.

### DIFF
--- a/quality_control_stock/models/stock_picking.py
+++ b/quality_control_stock/models/stock_picking.py
@@ -37,18 +37,19 @@ class StockPicking(models.Model):
     def do_transfer(self):
         res = super(StockPicking, self).do_transfer()
         inspection_model = self.env['qc.inspection']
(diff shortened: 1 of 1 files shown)


alfredoavanzosc:
@pedrobaeza , @oihane .... review please.

quality_control_stock/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Quality control - Stock",
-    "version": "8.0.1.0.1",
+    "version": "8.0.1.0.2",
```
umiphos: you are not making a migration, why changing the version?

alfredoavanzosc: Changed the theme of the version. Thanks.

anajuaristi:
I merge directly to solve the error as it is a very small change

Issue #367, opened 2019-05-27
[12.0] repair_refurbish : Issue with stock location movement


bodedra:
Steps to reproduce:

Configuration:
--------------------

- Create two Products. (Product Type | Storable Product, Tracking | By Unique Serial Number)

    - Wagon and
    - Wagon - Refurbish

- Increase product Wagon inventory with 1.0 and assign serial number XXXX01. QTY will increase on "Stock" location.

Sale Process:
------------------

- Create a sale order with product Wagon, confirm it and deliverer it to a customer (serial XXXX01)

- Review on hang quantity for Wagon or serial number traceability report. It will show location movement From "Stock" to "Customers". It will be decrease quantity from Stock location.

- We will receive back product Wagon from a Customer for new changes/modification.

- Open a delivery order, click on "Return" button and receive it. It will show location movement From "Customers" to "Stock". It will be increase quantity on Stock location.

Repair Process:
------------------

- Create a new repair order and select "Product to Repair" as Wagon, lot/serial number XXXX01.

- Checked "To Refurbish" and select product "Refurbished product" as Wagon - Refurbish, "Refurbished Delivery Location" as Customers location.

- Start repair.

- End repair.

Current Behaviour:
------------------------

- It shows correct movement for Wagon - Refurbish product.

- Review on hand quantity for Wagon or serial number traceability report. It will not decrease quantity from "Stock" location.

Excepted Behaviour:
--------------------------

- It should decrease quantity for product Wagon from "Stock" location. For example: location movement should be "Stock" to "Refurbish"

I have tested these on runbot http://3371411-12-0-c9b441.runbot3.odoo-community.org/web?debug=1#id=52&model=product.product&view_type=form&menu_id=232

PR #369, opened 2019-05-27
[12.0][FIX] repair_refurbish : Issue with stock location movement

bodedra:
Issue Reference : #367

### DIFF
--- a/repair_refurbish/models/stock_move.py
+++ b/repair_refurbish/models/stock_move.py
@@ -9,8 +9,9 @@ class StockMove(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        if 'force_refurbish_location_dest_id' in self.env.context:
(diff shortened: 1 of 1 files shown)


bodedra:
@jbeficent @lreficent @grindtildeath @pedrobaeza

bodedra:
@aheficent

AaronHForgeFlow (approved):
LGTM

grindtildeath (commented):
I don't really see what is different after this change, maybe can you add a test to explicit the issue ?

bodedra:
@grindtildeath Thanks for your review. Here is full scenario #367 and help us to improve behaviour in 12.0. It works fine 11.0

LoisRForgeFlow (changes requested):
For me the code diff look exactly equivalent. I also think that a unit test would help to highlight the error.

PR #372, opened 2019-06-05
[11.0][REW/IMP] mrp_multi_level: parametric computation, consider sale_blanket_order and purchase_requisition

andreampiovesana:
Initial version for parametric computation, considering purchase_requisition and sale_blanket_order very important for forecast of materials, added mrp_responsible_user_id to select material to manage by user

### DIFF
--- a/mrp_multi_level/__manifest__.py
+++ b/mrp_multi_level/__manifest__.py
@@ -3,11 +3,12 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     'name': 'MRP Multi Level',
-    'version': '11.0.2.2.0',
+    'version': '11.0.3.0.0',
     'development_status': 'Beta',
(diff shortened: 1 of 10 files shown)


andreampiovesana:
@lreficent what about consider sale_blanket_order and purchase_requisition ? it is a must in production make_to_stock ...

LoisRForgeFlow:
@andreampiovesana As I already told you, I have nothing against considering blanket orders or purchase requisition, however I would not add them to the main module to keep dependencies simple and I suggest you to propose a two new "glue" modules:
* mrp_multi_level_requisition: depending on mrp_multi_level and purchase_requisition.
* mrp_multi_level_sale_blanket_order: depending on mrp_multi_level and sale_blanket_order.

This way you don't force people currently using (or willing to) mrp_multi_level to also install purchase_requisition or sale_blanket_order as they are not strictly necessary.

PR #373, opened 2019-06-10
[10.0][FIX] mrp_mto_with_stock: Error singleton

igallart:
Loop second haven't production, it has self.


pedrobaeza (approved):
Fast-tracking as straight-forward

PR #386, opened 2019-07-16
[11.0][IMP] mrp_progress_button

AdriaGForgeFlow:
Basic improvement to module mrp_progress_button that adds a button to set the state back to confirmed when in progress and no products have been produced.

image

cc ~ @Eficent @florian-dacosta

### DIFF
--- a/mrp_progress_button/models/mrp_production.py
+++ b/mrp_progress_button/models/mrp_production.py
@@ -16,3 +16,11 @@ class MrpProduction(models.Model):
             'date_start': datetime.now(),
         })
         return True
+
+    @api.multi
+    def action_confirm(self):
+        self.write({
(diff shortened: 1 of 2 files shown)


matt454357:
I would like to see this merged.  Maybe add a test, like this one?

``` python
def test_production_state_change_confirmed(self):
    """
        Test action_confirm() method sets state back to confirmed
    """

    production = self.production_model.create({
        'product_id': self.product_manuf.id,
        'product_qty': 1,
        'product_uom_id': self.uom_unit.id,
        'bom_id': self.bom.id
    })
    production.action_progress()
    self.assertEqual(production.state, 'progress')
    production.action_confirm()
    self.assertEqual(production.state, 'confirmed')
```

PR #387, opened 2019-07-18
[FIX] mrp_operations_extension: update qty erases operations

oihane:
If a manufacturing order quantity is updated (using the wizard) it deletes all the work orders in the MO, so it does not assign any work order to the lines to consume, and you lose that information.
This code assigns to the moves already created the new work orders that have been created by the wizard.


anajuaristi (approved):

Daniel-CA (approved):

anajuaristi:
@alfredoavanzosc  @Daniel-CA @mikelarre
Review please!!

mikelarre (approved):

PR #391, opened 2019-08-08
[12.0][UPD] Update quality_control module logo

marcelsavegnago:
- Update quality_control module logo


bodedra:
@marcelsavegnago Thanks for your contribution. We have already logo for quality_control module.  Ref: https://github.com/OCA/manufacture/blob/12.0/quality_control/static/description/icon.png

pedrobaeza:
@bodedra It seems he is putting a higher resolution PNG (120x120 instead of 80x80). Why this change, @marcelsavegnago ?

marcelsavegnago:
@pedrobaeza Because visually it is smaller than the icons of other modules. I use web responsive module and in this case i see the difference.

pedrobaeza:
Maybe the problem is in `web_responsive` itself? cc @Yajo @Tardo

yajo (approved):
Not sure, but indeed Odoo 12 supplies icons of this size, so it seems a good patch to me anyways.

bodedra (approved):

pedrobaeza:
/ocabot merge minor

@Yajo please investigate the zoom question

PR #392, opened 2019-08-12
[REF] quality_control, quality_control_stock: update all fields once …

kos94ok-3D:
…per record

### DIFF
--- a/quality_control/models/qc_inspection.py
+++ b/quality_control/models/qc_inspection.py
@@ -5,7 +5,8 @@
 # Copyright 2017 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, exceptions, fields, models, _
+from odoo import api, fields, models, _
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why? Compute records are updated on cache, so no real improvement, and this has the contrary effect: you recompute everything each time you change any field.

kos94ok-3D:
Records depend on object_id, what is a problem?
I add only required fields by object_id in function _get_values_by_object, so not all field will update.
Or I not see something?

pedrobaeza:
But again, this is not a real improvement in performance not readability, so why change it?

kos94ok-3D:
Variable, which depends on this object model, stick together I think it is more readability.
I agree that for the performance little impact.

pedrobaeza:
I still see more readable the other way, with less indentation, each method for each field... so :-1: for this change, but others can give their opinion.

kos94ok-3D:
Ok. Nobody interest in it, so I close PR.

PR #393, opened 2019-08-14
[IMP] quality_control_stock: use read_group

kos94ok-3D:


quality_control_stock/models/stock_picking.py
```diff
@@ -36,6 +24,27 @@ def _compute_count_inspections(self):
     failed_inspections = fields.Integer(
         compute="_compute_count_inspections", string="Inspections failed")
 
+    @api.depends('qc_inspections_ids', 'qc_inspections_ids.state')
+    def _compute_count_inspections(self):
+        Inspection = self.env['qc.inspection']
```
pedrobaeza: This is only used once, so no need to let a variable for this.

quality_control_stock/models/stock_picking.py
```diff
@@ -36,6 +24,27 @@ def _compute_count_inspections(self):
...
+        inverse_field = 'picking_id'
```
pedrobaeza: Why putting this field name as variable and not `state`. It's used the same number of times... Any way, I prefer in this case to directly use the field name, which is more readable and even shorter in this case.

quality_control_stock/models/stock_picking.py
```diff
@@ -36,6 +24,27 @@ def _compute_count_inspections(self):
...
+        data = Inspection.read_group([
+            ('id', 'in', self.mapped('qc_inspections_ids').ids)
+        ], [inverse_field, 'state'], [inverse_field, 'state'], lazy=False)
+        picking_data = defaultdict(dict)
+        for d in data:
+            picking_data[d[inverse_field][0]].setdefault(d['state'], 0)
+            picking_data[d[inverse_field][0]][d['state']] += d['__count']
+        for picking in self:
+            count_data = picking_data.get(picking.id, {})
+            picking.created_inspections = sum(
```
pedrobaeza:
```suggestion
            picking.created_inspections = sum(count_data.values())
```

quality_control_stock/models/stock_production_lot.py
```diff
@@ -32,3 +21,24 @@ def _compute_count_inspections(self):
         compute="_compute_count_inspections", string="Inspections OK")
     failed_inspections = fields.Integer(
         compute="_compute_count_inspections", string="Inspections failed")
+
+    @api.depends('qc_inspections_ids', 'qc_inspections_ids.state')
+    def _compute_count_inspections(self):
+        Inspection = self.env['qc.inspection']
```
pedrobaeza: Same considerations as in the other.

pedrobaeza (changes requested):
Good performance improvement!

quality_control_stock/models/stock_picking.py
kos94ok-3D: Easier copy and past function to another model. If you said that it hardly understand, I`ll fix it.

quality_control_stock/models/stock_picking.py
kos94ok-3D: Oh, You are right.

kos94ok-3D:
@pedrobaeza Is it ok?

quality_control_stock/models/stock_picking.py
```diff
@@ -36,6 +24,23 @@ def _compute_count_inspections(self):
     failed_inspections = fields.Integer(
         compute="_compute_count_inspections", string="Inspections failed")
 
+    @api.depends('qc_inspections_ids', 'qc_inspections_ids.state')
+    def _compute_count_inspections(self):
+        data = self.env['qc.inspection'].read_group([
+            ('id', 'in', self.mapped('qc_inspections_ids').ids),
+        ], ['picking_id', 'state'], ['picking_id', 'state'], lazy=False)
+        picking_data = defaultdict(dict)
```
pedrobaeza: Why do you need `defaultdict`? I think regular dict works as well with `setdefault`

pedrobaeza:
You can include yourself in contributors and copyright in file.

LoisRForgeFlow:
@kos94ok-3D Can you rebase your branch?, travis has been fixed.

LoisRForgeFlow:
@kos94ok-3D have you signed CLA? (https://odoo-community.org/page/cla)

kos94ok-3D:
> @kos94ok-3D have you signed CLA? (https://odoo-community.org/page/cla)

Yes, I have.

rousseldenis (approved):
LGTM. Code review

pedrobaeza (approved):

PR #398, opened 2019-08-22
ADD module mrp_unbuild_tracked_raw_material

bealdav:
Odoo has a limitation on tracked components of products 
which are not manufactured in the ERP.

When you try to do it, you get this warning:

    Some of your components are tracked, you have to specify a manufacturing order in order to retrieve the correct components. 

Unfortunately, it doesn't cover all the use cases

Example:
You receive eggs and you want to unbuild them in 2 parts:

    - yellow part
    - white part

Each of the parts are tracked and not linked to a previous manufacturing order
because, you don't build the eggs yourself, you subcontract it to a chicken.

This module bypass this limitation

Please could you make a review ?
@lreficent 
@jbeficent 
@sebalix 
@alexis-via


bealdav:
@pedrobaeza Can I have a wildcard for a review here ?
Thanks

pedrobaeza:
I'm afraid I don't know enough mrp_unbuild* modules for doing an acceptable review.

bealdav:
You mean, you never eat chicken ? ;-)
Maybe some suggestion for reviewer names? Thanks

pedrobaeza:
Maybe @anajuaristi, but take into account that most of the people except you and me are luckily on holidays :stuck_out_tongue_closed_eyes:

mrp_unbuild_tracked_raw_material/models/unbuild.py
```diff
@@ -0,0 +1,62 @@
...
+        components_by_lot = {
+            unbuild: unbuild.bom_id.bom_line_ids.mapped(
+                "product_id").filtered(lambda s: s.tracking == "lot")
+            for unbuild in self}
+        for unbuild in self:
+            components = components_by_serial[unbuild] | components_by_lot[
+                unbuild]
+            # To bypass Odoo limitation in use cases we need
+            # to make components as untracked
+            components.write({"tracking": "none"})
```
LoisRForgeFlow, line 35: Personally I don't like this... and you are aware of the problem as you log it in the next line.

I think if you catch the exception [here](https://github.com/odoo/odoo/blob/12.0/addons/mrp/models/mrp_unbuild.py#L115) you can continue the process with a context (something like "bypass_unbuild_tracked_products=True") and then you can inherit the create of stock.move.line searching for that context to correct the values passes [here](https://github.com/odoo/odoo/blob/12.0/addons/mrp/models/mrp_unbuild.py#L140) with the correct lot.

This way you will only be affecting this process and not potentially breaking the data consistency in parallel processes.

What do you think?

PD: I loved the description of the PR :laughing:

LoisRForgeFlow (changes requested):

mrp_unbuild_tracked_raw_material/models/unbuild.py
bealdav: Thanks for your answer @lreficent.
Ok to catch the exception, but if I do that i have a problem with
https://github.com/odoo/odoo/blob/12.0/addons/mrp/models/mrp_unbuild.py#L134
because I have no manufacturing order then `original_move` becomes `False` and no creation of `stock.move.line` occurs.

LoisRForgeFlow: I see, then you don't even need the context, you can create the stock.move.lines (and any other action needed for your use case) after calling super.

bealdav:
replaced by #429

mrp_unbuild_tracked_raw_material/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Use the standard unbuild feature. Specific behavior is defined in chatter
```
bealdav: complete usage here

PR #403, opened 2019-09-12
[12.0][FIX]repair_refurbish. Set correct destination location in stock move lines

AaronHForgeFlow:
This is not only about being consistent between the stock moves and the stock move lines.

Inventory valuation moves are not created correctly or not created because of the inconsistency between stock moves and stock move lines.

This seems to solve the issue.

### DIFF
--- a/repair_refurbish/models/stock_move.py
+++ b/repair_refurbish/models/stock_move.py
@@ -14,3 +14,16 @@ class StockMove(models.Model):
                 vals['location_dest_id'] = self.env.context[
                     'force_refurbish_location_dest_id']
         return super().create(vals_list)
+
+
+class StockMoveLine(models.Model):
+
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
Ping @bodedra @lreficent

LoisRForgeFlow:
@ageficent could you just add a small test for this? (or add an extra assert in existing tests for stock move line)

AaronHForgeFlow:
> @ageficent could you just add a small test for this? (or add an extra assert in existing tests for stock move line)

Done!

LoisRForgeFlow (approved):
Tested locally, and you are right.

Thanks!

HviorForgeFlow (approved):

bodedra (approved):

PR #404, opened 2019-09-13
[12.0][MIG] mrp_production_request

LoisRForgeFlow:
Migration to v12.


mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+    @api.multi
+    def write(self, vals):
+        res = super(MrpProductionRequest, self).write(vals)
+        for request in self:
+            request._subscribe_assigned_user(vals)
+        return res
+
+    @api.multi
+    def button_to_approve(self):
+        self.write({'state': 'to_approve'})
```
Trivedi-Vacha-SerpentCS, line 211: this looks like old fashioned. IMHO it should be self.state='to_approve'

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+        return True
+
+    @api.multi
+    def button_approved(self):
+        self.write({'state': 'approved'})
```
Trivedi-Vacha-SerpentCS, line 216: same as above

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+        return True
+
+    @api.multi
+    def button_done(self):
+        self.write({'state': 'done'})
```
Trivedi-Vacha-SerpentCS, line 221: same as above

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+        if any([s in self._get_mo_valid_states() for s in self.mapped(
+                'mrp_production_ids.state')]):
+            raise UserError(
+                _("You cannot reset a manufacturing request if the related "
+                  "manufacturing orders are not cancelled."))
+
+    @api.multi
+    def button_draft(self):
+        self._check_reset_allowed()
+        self.write({'state': 'draft'})
```
Trivedi-Vacha-SerpentCS, line 235: same as above

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+    def _check_cancel_allowed(self):
+        if any([s == 'done' for s in self.mapped('state')]):
+            raise UserError(
+                _('You cannot reject a manufacturing request related to '
+                  'done procurement orders.'))
+
+    @api.multi
+    def button_cancel(self):
+        self._check_cancel_allowed()
+        self.write({'state': 'cancel'})
```
Trivedi-Vacha-SerpentCS, line 248: same as above

mrp_production_request/models/product.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    mrp_production_request = fields.Boolean(
+        string='Manufacturing Request', help="Check this box to generate "
+        "manufacturing request instead of generating manufacturing orders "
+        "from procurement.", default=False)
```
Trivedi-Vacha-SerpentCS: No need of default=False, the default value for the boolean field is already False

mrp_production_request/models/mrp_production.py
```diff
@@ -0,0 +1,27 @@
...
+        string="Manufacturing Request", copy=False, readonly=True)
+
+    def _generate_finished_moves(self):
+        """`move_dest_ids` is a One2many fields in mrp.production, thus we
+        cannot indicate the same destination move in several MOs (which most
+        probably would be the case with MRs).
+        Storing them on the MR and writing them on the finished moves as it
+        would happen if they were present in the MO, is the best workaround
+        without changing the standard data model."""
+        move = super(MrpProduction, self)._generate_finished_moves()
```
bealdav:
```suggestion
        move = super()._generate_finished_moves()
```

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+
+class MrpProductionRequest(models.Model):
+    _name = "mrp.production.request"
+    _description = "Manufacturing Request"
+    _inherit = "mail.thread"
+    _order = "date_planned_start desc, id desc"
+
+    @api.model
+    def _company_get(self):
+        company_id = self.env['res.company']._company_default_get(self._name)
```
bealdav: Because of https://github.com/OCA/ocb/blob/12.0/odoo/addons/base/models/res_company.py#L166-L167

```suggestion
        company_id = self.env['res.company']._company_default_get()
```

mrp_production_request/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class Orderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _quantity_in_progress(self):
+        res = super(Orderpoint, self)._quantity_in_progress()
```
bealdav:
```suggestion
        res = super()._quantity_in_progress()
```

mrp_production_request/tests/test_mrp_production_request.py
```diff
@@ -0,0 +1,159 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+from odoo import fields
+from odoo.exceptions import UserError
+
+
+class TestMrpProductionRequest(TransactionCase):
+    def setUp(self, *args, **kwargs):
+        super(TestMrpProductionRequest, self).setUp(*args, **kwargs)
```
bealdav:
```suggestion
        super().setUp(*args, **kwargs)
```

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'mrp.production.request') or '/'
+        return vals
+
+    @api.model
+    def create(self, vals):
+        """Add sequence if name is not defined and subscribe to the thread
+        the user assigned to the request."""
+        vals = self._create_sequence(vals)
+        res = super(MrpProductionRequest, self).create(vals)
```
bealdav:
```suggestion
        res = super().create(vals)
```

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+        res._subscribe_assigned_user(vals)
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super(MrpProductionRequest, self).write(vals)
```
bealdav:
```suggestion
        res = super().write(vals)
```

mrp_production_request/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+    @api.model
+    def create(self, vals):
+        if 'production_id' in vals:
+            production = self.env['mrp.production'].browse(
+                vals['production_id'])
+            if production.mrp_production_request_id:
+                vals['propagate'] = False
+        return super(StockMove, self).create(vals)
```
bealdav:
```suggestion
        return super().create(vals)
```

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+
+    name = fields.Char(
+        default="/", required=True,
+        readonly=True, states={'draft': [('readonly', False)]})
+    origin = fields.Char(
+        string='Source Document',
+        readonly=True, states={'draft': [('readonly', False)]})
+    requested_by = fields.Many2one(
+        comodel_name='res.users', string='Requested by',
+        default=_get_default_requested_by,
```
bealdav: To be overidable: Then _get_default_requested_by() method definition can be placed after field declaration

```suggestion
        default=lambda self: self._get_default_requested_by,
```

mrp_production_request/models/mrp_production_request.py
```diff
@@ -0,0 +1,266 @@
...
+        'Deadline Start', copy=False, default=fields.Datetime.now,
+        index=True, required=True,
+        states={'confirmed': [('readonly', False)]}, oldname="date_planned")
+    date_planned_finished = fields.Datetime(
+        'Deadline End', copy=False, default=fields.Datetime.now,
+        index=True,
+        states={'confirmed': [('readonly', False)]})
+    company_id = fields.Many2one(
+        comodel_name='res.company', string='Company',
+        required=True, default=_company_get)
```
bealdav: To be overridable

```suggestion
        required=True, default=lambda self: self._company_get)
```

mrp_production_request/models/stock_rule.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2018-19 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+
+class StockOrder(models.Model):
```
bealdav:
```suggestion
class StockRule(models.Model):
```

mrp_production_request/models/stock_rule.py
```diff
@@ -0,0 +1,83 @@
...
+            and product_id.mrp_production_request
+
+    @api.multi
+    def _run_production_request(self, product_id, product_qty, product_uom,
+                                location_id, name, origin, values):
+        """Trying to handle this as much similar as possible to Odoo
+        production orders. See `_run_manufacture` in Odoo standard."""
+        request_obj = self.env['mrp.production.request']
+        request_obj_sudo = request_obj.sudo().with_context(
+            force_company=values['company_id'].id)
```
bealdav, line 40: Not blocking but probably we need to ensure than "company_id" key is in values ?

bealdav (changes requested):
Thanks for your contribution.
I think we can remove all `api.multi` used for api change transition.
Default behavior on records is multi (since v10), then just decorate other cases

mrp_production_request/models/mrp_production_request.py
LoisRForgeFlow: this is not "old fashioned", it's the way to make it work with both a recordset or a single record.

mrp_production_request/models/stock_rule.py
LoisRForgeFlow: This is mimicking https://github.com/odoo/odoo/blob/12.0/addons/mrp/models/stock_rule.py#L36
if it fails here, it would also fail for standard manufacturing orders and Odoo itself should have to fix it.

LoisRForgeFlow:
@bealdav @Trivedi-Vacha-SerpentCS Thanks for your reviews, I attended all your comments. Could you review again?

mrp_production_request/models/stock_rule.py
bealdav: OK, thanks

bealdav (approved):
Thanks a lot @lreficent 

It seems ok for me, but I don't know why you don't keep suggested commit (changes squash into only one by author).
Not easy to distinguish changed work and lost different authors work. 
Ok to keep few commits but not merged work of different author.
No problem for this PR. 

Really thanks a lot

LoisRForgeFlow:
honestly, I have never used the suggested commit feature, I just added the changes myself and amended to my last commit, but will do next time.

bealdav:
Ok, no problem.

I think it's must have feature. Imply your reviewer, it'll save your time life ;-)

daramousk (approved):
After putting some time on this, I do not see something wrong with it, it works great. As for the failure in the tests it seems to be because of that `self.with_env(self.env(cr=cr))` where it seems that the `WebRequest` does not have a uid (which makes sense in the testing context)

LoisRForgeFlow:
Thanks all for your reviews! :smiley: 

/ocabot merge

PR #407, opened 2019-09-27
[12.0][FIX] repair_refurbish: do not set destination location for no refurbish repair

AaronHForgeFlow:
repair_refurbish fails when the repair order is not refurbish. Basically, if the repair is not refurbish we are overriding incorrectly the destination location of the stock move. 

The fix is quite dumb, but seems to work. We only change the destination location for the refurbish move.

### DIFF
--- a/repair_refurbish/models/repair.py
+++ b/repair_refurbish/models/repair.py
@@ -64,7 +64,8 @@ class RepairOrder(models.Model):
     @api.multi
     def action_repair_done(self):
         res = super(RepairOrder, self.with_context(
-            force_refurbish_location_dest_id=self.location_dest_id.id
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
ping @lreficent @hveficent

LoisRForgeFlow:
As usual, could you write a small test for this?

AaronHForgeFlow:
@lreficent done :smile:  The test seems dummy, but it fails without the new code added

LoisRForgeFlow (approved):
Great, Thanks!!

Could you squash your commits?

HviorForgeFlow (approved):
Code review

AaronHForgeFlow:
@lreficent  done :)

PR #436, opened 2019-11-25
[FIX] mrp_production_putaway_strategy: not work with rating module

kos94ok-3D:
TypeError: message_post() takes 1 positional argument but 2 were given
since: https://github.com/odoo/odoo/commit/d7c24089ee6cee6f728553cd61143c034e4f08d2#diff-a2cd85e51cd48bdb61a17bb7edfcc631


pedrobaeza (approved):
Fast-tracking this recurrent bugfix

PR #446, opened 2019-12-29
[12.0][mrp_multi_level] define a name_search strategy for the product_mrp_area

JordiBForgeFlow:
Otherwise the search is not effective.

cc @LoisRForgeFlow

### DIFF
--- a/mrp_multi_level/models/product_mrp_area.py
+++ b/mrp_multi_level/models/product_mrp_area.py
@@ -7,12 +7,14 @@
 from math import ceil
 
 from odoo import api, fields, models, _
+from odoo.osv import expression
 from odoo.exceptions import ValidationError
 
 
 class ProductMRPArea(models.Model):
     _name = 'product.mrp.area'
     _description = 'Product MRP Area'
+    _rec_name = "product_id"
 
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):

mrp_multi_level/models/product_mrp_area.py
```diff
@@ -7,12 +7,14 @@
...
 
 from odoo import api, fields, models, _
+from odoo.osv import expression
 from odoo.exceptions import ValidationError
 
 
 class ProductMRPArea(models.Model):
     _name = 'product.mrp.area'
     _description = 'Product MRP Area'
+    _rec_name = "product_id"
```
LoisRForgeFlow: A name_get method is already in place. what's the benefit of this?

LoisRForgeFlow: https://github.com/OCA/manufacture/pull/446/files#diff-9ca74319d3d61b1d491d69f9c387dba0R131

LoisRForgeFlow (approved):
I did a little fix to align the search with the display name. LGTM :stuck_out_tongue_closed_eyes:

PR #450, opened 2020-01-02
[11.0][FIX] mrp_multi_level: origin for "mv" was never being set.

LoisRForgeFlow:
Also, improve mrp.move by using picking name if the move is part of a picking.

@ForgeFlow


JordiBForgeFlow:
I added upstream requirements pegging. So, from an mrp move you can trace what were the mrp moves that created the gross requirements downstream.

Added a button in the MRP Inventory tree view to display the associated demand mrp moves, to allow the user to follow the upstream traceability.

LoisRForgeFlow:
@jordibforgeflow the upstream requirements pegging is moved to a separte PR as it need some tweaks and test. See https://github.com/OCA/manufacture/pull/455

HviorForgeFlow (approved):

PR #452, opened 2020-01-08
[12.0][MIG] product_quick_bom: Migration to 12.0

kevinkhao:


product_quick_bom/views/product_view.xml
```diff
@@ -7,7 +7,13 @@
             <field name="arch" type="xml">
                 <xpath expr="//page[@name='sales']" position="after">
                     <page name="bom_line" string="Bom line">
-                        <field name="bom_line_ids">
+                        <field name="display_bom_lines" attrs="{'invisible': 1}"/>
+                        <div attrs="{'invisible': [('display_bom_lines', '=', True)]}"
```
sebastienbeau: I think you can avoid creating the field display_bom_lines and simply set {'invisible': [('bom_count', '<', 2)]}.

You just need to not add again the bom_count inside the smart button

product_quick_bom/models/product.py
```diff
@@ -1,39 +1,47 @@
...
-        help='If your product is manufactured you can select '
-             'here the component to produce them')
+        comodel_name="mrp.bom.line",
+        inverse_name="product_tmpl_id",
+        string="Bom Line",
+        help="If your product is manufactured you can select "
+        "here the component to produce them",
+    )
+
+    # Need this field due to bom_count already being used on smart button,
```
sebastienbeau: not needed, see my comment on view

sebastienbeau (changes requested):

product_quick_bom/views/product_view.xml
kevinkhao: Usage of bom_count was not possible, because "product.product_template_form_view" (which impacts both product.product and product.template) doesn't have bom_count. Now, I've extended product- and template- specific views and it works. Thanks

sebastienbeau (approved):

product_quick_bom/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module was written to create quickly the bom for your product. In the product form there is a new "bom" tab that gives the possibility to add a bom line directly.
```
rvalyi: s/bom/BOM/
(acronyms are better with uppercase characters)

product_quick_bom/models/product.py
```diff
@@ -43,20 +41,18 @@ def create(self, vals):
             record._process_bom_vals(bom_vals)
         return record
 
-    @api.multi
     def write(self, vals):
         """ default_type is in product and bom so _context ovewrite the default
-                value for bom with typ:'product' need to becarefull
+                value for bom with type:'product' need to be careful
```
rvalyi: s/overwrite/overwrites/

rvalyi: the end of the sentence is a bit meaningless also

rvalyi (commented):
just a few English remarks

product_quick_bom/models/product.py
```diff
@@ -0,0 +1,60 @@
...
+            ctx = self._context.copy()
+            ctx.pop("default_type", None)
+            self.with_context(ctx)._process_bom_vals(bom_vals)
+        return res
+
+    def unlink(self):
+        for record in self:
+            if record.bom_ids:
+                record.bom_ids.unlink()
+        return super(ProductTemplate, self).unlink()
```
legalsylvain: I don't understand why this code is here in this module. As far as I understand, it allows to unlink bom, unlinking template, but it should be in a nother module. Don't you think ? (or in Odoo core in fact).

you can also replace ``super(ProductTemplate, self).unlink()`` by ``super().unlink()``

product_quick_bom/readme/DESCRIPTION.rst
legalsylvain: in fact, it's BoM for Bill of Material in Odoo ;-)
(like PoS for Point of Sale)

legalsylvain (changes requested):
Hi. some minor remarks inline. 

Just tested on runbot. I have a question regarding products (template / variant) : 
Use case : 

- open your template "bolt"
- Add 2 bom lines with new created products "test 1" & "test 2" like that : 

image

the BoM is well created. --> OK. 

image

- Open now your Variant "Bolt" : 
The settings is weird : 

image

Is it the normal behaviour expected ?

product_quick_bom/models/product.py
```diff
@@ -0,0 +1,60 @@
...
+        for record in self:
+            if record.bom_ids:
+                record.bom_ids[0].write({"bom_line_ids": vals})
+            else:
+                record.env["mrp.bom"].create(self._prepare_bom_vals(vals))
+
+    @api.model
+    def create(self, vals):
+        bom_vals = self._extract_bom_line(vals)
+        record = super(ProductTemplate, self).create(vals)
```
bealdav:
```suggestion
        record = super().create(vals)
```

product_quick_bom/models/product.py
bealdav:
```suggestion
        return super().unlink()
```

product_quick_bom/tests/test_quick_bom.py
```diff
@@ -0,0 +1,66 @@
...
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+class TestQuickBom(TransactionCase):
+    def setUp(self):
+        # as we test new boms that must not already exist in demo data,
+        # bolt + screw + ply = lamp
+        super(TestQuickBom, self).setUp()
```
bealdav:
```suggestion
        super().setUp()
```

product_quick_bom/models/product.py
```diff
@@ -0,0 +1,60 @@
...
+        if bom_vals:
+            record._process_bom_vals(bom_vals)
+        return record
+
+    def write(self, vals):
+        """ default_type is in product and bom so _context ovewrite the default
+                value for bom with type:'product' need to be careful
+        """
+        bom_vals = self._extract_bom_line(vals)
+        res = super(ProductTemplate, self).write(vals)
```
bealdav:
```suggestion
        res = super().write(vals)
```

bealdav (changes requested):

kevinkhao:
Thanks for your feedback, following remarks I think my module logic was wrong from the start because:
- bom_line_ids already exists in core in v12
- both product_id/product_tmpl_id have completely different meanings for mrp.bom and for mrp.bom.line (respectively, they are components/the origin template of that component, and an additional detail/the product we want to make), thus creating confusion
To fit the spec of this module (quickly create or update the BoM of a particular product), I've made changes to its logic

legalsylvain (approved):
Hi @kevinkhao 
thanks for taking account our remarks.

image

Looks great ! 

On product.product : 

image

on product.template : 

image

All is OK.

2 questions : 
- I'm not a variant expert but I don't understand why there are variant column on both product.product and product.template form. Is it logic ?

- When I try to delete a template, that has a bom, there are an odoo error : 
image
Is it a core ~bug~ feature ? or is it introduced by your module ?

Regards.

kevinkhao:
@legalsylvain right, I didn't see the error on delete, will fix. For the attributes/variants, I tried to go as simple as possible, because of lack of details and knowledge of use case/MRP. For now I will remove them.
@sebastienbeau , could you please help to clarify issues when we deal with product variants:

- do we simplify the process and assume 1 template <=> 1 BoM, and its variants will be using that same BoM ? 
- do we make it more generic and go with the following logic for variants : 
  - product template has BoM with no product_id: variant uses that BoM
  - product template has BoM with product_id: variant uses that BoM in priority
  - product  template has no BoM: 
    - on variant form, create BoM button creates a BoM with template_id = its template, product_id = itself ?
    - on variant form, create BoM button creates a BoM with template_id = its template, product_id = None ?

Please check screenshots for clarification about the BoM's product_id and the variant's BoM creation issue

Screenshot from 2020-02-21 13-53-09

Screenshot from 2020-02-21 13-54-26

- security-side : users with rights on products but not on mrp, do we give product users r/w access on BoM lines, or hide the tab for non-mrp users ?

kevinkhao:
Fixes have been made, for simplicity BoM edition only works on product.template form view, not on product.product.
@legalsylvain yes, it seems this is a core Odoo ~~bug~~ feature unrelated to this module.

bealdav (approved):

PR #466, opened 2020-02-10
[12.0] [ADD] [WIM] mrp_bom_line_description

FFernandez-PlanetaTIC:
Allows to set a description on BoM lines

cc @PlanetaTIC


FFernandez-PlanetaTIC:
Depends on odoo/odoo#34618

PR #473, opened 2020-03-11
[12.0][FWPORT] mrp_multi_level: origin for "mv" was never being set.

LoisRForgeFlow:
Also, improve mrp.move by using picking name if the move is part of a picking.

Forward port of https://github.com/OCA/manufacture/pull/450.

@ForgeFlow

PR #480, opened 2020-03-13
[13.0][FIX] mrp_multi_level: origin for "mv" was never being set.

LoisRForgeFlow:
Also, improve mrp.move by using picking name if the move is part of a picking.

Forward port of https://github.com/OCA/manufacture/pull/473.

@ForgeFlow


AaronHForgeFlow (approved):
:+1: LGTM

HviorForgeFlow (approved):

PR #485, opened 2020-03-20
[12.0][FIX] round icon

jarroyomorales:
Rounded a bit the module's icon to fit better in the `web_responsive` main menu.

image

image


LoisRForgeFlow (approved):

pedrobaeza (changes requested):
You can't remove the source SVG file. Modify that one instead with the round corners.

jarroyomorales:
@pedrobaeza Okay! Where is it used?

pedrobaeza:
It's used for:

- Generating the PNG from it.
- Be used as template for derivative modules that want to use this icon as base for adding a specific icon. Examples:

  https://github.com/OCA/vertical-association/blob/12.0/membership_extension/static/description/icon.svg
  https://github.com/OCA/vertical-association/blob/12.0/membership_initial_fee/static/description/icon.svg
  https://github.com/OCA/vertical-association/blob/12.0/membership_prorate/static/description/icon.svg
- Adapt on each version the style easily from an editable source.

jarroyomorales:
@pedrobaeza Okay, I think it should be okay now

pedrobaeza (approved):

PR #515, opened 2020-07-14
[12.0] Add new module mrp_production_line_group

sergiocorato:
When you have a BOM with many sub-BOM it is possible that a product is present in many of this.
When a MO is created, each sub-BOM of type phantom is exploded adding all products in the MO sequentially.
So the result is that the same product can be repeated in many rows, which can be misleading to the user.

This module add a button which unify the lines with the same product coming from phantom BOM. When this function is used, it is not possible to change the quantity manufactured anymore, as the single line cannot be linked to many bom lines which the initial ones where linked, so the button 'Update' quantity manufactured become invisible.


pedrobaeza:
Please avoid plurals in module name.

sergiocorato:
> Please avoid plurals in module name.

@pedrobaeza done, thanks

mrp_production_line_group/demo/product.xml
```diff
@@ -0,0 +1,132 @@
+<?xml version="1.0"?>
+<!--
```
OpenCode, line 2: I think it's wrong

mrp_production_line_group/__init__.py
```diff
@@ -0,0 +1,3 @@
+
```
OpenCode, line 3: Header in every python file

mrp_production_line_group/tests/test_mrp_production_line_group.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2020 Sergio Corato <https://github.com/sergiocorato>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```
OpenCode: Manifest says AGPL-3 but file header LGPL-3.0

mrp_production_line_group/views/mrp_production.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="inherit_id" ref="mrp.mrp_production_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='is_locked']" position="after">
+                <field name="lines_grouped" invisible="1"/>
+            </xpath>
+            <xpath expr="//header" position="inside">
+                <button type="action"
+                    name="%(mrp_production_line_group.action_production_group_line)d"
+                    string="Group Lines"
+                    attrs="{'invisible': ['|', ('state', 'not in', ('confirmed','planned')), ('id', '=', False), '|', ('lines_grouped', '=', True)]}"/>
```
OpenCode: Can you explain me this domain, please?

mrp_production_line_group/views/mrp_production.xml
```diff
@@ -0,0 +1,22 @@
...
+            </xpath>
+            <xpath expr="//button[@name='%(mrp.action_change_production_qty)d']" position="attributes">
+                <attribute name="attrs">{'invisible': ['|', ('lines_grouped', '=', True)]}</attribute>
```
OpenCode: OR operator is useless in this case

mrp_production_line_group/wizard/change_production_qty.py
```diff
@@ -0,0 +1,17 @@
...
+class ChangeProductionQty(models.TransientModel):
+    _inherit = 'change.production.qty'
+
+    @api.multi
+    def change_prod_qty(self):
+        for wizard in self:
+            production = wizard.mo_id
+            if production.lines_grouped:
+                raise UserError(_(
+                    'This MO has been grouped, so change product qty is no more'
```
OpenCode: I propose to insert production number in error message

OpenCode (changes requested):

mrp_production_line_group/views/mrp_production.xml
sergiocorato: I misunderstood concatenation, put entire domain.

mrp_production_line_group/tests/test_mrp_production_line_group.py
sergiocorato: fixed

mrp_production_line_group/views/mrp_production.xml
sergiocorato: Fixed (moved '|' before)

mrp_production_line_group/wizard/change_production_qty.py
sergiocorato: done!

mrp_production_line_group/demo/product.xml
sergiocorato: fixed

mrp_production_line_group/__init__.py
sergiocorato: fixed

PR #523, opened 2020-08-28
[12.0][ADD] quality_control: Menu for quality triggers

diggy128:
Also adds a named group in form view to allow putting trigger lines
in one group

### DIFF
--- a/quality_control/views/qc_trigger_view.xml
+++ b/quality_control/views/qc_trigger_view.xml
@@ -20,6 +20,8 @@
             <group>
                 <field name="company_id" groups="base.group_multi_company"/>
             </group>
+            <group name="trigger_lines" colspan="2">
+            </group>
             </form>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


quality_control/views/qc_trigger_view.xml
```diff
@@ -28,12 +30,24 @@
         <field name="name">qc.trigger.tree</field>
         <field name="model">qc.trigger</field>
         <field name="arch" type="xml">
-            <tree string="Quality control triggers" editable="bottom">
+            <tree string="Quality control triggers">
```
pedrobaeza, line 44: This shouldn't be removed.

diggy128: When editable is set, the view change button is not visible and puts the tree view in edit mode when clicking instead of opening form view.
Am I missing something here?

pedrobaeza: Yes, and that's the desired behavior for me.

pedrobaeza: If you want to make it different on your new menu, create a new view for that menu.

diggy128: @pedrobaeza I didn't mean to be rude.
I just thought there was not use being able to edit just the trigger name in list view.
Anyway, I made the requested change.

pedrobaeza: No worry, I don't think you were rude. Just debating :)

My point is that on product form, the trigger should be an editable tree, or you will lose usability. That's why I want to preserve this.

diggy128: I totally agree with you but qc_triggers defines its tree view in product_template_qc_form_view inline.
In any case I see your point, if the inline definition of view is removed in the future, the tree in product view should be editable. Better play it safe.

pedrobaeza (approved):

quality_control/views/qc_trigger_view.xml
kos94ok-3D: What do you think about use?
`primary mode`
`sequence > 16`
`<tree position="attributes">`
 ` <attribute name='editable'></attribute>`
`</tree>`

marcelsavegnago (approved):

pedrobaeza:
@diggy128 can you please rebase your branch and see the problem?

diggy128:
@pedrobaeza It does not seem related as it comes from mrp_subcontracting failing tests on picking confirmation.

pedrobaeza:
Let's try again to see if it's something transient:

/ocabot merge minor

PR #527, opened 2020-09-16
[12.0][FIX] mrp_subcontracting: Overwrite active_id context in create()

### LINKED ISSUE #526: [12.0] mrp_subcontracting: picking cannot be validated
Steps to reproduce:
1. Create a subcontract BoM.
image
2. Create a purchase order with the product above and confirm the order.
3. Make sure there are more stock.picking/purchase.order records than mrp.production records.
4. Browse the picking from the purchase order / from the Inventory menu and validate the picking.

Current behavior:
"Record does not exist or has been deleted." is raised and the picking cannot be confirmed.

From Purchase Order:
image

From Inventory menu:
image

TimLai125:
Overwrite the active_id in context when creating the `mrp.product.produce` record in `action_done`
Fixes: #526

### DIFF
--- a/mrp_subcontracting/models/stock_picking.py
+++ b/mrp_subcontracting/models/stock_picking.py
@@ -72,10 +72,9 @@ class StockPicking(models.Model):
                         })
                 else:
                     for move_line in move.move_line_ids:
-                        active_id = production and production.id or False
                         produce = self.env['mrp.product.produce'].with_context(
                             default_production_id=production.id,
-                            active_id=active_id).create({
+                            active_id=production.id).create({
                                 'production_id': production.id,
                                 'product_id': production.product_id.id,
                                 'product_qty': move_line.qty_done,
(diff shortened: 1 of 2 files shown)


mrp_subcontracting/models/stock_picking.py
```diff
@@ -72,8 +72,10 @@ def action_done(self):
                         })
                 else:
                     for move_line in move.move_line_ids:
+                        active_id = production and production.id or False
```
pedrobaeza: You can simplify this:

```suggestion
                        active_id = production.id
```

pedrobaeza (approved):

pedrobaeza:
As this code is backported from Odoo, can this happen also there in v13?

marcelsavegnago (approved):
LGTM

bodedra (approved):

Issue #529, opened 2020-09-17
[13.0] Only one Manufacture Request being created by the  sale.


guimarc-br:
Hello Guys

On  a Project that I am   working we have the expectation to get  one MR for each  product from the sale/order that does not have the quantity and is Manufacture Request.

**model:** mrp_production_request
**file** : stock_rule.py

I made a debug here and check that is created in the  method  

```
def _run_manufacture(self, procurements):
        for procurement, _rule in procurements:
            if self._need_production_request(procurement.product_id):
                return self._run_production_request(
                    procurement.product_id,
                    procurement.product_qty,
                    procurement.product_uom,
                    procurement.location_id,
                    procurement.name,
                    procurement.origin,
                    procurement.values,
                    procurement.company_id,

```
I could see that is  happening due the **_return_** command,  so when find one product that is manufacture request return the method.

So to handle the expectation on  my project  I just removed the  **_return command_**  and now all products that needs a MR are created.

My question  is guys : This is how this should work ?  or its a bug?.  If it is how was designed I can  change on my project as I did , I am just confirming.

pedrobaeza:
Sorry, this is not an Odoo issue, but a question. You should make it on Odoo official forums (https://www.odoo.com/forum/help-1) or mailing lists (https://odoo.com/groups), but not here.

Closing this by the reason above.

Little tip: you must activate MTO route in the product.

LoisRForgeFlow:
@pedrobaeza I think this is actually a bug, I'm reopening the issue.

@guimarc-br Hi, thanks for reporting it. Would you mind opening a Pull Request fixing the issue? Thanks!

guimarc-br:
Hi Folks

Pull Request created : https://github.com/OCA/manufacture/pull/532

Please take a look.

LoisRForgeFlow:
Fixed (https://github.com/OCA/manufacture/pull/549).

PR #532, opened 2020-09-22
[13.0][FIX]  mrp_product_request : stock_rule.py   Allow  to create more than one MR per sale/order for different products.

guimarc-br:
This is a FIX to allow create more MR per sale/order for  different Products.
The current version is only allowed to have one MR per sale/order

https://github.com/OCA/manufacture/issues/529
Follow the discussion.

### DIFF
--- a/mrp_production_request/models/stock_rule.py
+++ b/mrp_production_request/models/stock_rule.py
@@ -116,7 +116,7 @@ class StockRule(models.Model):
     def _run_manufacture(self, procurements):
         for procurement, _rule in procurements:
             if self._need_production_request(procurement.product_id):
-                return self._run_production_request(
+                self._run_production_request(
                     procurement.product_id,
                     procurement.product_qty,
                     procurement.product_uom,
--- a/mrp_production_request/models/stock_rule.py
+++ b/mrp_production_request/models/stock_rule.py
@@ -79,6 +79,14 @@ class StockRule(models.Model):
             )
 
         # create the MR as SUPERUSER because the current user may not
+        # to avoid everytime MR creation
+        existing_MRs = request_obj.search(
+            [('product_id', '=', product_id.id),
+             ('state', '=', 'to_approve'),
+             ('company_id', '=', values['company_id'].id),
+             ('procurement_group_id', '=', values.get('group_id').id)])
+        if existing_MRs:
+            return True
         # have the rights to do it (mto product launched by a sale for example)
         request = request_obj_sudo.create(
             self._prepare_mrp_production_request(


bodedra (approved):

AaronHForgeFlow (approved):

mrp_production_request/models/stock_rule.py
```diff
@@ -116,7 +116,7 @@ def _run_production_request(
     def _run_manufacture(self, procurements):
         for procurement, _rule in procurements:
             if self._need_production_request(procurement.product_id):
-                return self._run_production_request(
+                self._run_production_request(
```
LoisRForgeFlow, line 127: You need to pop the procurements you already covered with production requests, otherwise both an MR and MO will be created.

LoisRForgeFlow (changes requested):

mrp_production_request/models/stock_rule.py
guimarc-br: @LoisRForgeFlow 

Sorry Lois but what do you mean? I have created this PR because I have different products under the procurements. 
For example
procurements[0] : product 1 
procurements[1] : product 2

Then I  remove it to have one more MR for each product under the procurements. 

Could you  please change the code  your suggestion? I did not get or  are you not approving this new  approach?

Thank you so much in advance.

LoisRForgeFlow: This iteration is resolving procurements of some products in a different way than Odoo standard would do(using manufacturing requests), if you after that call super with the exact same set of procurements you will also create MO's for the procurements you have just created MR's in that iteration.

you need to pop (extract from the set of `procurements`) the ones that have already created a MR and then call super with a reduced set of procurements:

`return super()._run_manufacture(procurements_not_resolved_with_manufacturing_request)` (this is a pseudo-code to point you the error, not the fix)

guimarc-br: Hi @LoisRForgeFlow 

Commit created for this

guimarc-br: Hi Lois 

New commit pushed

guimarc-br: Hi @LoisRForgeFlow  Could u please review it ?

guimarc-br: Hey @LoisRForgeFlow 

Could u pleae review it ?

LoisRForgeFlow:
Hi @guimarc-br 

sorry for the delay checking this, your proposal still makes both a MO and an MR for the same procurement, I don't want to make you wait more time and I have proposed the fix here: https://github.com/OCA/manufacture/pull/549.

mrp_production_request/models/stock_rule.py
```diff
@@ -79,6 +79,14 @@ def _run_production_request(
             )
 
         # create the MR as SUPERUSER because the current user may not
+        # to avoid everytime MR creation
+        existing_MRs = request_obj.search(
```
LoisRForgeFlow, line 83: I also do not agree with this, if you are thinking on orderpoints this is already handled in _quantity_in_progress. Maybe I'm missing some detail here? but it seems to me quite custom process if you need this check.

guimarc-br: Hi Lois ... yes you are right , was my mistake and I  checked your PR and agree with your solution.

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/manufacture/pull/549.

PR #540, opened 2020-10-12
14.0 ADD mrp_schedule_state

bealdav:


mrp_schedule_state/models/production.py
```diff
@@ -0,0 +1,68 @@
...
+        for mo in self:
+            # If MO has been scheduled when it was not ready yet (still in
+            # waiting schedule_state, we can jump to schedule state already)
+            if mo.date_planned_start:
+                mo.schedule_state = "scheduled"
+            else:
+                mo.schedule_state = "todo"
+        return True
+
+    def action_ready(self):
```
florian-dacosta: This does not exist in v14 right ?
I think that was called by workflow in v9 when state become ready to produce.
I guess the equivalent would be when Material Availability field become Ready.
But since this state may go from Ready to waiting (which was impossible in v9 and before), I guess it should impact the schedule state when going from ready to waiting, what do you think?

It also would be nice to have a test to check the automatic change from Waiting to TODO or scheduled (by `set_planable_mo`)

florian-dacosta (commented):
It would be nice to split and update the readme (for example the `Ready To Produce` state does not exist anymore

mrp_schedule_state/models/production.py
bealdav: There is components_availability_state in v14

PR #541, opened 2020-10-12
[FIX] when a bom has quantity multiplier, consider it

JordiBForgeFlow:
When the BOM is specified with a certain quantity, then consider it as a factor during the BOM explosion.

image

cc @LoisRForgeFlow @HviorForgeFlow @AdriaGForgeFlow


JordiBForgeFlow:
This is equivalent to https://github.com/odoo/odoo/blob/13.0/addons/mrp/models/mrp_production.py#L599

LoisRForgeFlow:
Just applied pre-commit.

AdriaGForgeFlow (approved):

HviorForgeFlow (approved):

LoisRForgeFlow (approved):
Tested locally, looks good :+1:

PR #544, opened 2020-10-17
ADD 14.0 mrp_workcenter_hierarchical

bealdav:
Selection_142


dreispt (approved):

mrp_workcenter_hierarchical/models/mrp_workcenter.py
```diff
@@ -0,0 +1,61 @@
...
+from odoo import api, fields, models
+
+
+class MrpWorkcenter(models.Model):
+    _inherit = "mrp.workcenter"
+    _parent_name = "parent_id"
+    _parent_store = True
+
+    parent_id = fields.Many2one(
+        comodel_name="mrp.workcenter", string="Parent", index=True, ondelete="cascade"
```
florian-dacosta: I am not sure about the` ondelete="cascade"` here.
Since a bit hard to try to delete it if parent is deleted... I think it would be best to leave the workcenter alone, with no parent and the user will delete it if needed what do you think?

florian-dacosta (commented):
Code review, LGTM just one remark

mrp_workcenter_hierarchical/models/mrp_workcenter.py
dreispt: @bealdav ?

bealdav: Resolved

florian-dacosta (approved):

PR #559, opened 2020-11-19
[12.0] [FIX] rename quality_control to quality_control_oca

PieterPaulussen:
There is a name collision with the Odoo enteprise module `quality_control' which blocks developers from using other modules in this repository. (Depending on the structure of the project source of course).

f.i. using this repository with a buildout setup combining both enterprise and manufacture repositories


diggy128:
Have a look at [this](https://github.com/OCA/OpenUpgrade/issues/2211).

marcelsavegnago:
@PieterPaulussen If I'm not mistaken, you need to change the module version and build a migration script with openupgradelib to change the module name in cases where the module is already installed.

PieterPaulussen:
I completely agree @marcelsavegnago. This was just a quick rename so I could deploy this repository to my clients server and the migration is not required.
I'll try to include a decent refactoring.

pedrobaeza:
A renaming can't be done in a stable version, as this will break all the existing installations. This renaming can be performed in 13.0 though. I'm not closing this for not being "crude", but this can't be merged.

PieterPaulussen:
@pedrobaeza I see. Guess we'll have to rename it in 13.0 then.

pedrobaeza:
Yes, it's not the ideal thing, as the module doesn't have visibility in Odoo Apps Store, but stability matters more in this case (it's not something that can be handled through migration scripts). I have noted this in https://github.com/OCA/manufacture/pull/481#issuecomment-730439422

inspiredbusiness:
This breaks an Odoo instance with Odoo Enterprise and this repo and Odoo will not boot if enterprise quality_control is installed. I understand it has stable issues with a rename, but also means we can not use any module in this repo without renaming this module and have Odoo boot up. I wonder if future OCA modules should have an oca suffix to stop this from being able to happen when Odoo clashes on an oca name

pedrobaeza:
@inspiredbusiness Don't include such modules on the same "build".

PR #561, opened 2020-11-26
[12.0][FIX] mrp_subcontracting: Skip processed MO when validating receipt

### LINKED ISSUE #560: [12.0] mrp_subcontracting: Receipt cannot be validate if MO is processed first
Steps to reproduce:
1. Create a subcontract BoM with material that is tracked by lot, e.g. Test1 is tracked by lot.
image
image
2. Create a PO with the product and confirm the order.
3. Process the MO first (state: done).
4. Validate The receipt of the PO.

Current behavior:
Odoo Server Error is raised and the picking cannot be validated.
image

TimLai125:
Check state of the MO before trying to process it, skip processed MO when validating receipt.
Fixes: #560

### DIFF
--- a/mrp_subcontracting/models/stock_picking.py
+++ b/mrp_subcontracting/models/stock_picking.py
@@ -49,9 +49,9 @@ class StockPicking(models.Model):
         productions = self.env['mrp.production']
         for picking in self:
             for move in picking.move_lines:
-                if not move.is_subcontract:
-                    continue
                 production = move.move_orig_ids.mapped('production_id')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This code is from odoo, so maybe Odoo v13 is affected as well by this?

pedrobaeza (approved):

TimLai125:
@pedrobaeza Yes, I just noticed https://github.com/odoo/odoo/issues/61927 but I think Odoo has not backport the fix from 14.0 to 13.0.

yostashiro (approved):
LGTM 👍

PR #566, opened 2020-12-04
[14]Add mrp_restrict_lot

florian-dacosta:
Depends on https://github.com/OCA/stock-logistics-workflow/pull/748

The module has no other dependency and is not really usefull by itself, but it it becomes usefull with other modules installed.
It is the case of sale_order_lot_selection, so the selected lots are propagated to the manufacturing orders

It has no dependency with sale_order_lot_selection because it could be used with any other module generating MO with lots, not especially from sale orders.

@sebastienbeau @bealdav

### DIFF
--- a/mrp_restrict_lot/__manifest__.py
+++ b/mrp_restrict_lot/__manifest__.py
@@ -1,0 +1,17 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "MRP Restrict Lot",
+    "version": "14.0.1.0.0",
+    "category": "Manufacturing",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/manufacture",
+    "license": "AGPL-3",
+    "maintainers": ["florian-dacosta"],
+    "development_status": "Beta",
+    "depends": [
+        "stock_restrict_lot",
+        "mrp",
+    ],
+    "installable": True,
+}
--- a/mrp_restrict_lot/models/stock_rule.py
+++ b/mrp_restrict_lot/models/stock_rule.py
@@ -1,0 +1,37 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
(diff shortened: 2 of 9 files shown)


mrp_restrict_lot/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Florian da Costa <florian.dacosta@akretion.com>
+* Valentin Chemiere <valentin.chemiere@akretion.com>
```
bealdav:
```suggestion
* Valentin Chemiere
```

mrp_restrict_lot/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "category": "Manufacturing",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/manufacture",
+    "license": "AGPL-3",
+    "depends": [
+        "stock_restrict_lot",
+        "mrp",
+    ],
+    "installable": True,
+}
```
bealdav, line 17: missing keys : maintainers, development_status

mrp_restrict_lot/models/stock_rule.py
```diff
@@ -0,0 +1,37 @@
...
+            location_id,
+            name,
+            origin,
+            company_id,
+            values,
+            bom,
+        )
+        lot_id = values.get("restrict_lot_id")
+        if lot_id:
+            vals["lot_producing_id"] = lot_id
```
bealdav, line 34: Just a question. "lot_producing_id" is a key used in another module or only here ?

florian-dacosta: That's the lot field on MO, from main mrp module : https://github.com/odoo/odoo/blob/14.0/addons/mrp/models/mrp_production.py#L104

mrp_restrict_lot/__manifest__.py
florian-dacosta: not mandatory I guess, but I've added it anyway, thanks

bealdav (approved):
LGTM thanks

bealdav:
cc @dreispt @LoisRForgeFlow

PR #572, opened 2020-12-11
[FIX] Update maintainer-tools, fix TravisCI builds

dreispt:
As per OCA/oca-addons-repo-template#44 (comment)


dreispt:
As per OCA/oca-addons-repo-template#44 (comment)

PR #583, opened 2020-12-29
[WIP][14.0][MIG] mrp_multi_level

DavidBForgeFlow:
Standard migration to v14. Depends on #582 

cc @LoisRForgeFlow


mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -0,0 +1,740 @@
...
+        self.env["mrp.move"].search(domain).unlink()
+        self.env["mrp.inventory"].search(domain).unlink()
+        domain += [("fixed", "=", False)]
+        self.env["mrp.planned.order"].search(domain).unlink()
+        logger.info("End MRP Cleanup")
+        return True
+
+    @api.model
+    def _low_level_code_calculation(self):
+        logger.info("Start low level code calculation")
```
bealdav, line 286: @DavidBForgeFlow  Thanks to port this module. 
I think it could be interesting to extract this function to a standalone module because it has few dependencies (only llc field) and this concept could be used in many other uses.

i.e: if you want to recompute cost price of all your products, it's required to begin by those with a higher level code first

Also in a separate module it could "spread the word" to more people of this fundamental concept. (module name could be mrp_low_level_code or any other)

Ok that's another task than migration and may could be done in another module but I need to know if others are on the same line than me.

What do you think ? cc @LoisRForgeFlow @hparfr @dreispt @florian-dacosta @pedrobaeza

PR #585, opened 2021-01-09
[12.0][ADD] Module to preserve product procure_method with phantom bom

sergiocorato:
By default Odoo, when a product with bom is ordered having a sub-bom of type phantom, put in all move of components the same route of product, like MTO.
With this module, the route of components is preserved, so if they do not have MTO it remains nothing, so MTS.


JordiBForgeFlow:
@sergiocorato I think that it would be more accurate that the module would search for a valid rule for this component, in the location where the stock move is triggered, instead of assuming that it it MTS by default.

sergiocorato:
> @sergiocorato I think that it would be more accurate that the module would search for a valid rule for this component, in the location where the stock move is triggered, instead of assuming that it it MTS by default.

@JordiBForgeFlow thanks, I updated PR title, I put wrongly `rule` instead of `procure_method`.
As visibile [here](https://github.com/odoo/odoo/blob/7c5186e5bebfe43c70afe3d3895d37cf6d568344/addons/stock/models/stock_warehouse.py#L618) there are only two possibilities: `make_to_stock` or `make_to_order`. (mto/mts rule is in another module, I'll evaluate this one too, perhaps in a separate module)

PR #587, opened 2021-01-12
[13.0][FIX] To be able to  have the WO created automatically as MR created manually.

guimarc-br:
Hi Folks,

When the MR is created and then we created the MO, the code checks  the product and the states from the components and then create automatically the WO.

When  we load one order and uses this creating of MR , the MR  is created and then when the MO is created does not check the states of components only the Product.

This is happening due this procurement_id on the MR object, so once we remove it from the object the states  is checked by the Products and components and then the MR/MO   created by the order will have the same behavior when created manually.


guimarc-br:
Hello @LoisRForgeFlow 

This  build failed on 

test_01_manufacture_request

Due I removed the procurement_group_id  - Could u please take a look ?

PR #591, opened 2021-01-29
[IMP] mrp_subcontracting: Partial subcontracting receptions

ioans73:
We have had some case in which we need to partially receive a subcontracting purchase order, but current code does not allow it because if there is no quantity in the picking line, Odoo blocks the process (https://github.com/odoo/odoo/blob/12.0/addons/mrp/wizard/mrp_product_produce.py#L55). 

Another case, if we, for some reason, cancel the production order and the relations between movements and it are lost, 
it also blocks the process.

### DIFF
--- a/mrp_subcontracting/models/stock_picking.py
+++ b/mrp_subcontracting/models/stock_picking.py
@@ -50,7 +50,12 @@ class StockPicking(models.Model):
         for picking in self:
             for move in picking.move_lines:
                 production = move.move_orig_ids.mapped('production_id')
-                if not move.is_subcontract or production.state in ('done', 'cancel'):
+                if (
+                    not move.is_subcontract or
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is a backport of v13 code, so... is this problem also in that version? If not, you should check the changes that have come on core from the merge date to now for allowing it. If yes, then you should propose the patch to Odoo core for not having again the constraint on next version.

jado95:
@ioans73 I have the error that you linked also with this PR.

jado95 (changes requested):
Can you add in if condition also `or move.quantity_done < move.product_uom_qty` for partially reception?

ioans73:
We are going to proceed to close this PR, since finally the functionality that our client needs is different and we've taken another approach

PR #611, opened 2021-03-01
FIX - If product for subcontracting have track serial/lot and have co…

jado95:
…mponent with no track, now the button 'Record Components' is displayed and also if click on stock move of picking, is displayed wizard for record production of product(also field serial/lot)


sergiocorato (commented):
Thanks.
Can you add test?

pedrobaeza:
The icon/button is intended to be shown only if there are components with lots/serial numbers, but if not, the screen is simplified and the subcontracting is done anyway, so the current state is correct.

PR #618, opened 2021-03-05
[14.0][MIG] repair_calendar_view: Migration to 14.0

Reyes4711-S73:
Standard migration to 14.0


marcelsavegnago:
i think its not necessary.. base_repair replace repair_calendar_view

marcelsavegnago (changes requested):
I think this is no longer necessary because the repair_base module replaces repair_calendar_view

I even intend to suggest removing repair_calendar_view from branch 12 and 13

marcelsavegnago:
@Reyes4711 I apologize for insisting, but I recommend that you consider using the base_repair module. I migrated to 13 recently and am migrating to 14 (PR #621 If you can review this PR I would appreciate it). 

Honestly, I understand that the repair_calendar_view module should not go to version 14, because its coexistence with the base_repair module will generate duplication mainly for those who use the runbot for testing.

I chose to work on a basic improvement module when I realized that modules like repair_calendar_view can cause problems when creating and installing other modules that add new views (for example, pivot table, timeline) depending on the order of installation of the modules.

Reyes4711-S73:
Sorry @marcelsavegnago. Now we must use this module for once client. I did the modifications that you done in version 12.0 of this module but doesn't appear in version 13.0. I review your PR

marcelsavegnago:
@Reyes4711 On your client you can install base_repair and then uninstall repair_calendar_view. Since the repair_calendar_view fields are within base_repair it is likely that the data remained persistent. Confirm on a test basis first.

PR #622, opened 2021-03-10
12.0 add mrp production unreserve line

MPoch-PlanetaTIC:
Permits to unreserve lines of mrp production


ypapouin (changes requested):
Same comment that https://github.com/OCA/stock-logistics-workflow/pull/809
Don't you think that a specific security group could be added for this ?

PR #640, opened 2021-04-30
[12.0][FIX] mrp_subcontracting: fix action_done

Borruso:
If I make a subcontractor order with more product lines when I receive products and I receive only one of products and I create backorder the action_done function fails because in the other products I have zero as qty_done

### DIFF
--- a/mrp_subcontracting/models/stock_picking.py
+++ b/mrp_subcontracting/models/stock_picking.py
@@ -72,17 +72,18 @@ class StockPicking(models.Model):
                         })
                 else:
                     for move_line in move.move_line_ids:
-                        produce = self.env['mrp.product.produce'].with_context(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Hi @Borruso, thanks for the PR. You are right about the problem, but I have fixed it in a better way in 4875fd4, for avoiding rounding problems and also covering the other part of the if.

Borruso:
> Hi @Borruso, thanks for the PR. You are right about the problem, but I have fixed it in a better way in [4875fd4](https://github.com/OCA/manufacture/commit/4875fd46eb11449df6ce16aa2d01a48e90eb2b4b), for avoiding rounding problems and also covering the other part of the if.

GJ, I’m glad I was able to help

PR #646, opened 2021-05-17
[14.0] [NEW] product_cost_rollup_to_bom: New module to update the standard cost of the product.

Chandresh-OSI:


Chandresh-OSI:
CC @dreispt

pedrobaeza:
Please don't abbreviate `product` to `prod` in the module name.

dreispt:
@Chandresh-OSI Thank you
- I believe the original authors are other people, not mentioned. Can you add them?
- Can you replace the hardcoded email with an email template? The Config options can then go away, as the Email Template is the config.

product_cost_rollup_to_bom/views/product_views.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" ?>
+<odoo>
+    <!-- product product form view -->
+    <record id="view_std_price_product_product" model="ir.ui.view">
+        <field name="name">view.std.price.product.product</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view" />
+        <field name="arch" type="xml">
+            <field name="company_id" position="before">
+                <field name="std_cost_update_date" readonly="1" />
```
dreispt, line 10: Also needed: the Product from `<button name="button_bom_cost" string="Compute Price from BoM" ...>` should be visible also for standard price products.

product_cost_rollup_to_bom/models/product.py
```diff
@@ -0,0 +1,135 @@
...
+    def _set_price_from_bom(self, boms_to_recompute=False):
+        self.ensure_one()
+        bom = self.env["mrp.bom"]._bom_find(product=self)
+        if bom:
+            self.standard_price = self.with_context(cost_all=True)._compute_bom_price(
+                bom, boms_to_recompute=boms_to_recompute
+            )
+            bom.std_cost_update_date = datetime.now()
+
+    def _compute_bom_price(self, bom, boms_to_recompute=False):
```
dreispt, line 76: This is overriding a core method. That's not good, I'm not sure we can use this...

dreispt: @Chandresh-OSI Let's start my having an email template instead of an email config, and have the Update From BoM button available also for Standard Cost products. Then we'll see f any of the compute price code s really needed.

product_cost_rollup_to_bom/views/product_views.xml
Chandresh-OSI: It will be visible for standard price costing method products.
`attrs="{'invisible': ['|', ('bom_count', '=', 0), '&amp;', ('valuation', '=', 'real_time'), ('cost_method', '=', 'fifo')]}"`
attrs given in the base.

PR #655, opened 2021-07-05
[13.0][mrp_kit_drop_ship] Kits behave as normal products when drop shipped

JordiBForgeFlow:
MRP Kit Drop Ship
==============
Without this module, if a product is being procured and has an existing Bill of
Materials of type Kit, the components of the product are going to be
procured, instead of the main product.

This module introduces the capability to ignore the existence of the kit
during procurement, when the procurement is associated with a dropship. That
is, when the product is going to be sourced from a third party vendor to the
customer.

Upon confirmation of the purchase order the system will still explode the
product into components within the drop ship picking.

cc @JordiMForgeFlow


thomaspaulb:
Hey @JordiBForgeFlow, our client is running into this exact problem. I was about to suggest building this module, which you apparently already did. But, you closed it same day - did you figure out a better way to tackle this?

PR #608, opened 2021-02-24
[ADD] mrp_account_analytic_wip

dreispt:
Depends on:
- [x] mrp_analytic_cost #607 (bundled!)
- [ ] account_analytic_wip https://github.com/OCA/account-analytic/pull/360

Supersedes #599 and #605:
The previous `mrp_account_material_wip` and `mrp_account_work_order_wip` modules are merged into a single one `mrp_account_analytic_wip`.

PR #656, opened 2021-07-07
[ADD] mrp_account_analytic: generate Analytic Items for manufacturing consumptions.

dreispt:
Extracted from the original proposed code for mrp_account_analytic_wip.
Has only the essential code for Analytic Item generation. 

Extracted from #608, that should be refactored to depend on this.

### DIFF
--- a/mrp_account_analytic/models/account_analytic_line.py
+++ b/mrp_account_analytic/models/account_analytic_line.py
@@ -1,0 +1,21 @@
+# Copyright (C) 2020 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    manufacturing_order_id = fields.Many2one(
+        "mrp.production",
+        string="Related Manufacturing Order",
+    )
+    stock_move_id = fields.Many2one(
+        "stock.move",
+        string="Related Stock Move",
+    )
+    workorder_id = fields.Many2one(
+        "mrp.workorder",
+        string="Work Order",
+    )
--- a/mrp_account_analytic/README.rst
+++ b/mrp_account_analytic/README.rst
@@ -1,0 +1,1 @@
+To generate
--- a/mrp_account_analytic/__init__.py
+++ b/mrp_account_analytic/__init__.py
@@ -1,0 +1,4 @@
+# Copyright (C) 2021 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import models
--- a/mrp_account_analytic/__manifest__.py
+++ b/mrp_account_analytic/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright (C) 2021 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Manufacturing Analytic Items",
+    "summary": "Consuming raw materials and operations generated Analytic Items",
+    "version": "14.0.1.0.0",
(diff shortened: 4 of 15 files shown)


mrp_account_analytic/models/account_analytic_line.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    manufacturing_order_id = fields.Many2one(
+        "mrp.production",
+        string="Related Manufacturing Order",
+    )
+    stock_move_id = fields.Many2one(
```
anajuaristi, line 14: Hi Daniel. I would like to give you a couple of ideas that I think could help you to simplify the code a little bit. 
Since stock_move is related to mrp.production and mrp.workorder,  they could be taken from stock move in analytic line, directly
This is.. each stock_move (from order_id.raw_material_ids) is mandatory linked to an order_id and could be also linked to a workorder_id, so.. you would not need to create functions on each object. You could just link the stock_move to an analytic line and update values from him.

Fields order_id, workorder_id can not be directly related to stock_move, just because if you think about including workorders(operations) time cost in analytic on future, you will need both fields free to assign a value. 

But... I think both fields could be computed. Very easy. If analytic line is linked to stock, move then take from him. If it's linked to workorder.time_ids, take directly from him. 

I think this couple of changes could allow you to reduce the amount of code lines. Anyway, if you need me to test functionally the module, let me know!
Thank you for the contribution.
Ana

dreispt: Hello Ana, thank you for your feedback.
Indeed the `manufacturing_order_id` here could be computed from the `stock_move_id`or the `workorder_id`, instead on being directly set.

I felt the current approach would give more flexibility to developers, and changing it would remove only two lines of code, to be replaced by a computed method. I'm not sure the change would make the code simpler, but I'm open for more suggestions.

If you can test this, it would be great.
Thank you

anajuaristi:
Better from stock_move_id, since workorder_id is not mandatory and could be empty. Any raw_stock_move included in manufacturing order will have manufactured_order_id informed but not always will have workorder_id, depending on BOM configuration. BUT it's better not making it related since on future you could think about creating also analytic lines from machines working time as said before.

My previous suggestion was... creating the functions directly on stock.move
object instead on manufacturing order or workorder.
You would avoid almost all code built on mrp.production and mrp.workorder.

Just... in stock_move_line
--> On making "done" an stock.move.line, if stock.dest_id is virtual and stock.location_id is internal
          --> if stock_move_line.stock_move_id.mrp.production_id si set -->Then create analytic line for production_id
                    if stock_move_line.stock_move_id.mrp.workorder_id is set--> Set analytic_line.workorder_id with it.

That's all. Just a couple of lines of code.

My2cents.

dreispt:
@anajuaristi 

> Better from stock_move_id, since workorder_id is not mandatory and could be empty. Any raw_stock_move included in manufacturing order will have manufactured_order_id informed but not always will have workorder_id, depending on BOM configuration. BUT it's better not making it related since on future you could think about creating also analytic lines from machines working time as said before.

Creating analytic lines from machine working time is also implemented here.
That's why the `workorder_id` field is there.
For raw material consumption, the analytic item `stock_move_id` is set.
For working time, the analytic items `workorder_id` is set.
It is either one or the other.  

> My previous suggestion was... creating the functions directly on stock.move
object instead on manufacturing order or workorder.
You would avoid almost all code built on mrp.production and mrp.workorder.

The code is `mrp.workorder` is to track machine working time, and is unrelated to raw material consumption.
So it can't be removed.

Does this make sense?

dreispt:
Superseded by #657 (to simplify code, the WIP Analytic Items and Cost Type breakdown feature were merged together, otherwise dependencies were becoming too complex)

PR #659, opened 2021-07-17
[14.0][ADD] mrp_multiple_serial

remi-filament:
This module allows for producing more than 1 product tracked by S/N


remi-filament:
Hi @sbidoul , would you mind reminding me how to instruct Travis to load a dependent module from another PR ?
I have tried :
- adding the repo to oca_dependencies.txt
- changing travis configuration with MQT_DEP=OCA

The problem is that oca_dependencies are seldomly filled since PIP is used instead of these files and I have to go to each repo to fix missing dependencies in oca_dependencies.txt and I would still need these to be merged for getting Travis working on this PR...

I am sure there is a better way to get this properly, if you may help me ?

pedrobaeza:
Please don't use plurals on module name. You can put it as `mrp_multiple_serial`.

sbidoul:
@remi-filament if you add `odoo14-addon-product_lot_sequence @ git+https://github.com/lefilament/product-attribute@14.0-mig-product_lot_sequence#subdirectory=setup/product_lot_sequence` in `test-requirements.txt`, it should work with `MQT_DEP=PIP`.

Alternatively, you can point to the PR with `odoo14-addon-product_lot_sequence @ git+https://github.com/OCA/product-attribute@refs/pull/886/head#subdirectory=setup/product_lot_sequence`.

[updated to add #subdirectory]

remi-filament:
Thanks Stéphane I have just pushed a new commit to try this solution.
I have also renamed the module and the PR Pedro to remove plural as requested.

remi-filament:
Any idea why runbot is failing ?
There are only 3 lines of logs without errors... 
(I have the same behavior on other PRs created today)

rousseldenis:
> Any idea why runbot is failing ?
> There are only 3 lines of logs without errors...
> (I have the same behavior on other PRs created today)

image

gurneyalex:
hello this module looks interesting. Could you add an automated test ?

remi-filament:
Hi @gurneyalex I have added some tests, it is the first time I write unit tests so let me know if this is enough from your point of view ? (I checked the new behaviour and the old one to make sure it is not broken by this module)

remi-filament:
Hi @OCA/manufacturing-maintainers 
This PR was stale then closed, does that mean it makes no sense for OCA ?
I actually have a second customer with the same need (in v13.0) so I would port it to 13.0 but would not lose time making a PR if it does not make sense for OCA !
Thanks !
Best Regards

PR #666, opened 2021-08-26
[14.0[ADD] mrp_account_analytic: generate Analytic Items for manufacturing consumptions.

dreispt:
Extracted from the original proposed code for mrp_account_analytic_wip

### DIFF
--- a/mrp_account_analytic/README.rst
+++ b/mrp_account_analytic/README.rst
@@ -1,0 +1,1 @@
+To generate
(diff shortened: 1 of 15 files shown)


patrickrwilson:
@dreispt It looks like there are some errors in travis, I'll review after those are resolved. Also in the PR description it's referencing a dependency but the manifest doesn't have that in the depends so maybe the PR description just needs updated?

dreispt:
That is a good point. There were merges after last build, re-running it to see if there are changes.

dreispt:
Re-build turned Travis green 😺

patrickrwilson (approved):
LGTM

PR #669, opened 2021-08-31
[13.0][mrp_multi_level]add a values parameter to _prepare_planned_order_data()

robinhli:


mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -101,7 +101,7 @@ def _prepare_mrp_move_data_from_stock_move(
 
     @api.model
     def _prepare_planned_order_data(
-        self, product_mrp_area, qty, mrp_date_supply, mrp_action_date, name
+        self, product_mrp_area, qty, mrp_date_supply, mrp_action_date, name, values
```
dreispt: This an API breaking change, and will break custom code using it.
```suggestion
        self, product_mrp_area, qty, mrp_date_supply, mrp_action_date, name, values=None
```

dreispt (changes requested):

mrp_multi_level/wizards/mrp_multi_level.py
robinhli: > This an API breaking change, and will break custom code using it.

This is what I want to change. Also, this parameter have a default value, other calls also work even with new code.

PR #680, opened 2021-09-25
[14.0][ADD] repair_purchase_return

JordiBForgeFlow:
Depends on https://github.com/OCA/purchase-workflow/pull/1273

TODO:
* [ ] Add tests

@ForgeFlow


repair_purchase_return/models/purchase_return_order.py
```diff
@@ -0,0 +1,48 @@
...
+    repair_order_count = fields.Integer(
+        compute="_compute_repair_orders",
+        string="Repair Order Count",
+        copy=False,
+        default=0,
+    )
+    repair_order_ids = fields.Many2many(
+        comodel_name="repair.order",
+        compute="_compute_repair_orders",
+        copy=False,
```
dnplkndll: @ChrisOForgeFlow  does this need review to merge? 
```
purchase.return.order: inconsistent 'compute_sudo' for computed fields: repair_order_count, repair_order_ids
```
do we need to override the compute_sudo to False when adding compute?
```suggestion
        copy=False,
        compute_sudo=False,
```

ChrisOForgeFlow: The reason for this issue, is related with definition of computed field, two field have same computed method, but one is stored and another not, because that is the issue, or all are stored, or all are not stored if share method to compute

dnplkndll:
@ChrisOForgeFlow this looks like it needs rebased, still seems relevant in our case.

dnplkndll:
@JordiBForgeFlow do we need to review to get merged? I guess getting https://github.com/OCA/purchase-workflow/pull/1273 merged is a blocker?

PR #683, opened 2021-10-04
[14.0][MIG] quality_control_stock_oca

enriquemartin:
standard migration + "refactor" of ir.model.acess.csv: shorten up the names, fill missing groups and add wizard rule


enriquemartin:
Hi @anajuaristi!
This is ready to be reviewed.
Thanks!

jonerikceberio:
Due to the change in permissions in access_manager_qc_trigger_product_template_line_user and access_manager_qc_trigger_product_line_user ([v13](https://github.com/OCA/manufacture/blob/13.0/quality_control_oca/security/ir.model.access.csv#L15) has no group, [v14](https://github.com/OCA/manufacture/blob/14.0/quality_control_oca/security/ir.model.access.csv#L16) has group group_quality_control_user to be consistent), now a stock user has no rights to see those models. 
So the [action_done ](https://github.com/Digital5-Odoo/manufacture/blob/8ff3402ca2cdd030ea2c7f04ae856f4c31db0427/quality_control_stock_oca/models/stock_picking.py#L65)cannot be done. The solution is to give reading access to stock_user

pedrobaeza:
Or to do the operation with sudo.

jonerikceberio:
@pedrobaeza 
That was plan B, as it wouldn't be a standard migration. Do the stock users need to read the triggers? Not really, with the inspection and inspection lines should be enough. However when trying it, the `self.env[model].sudo().get_trigger_line_for_product(` part kept on giving permission errors. Don't know why, maybe because of the inheritance?

pedrobaeza:
Maybe the sudo should be in another part. Always that you do `self.env["model"]` in the code, you need to do the `sudo` there, as it's not transferred from current status.

jonerikceberio:
Got it, it was the triggers gotten from the product's category, so the product has to be sudoed. All in all is rather dirty, with three sudos. But they take out five model acceses, so both solutions work.

>     def _action_done(self):
        res = super()._action_done()
        inspection_model = self.env["qc.inspection"]
        qc_trigger = self.env["qc.trigger"].sudo().search(
            [("picking_type_id", "=", self.picking_type_id.id)]
        )
        for operation in self.move_lines:
            trigger_lines = set()
            for model in [
                "qc.trigger.product_category_line",
                "qc.trigger.product_template_line",
                "qc.trigger.product_line",
            ]:
                partner = self.partner_id if qc_trigger.partner_selectable else False
                trigger_lines = trigger_lines.union(
                    self.env[model].sudo().get_trigger_line_for_product(
                        qc_trigger, operation.product_id.sudo(), partner=partner
                    ))
            for trigger_line in _filter_trigger_lines(trigger_lines):
                inspection_model._make_inspection(operation, trigger_line)
        return res

Daniel-CA (approved):
LGTM

oihane (approved):
LGTM

PR #684, opened 2021-10-04
[14.0][FIX] mrp_production_serial_matrix: use qty reserved for selected MO.

LoisRForgeFlow:
User should be able to use the matrix when the MO is reserved and
select the lots that are available plus the ones reserved
specifically for the MO.

@ForgeFlow


AaronHForgeFlow (approved):
Makes sense :+1:

MiquelRForgeFlow (approved):
LGTM

PR #688, opened 2021-10-13
[FIX] mrp_subcontracting_purchase_link: action button with sudo

clementmbr:
Avoid security block for users with no "Admin/Settings" rights using `_for_xml_id()` instead of direct `read()`.
(very annoying)


clementmbr:
Duplicate with #739

PR #690, opened 2021-10-15
[14.0] [mrp_multi_level][imp] improve traceability

JordiBForgeFlow:
- The description of the planned order includes the topmost requirement that
  caused the planned order. This makes it easier to trace, for example,
  what planned orders has a given sales order generated. The description
  of the planned order is passed on to the Manufacturing Orders / Purchase Orders / Transfers
  as the origin, so it can be possible to search for PO's / MO's that
  were originated as a result of a given sales order, for example.

- The MRP Inventory tree view is improved so as to add a button to allow you to
  jump to the planned orders.

@ForgeFlow @LoisRForgeFlow

### DIFF
--- a/mrp_multi_level/models/mrp_inventory.py
+++ b/mrp_multi_level/models/mrp_inventory.py
@@ -6,7 +6,7 @@
 
 from datetime import date, timedelta
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow:
Superseded by https://github.com/OCA/manufacture/pull/806

PR #703, opened 2021-10-31
[15.0][MIG] mrp_bom_location

dnplkndll:


LoisRForgeFlow:
@dnplkndll you should have 2 commits, one for the pre-commit stuff and other for the actual migration (["[MIG] $MODULE: Migration to 15.0"](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0#technical-method-to-migrate-a-module-from-140-to-150-branch)), in this case this second commit would only contain the version bump.

dnplkndll:
I was unsure since the version check  was in precommit. But I guess just matters that it passes in whole on the pr. Thanks

Don Kendall
________________________________

LoisRForgeFlow:
The idea of the command `git commit -m "[IMP] $MODULE: black, isort, prettier"  --no-verify` is that you can commit even if there are issues in the precommit (no-verify optoin), then you can do what it is the actual migration task.

LoisRForgeFlow (approved):
it would be good if the commit history is aligned with guidelines but not a blocking thing - the migration is quite straight forward.

PR #711, opened 2021-11-17
[13.0][fix] mrp_multi_level: update supply_method

JordiBForgeFlow:
Force to update the supply method in mrp.inventory even when is a
computed field.

cc @ForgeFlow @LoisRForgeFlow


MiquelRForgeFlow:
conflicts

MiquelRForgeFlow (approved):

ivantodorovich (approved):

LoisRForgeFlow:
@JordiBForgeFlow could you describe the bug being solved?

LoisRForgeFlow (changes requested):
I tested in roboat and it seems the field is being filled.

image

Waiting for further description of the issue

PR #721, opened 2021-12-14
[14.0][IMP] mrp_multi_level_estimate: do not require date ranges on stock demand estimates

NuriaXifre:
This improvement disassociates date ranges from stock demand estimates when running MRP Inventory. Before, it was compulsory that the Stock Demand Estimate had a Date Range associated to it. With this improvement the MRP takes as start and end date the fields specific of the Stock Demand Estimate. 

@LoisRForgeFlow @ForgeFlow


LoisRForgeFlow (approved):
Indeed this can be done since the dependency on date ranges was removed in estimates.

LGTM

LoisRForgeFlow:
@NuriaXForgeFlow could you just extend a bit your commit message to describe the change on it?

JordiBForgeFlow (approved):

PR #730, opened 2022-01-01
[ADD] mrp_production_filter_lot

monen17:
Adding module to filter selectable lots during production, based on their location and availability


mrcast (approved):
Functional test ok

PR #735, opened 2022-01-13
[14.0][ADD] repair_sale_order

ChrisOForgeFlow:
Implement a way to make sales order to sell and delivery spare parts from repair order
@ForgeFlow


LoisRForgeFlow:
@ChrisOForgeFlow could you move this to the new specific repo for repairs? https://github.com/OCA/repair

LoisRForgeFlow:
> @ChrisOForgeFlow could you move this to the new specific repo for repairs? https://github.com/OCA/repair

well, actually I think v14 will still be in manfuacture, the transition will be done in v15. Sorry for the noise, to be considered on migration though :)

AaronHForgeFlow (changes requested):
Overall it looks like a very useful module. I have a kind of an issue I think.

I created a repair type for sales. And then I created a repair order for a product and created the sale quotation from the repair order. I delivered the picking form the sales order and the repair order was set to "Repaired", that is fine. However, it seems those unwanted stock moves were created, aside from the picking of the sales order:

image

As far as I understand those stock moves are created automatically in the repair order. However, I cannot control the location for those stock moves, because in the repair type the location fields are hidden if I select "Create Sales Order". Also, the status of those stock moves are open after the repair order is set to "Repaired".

So, IMO, I would prevent the creation of those stock moves in this type of repair order. And I think if we can link the stock move of the picking of the sales order to the stock moves button in the repair order, that would be awesome. WDYT?

ChrisOForgeFlow:
> Overall it looks like a very useful module. I have a kind of an issue I think.
> 
> I created a repair type for sales. And then I created a repair order for a product and created the sale quotation from the repair order. I delivered the picking form the sales order and the repair order was set to "Repaired", that is fine. However, it seems those unwanted stock moves were created, aside from the picking of the sales order:
> 
> image
> 
> As far as I understand those stock moves are created automatically in the repair order. However, I cannot control the location for those stock moves, because in the repair type the location fields are hidden if I select "Create Sales Order". Also, the status of those stock moves are open after the repair order is set to "Repaired".
> 
> So, IMO, I would prevent the creation of those stock moves in this type of repair order. And I think if we can link the stock move of the picking of the sales order to the stock moves button in the repair order, that would be awesome. WDYT?

This behavior is for this module

https://github.com/OCA/manufacture/blob/2b97a951241bf6d9878aecaaea4b0194d2777956/repair_stock_move/models/repair_order.py#L81

@LoisRForgeFlow Do you think this two modules should be compatible?

MiquelRForgeFlow:
please, could you rebase?

AaronHForgeFlow:
Rebased again to restart builds

repair_sale_order/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Repair To Sale Order",
+    "version": "14.0.1.0.0",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/manufacture",
+    "summary": "Repair To Sale Order",
+    "category": "Repair",
+    "depends": ["repair", "repair_type", "sale", "sales_team"],
```
AaronHForgeFlow, line 11: I think this should depend on repair_type and sale only, the others  would be implied dependencies

repair_sale_order/models/repair_order.py
```diff
@@ -0,0 +1,106 @@
...
+                    line._get_sale_line_data(sale_order)
+                )
+                line.sale_line_id = sale_order_line.id
+        return self.action_show_sales_order(orders)
+
+    def action_validate(self):
+        if self.filtered(lambda x: x.create_sale_order and not x.operations):
+            raise UserError(
+                _(
+                    "You should input almost one part line to"
```
AaronHForgeFlow:
```suggestion
                    "You should input at least one part line to"
```

AaronHForgeFlow (commented):
Sorry for the late update here. My previous review is incorrect, so you can disregard it.

It looks good to me. Just a couple of minor comments and a question. I would like to include in the sales order also the possibility of include the operations services in the sales quotation. This is because the following use case:

- Customer returns the product out of warranty
- We spend time to investigate why it is broken
- We sell the spare parts so the customer can repair it but we also charge the cost of investigating the cause of the breakdown. 

Perhaps, if the repair type is sale, we can add another checkbox that says "include Operations". So the services can be included in the sales order as well. WDYT?

repair_sale_order/models/repair_order.py
```diff
@@ -0,0 +1,106 @@
...
+                    (state, view) for state, view in action["views"] if view != "form"
+                ]
+            else:
+                action["views"] = form_view
+            action["res_id"] = orders.ids[0]
+        else:
+            action["domain"] += [("id", "in", orders.ids)]
+            domain = safe_eval(action["domain"]) or []
+            domain += [("id", "in", orders.ids)]
+            action["domain"] = action["domain"]
```
miguel-S73:
```suggestion            
            action["domain"] = [("id", "in", orders.ids)]
```

miguel-S73:
@ChrisOForgeFlow please, fix pre-commit error.

ChrisOForgeFlow:
> @ChrisOForgeFlow please, fix pre-commit error.

Done!

ferran-S73:
@ChrisOForgeFlow could you please rebase to generate runboat again?

bosd (approved):
Functional test, lgtm

AaronHForgeFlow (approved):

LoisRForgeFlow (approved):

PR #739, opened 2022-01-29
[14.0][FIX] Fix access right issues when using actions

sebalix:
Impact:

- `mrp_multi_level`
- `mrp_subcontracting_purchase_link`


simahawk:
we should require tests for actions :smile:

simahawk (approved):

sebalix:
@simahawk tests running with a user other than admin yes

PR #742, opened 2022-02-02
[15.0][FIX] mrp_warehouse_calendar: date_planned_finished calculated with warehouse calendar

BernatPForgeFlow:
Description of the issue/feature this PR addresses:
When creating a MO _date_planned_finished_ is not calculated with warehouse calendar.

Current behavior before PR:
There is not calculation of _date_planned_finished_ taking into account warehouse calendar. 

Desired behavior after PR is merged:
_date_planned_finished_ is calculated taking into account warehouse calendar. 

Steps to reproduce:
1. Add a calendar to the warehouse.

2. Add to a product a Manufacturing Lead Time

3. Create a MO of the product, and see Scheduled End Date. To see Scheduled End Date you have to edit the form view manually in Edit View: Form option.

Calendar:
Captura de pantalla de 2022-02-03 15-29-46

**As we can observe, Scheduled End Date falls on Sunday, which is not taking into account the warehouse calendar**

Using the PR fix, the result would be like this:

### DIFF
--- a/mrp_warehouse_calendar/models/mrp_production.py
+++ b/mrp_warehouse_calendar/models/mrp_production.py
@@ -8,14 +8,25 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     @api.onchange("date_planned_start", "product_id")
-    def onchange_date_planned(self):
-        dt_planned = self.date_planned_start
-        warehouse = self.picking_type_id.warehouse_id
-        if warehouse.calendar_id and self.product_id.produce_delay:
-            date_expected_finished = warehouse.calendar_id.plan_days(
-                +1 * self.product_id.produce_delay + 1, dt_planned
-            )
-            self.date_planned_finished = date_expected_finished
+    def _onchange_date_planned_start(self):
+        res = super(MrpProduction, self)._onchange_date_planned_start()
+        if self.date_planned_start and not self.is_planned:
+            warehouse = self.picking_type_id.warehouse_id
+            if warehouse.calendar_id:
+                if self.product_id.produce_delay:
+                    self.date_planned_finished = warehouse.calendar_id.plan_days(
+                        +1 * self.product_id.produce_delay + 1, self.date_planned_start
+                    )
+                if self.company_id.manufacturing_lead:
+                    self.date_planned_finished = warehouse.calendar_id.plan_days(
+                        +1 * self.company_id.manufacturing_lead + 1,
+                        self.date_planned_finished,
+                    )
+                self.move_finished_ids = [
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
Code review LGTM, thanks for the explanation!

AaronHForgeFlow:
@JoanMForgeFlow can you review as well?

JoanMForgeFlow (approved):
Functional review, works as desired. Code review OK!
Thanks for the fixing.

mrp_warehouse_calendar/models/mrp_production.py
```diff
@@ -28,3 +28,38 @@ def copy(self, default=None):
...
+            product_uom_qty,
+            product_uom,
+            operation_id,
+            byproduct_id,
+            cost_share,
+        )
+        warehouse = self.picking_type_id.warehouse_id
+        if warehouse.calendar_id:
+            if self.product_id.produce_delay:
+                self.date_planned_finished = warehouse.calendar_id.plan_days(
```
LoisRForgeFlow: A hook method to get the finished move values is not a good place to change the date_planned_finished of the MO. The proper place to do this fix is probably at `_onchange_date_planned_start` method.

Also, can you add a test case that fails before the fix and it is fixed with it? this will help to avoid regressions in the future

LoisRForgeFlow (changes requested):
The issue analysis is perfect :+1: I have some comments on the fix impelmentation:

mrp_warehouse_calendar/models/mrp_production.py
```diff
@@ -8,14 +8,29 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     @api.onchange("date_planned_start", "product_id")
-    def onchange_date_planned(self):
```
BernatPForgeFlow, line 11: fsdfds

mrp_warehouse_calendar/models/mrp_production.py
BernatPForgeFlow: @LoisRForgeFlow 

Checking the tests I noticed that there is already a test that checks this but it was calling an unused function.

AaronHForgeFlow: @LoisRForgeFlow can you check the new code?

MiquelRForgeFlow (approved):

LoisRForgeFlow:
I just removed the call to plan_hours because it would have given error (it expect int not datetime), but it was also dead code as in the super one hour of differences is ensured, so this code was never executed.

LoisRForgeFlow (approved):
LGTM

PR #748, opened 2022-02-08
[14.0][ADD] mrp_bom_recursion_restriction

ChrisOForgeFlow:
Module adds possibility to filter boms with recursive structure between their components
@ForgeFlow


BernatPForgeFlow:
Really nice module! Maybe it would be nice if you also add a log message like "The recursion in this BOM is solved" when there is no recursion anymore with that BOM. WDYT?

ChrisOForgeFlow:
> Really nice module! Maybe it would be nice if you also add a log message like "The recursion in this BOM is solved" when there is no recursion anymore with that BOM. WDYT?

Sure, nice feature, currently I'm testing on customer, and adding some test cases, after that we can add

LoisRForgeFlow (approved):
Tested locally, works smoothly. It is also good that you include an init hook. :+1:

PR #759, opened 2022-02-21
[14.0][FIX] mrp_multi_level: Get MRParea company_id for supply method

BernatPForgeFlow:
Change from self.env.company to mrp_area_id.company_id to compute the supply method correctly in the area currently working.


LoisRForgeFlow (approved):
Nice catch

PR #764, opened 2022-02-25
[14.0][FIX] mrp_account_analytic: dr fix mrp Account analytic

dreispt:
- [FIX] mrp_account_analytic: MO Operations error clicking Unblock button


patrickrwilson:
@dreispt looks like test results may have changed causing travis to be red.

dreispt:
Update: it looks like this introduced issues and WIP was no longer generated.
From analysis, loss_id is always set at `mrp/models/mrp_workorder.py +312`

PR #766, opened 2022-03-01
[13.0][FIX] mrp_multi_level: Fix case when Purchase Uom is different from Uom

NuriaXifre:
This fix solves a problem when computing the supply quantities of the MRP Inventory if the following conditions are accomplished:

* The product has a Purchase unit of measure different from the unit of measure.
* A Purchase Order is created but is kept in Draft.

If these 2 conditions were accomplished, the MRP Inventory did not compute correctly the supply_qty, as it had into account the product_qty and not the product_uom_qty, which led into an inconsistency, as the quantities in supply did not match the actual ordered quantities.

@LoisRForgeFlow @ForgeFlow


LoisRForgeFlow (approved):
LGTM, thanks for the test case :+1:

PR #773, opened 2022-03-08
[14.0][ADD] mrp_attachment_mgmt: New addon.

victoralmau:
This module allows to get all attachments of a bill of materials (all levels) on a view.
This module allows to get all attachments of a work orders products to produce on a view.

Please @pedrobaeza and  @chienandalu can you review it?

@Tecnativa TT33396


mrp_attachment_mgmt/README.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Usage
+=====
+
+#. Go to *Manufacturing -> Operation > Work Orders* and click in window open button (after the "Start" and "Block" options).
+#. The smart-button "Attachments" display the attachments of the related product.
+
+
+#. Go to *Manufacturing -> Products > Bill of material -> Desk Combination*.
+#. Go to "Attachment" icon related to "Office Chair Black" component and upload some file.
```
chienandalu, line 51: I get this error:
```
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
Peek 08-03-2022 12-40

chienandalu (changes requested):

mrp_attachment_mgmt/README.rst
victoralmau: Sorry, it is necessary to upload the files from the other icon and use the smart-button only to see all the attachments.
ejemplo

chienandalu (approved):
Perfect! :+1:

pedrobaeza:
We can't it be done from both places? I think that a not good UX experience. Fix it for being done in both places, or remove the smart-button, or avoid the create button there.

victoralmau:
> We can't it be done from both places? I think that a not good UX experience. Fix it for being done in both places, or remove the smart-button, or avoid the create button there.

The button at the component line level sends by context the product of that line.
The `mrp.document `model is not intended to create attachments that are not associated with a product (component).
In my opinion, the top smart-button is useful to see all attachments together from all components (useful if you have for example 20 components with attachments).

pedrobaeza:
And why you can't send the same product in the context on the top smart-button?

victoralmau:
Because the context that is currently sent is that of each of the components. The product from the list of materials could be sent by context in the upper smart-button, but I think it would be a bit strange.

pedrobaeza:
OK, I see. Then can you avoid the attachment creation from that view? There's a way to do it, but I don't know which one right now.

victoralmau:
Currently, from the top smart-button, the "upload" button is displayed but if we enter from the button of each component the "upload" button is displayed and attachments can be created.

pedrobaeza:
Then can you disable the upload button from the top smart-button?

victoralmau:
Yes, it is already disabled.

ejemplo-bom

pedrobaeza (approved):

PR #775, opened 2022-03-09
[14.0][IMP] mrp_attachment_mgmt: Show in the smart-button of the bom the attachments (ir.attachment) uploaded from bom and the attachments of the components.

victoralmau:
Show in the smart-button of the bom the attachments (ir.attachment) uploaded from bom and the attachments of the components.
Improve tests to allow inheritance in other addons that extend it.

Please @pedrobaeza and  @chienandalu can you review it?

@Tecnativa TT33396


chienandalu (commented):
Shouldn't this button be invisible?

image

chienandalu (approved):

pedrobaeza (approved):

PR #788, opened 2022-05-03
[14.0][IMP] mrp_production_serial_matrix: handle flexible manufacturing.

LoisRForgeFlow:
If the component lines are changed they might differ from what it
is stated in the BoM. It is better to rely on compoent lines to
generate Matrix.

New option to decide whether to add or not products tracked by lots
to the matrix is added. Product tracked by serials are always added.

@ForgeFlow


JordiMForgeFlow (approved):
LGTM 👍🏼

MateuGForgeFlow (approved):

PR #789, opened 2022-05-04
[14.0][FIX] mrp_production_grouped_by_product: duplicated finished moves

LoisRForgeFlow:
When the destination MO already has finished stock moves, they
ended up duplicated because they were created again in the
procurement logic (`_run_buy`).

@ForgeFlow


mariadforgeflow (approved):
LGTM!

PR #793, opened 2022-05-06
[14.0][ADD] Subcontractor Partner Management

ivs-cetmix:
Manage subcontracting related entities


elvise (approved):
LGTM!

elvise:
@OCA/manufacturing-maintainers please, take a look here :)

elvise:
@OCA/manufacturing-maintainers is there any soul good that can to check this cool PR ? :)

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -0,0 +1,267 @@
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    is_subcontractor_partner = fields.Boolean(string="Subcontractor")
+    subcontracted_created_location_id = fields.Many2one("stock.location")
+    partner_picking_type_id = fields.Many2one("stock.picking.type")
```
pedrobaeza, line 9: I don't think a new picking type per subcontractor is needed. Too much noise IMO.

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -0,0 +1,267 @@
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    is_subcontractor_partner = fields.Boolean(string="Subcontractor")
```
pedrobaeza, line 7: Instead of this field, I would prefer to simply have a button to populate the subcontractor fields instead.

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -0,0 +1,267 @@
...
+    subcontracted_created_location_id = fields.Many2one("stock.location")
```
pedrobaeza, line 8: Why do you need this extra field instead of simply relying on the standard one that defines the subcontractor location?

mrp_subcontracting_partner_management/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,11 @@
+MRP Subcontracting Partner Management
```
pedrobaeza: Don't add the title here in the fragment

mrp_subcontracting_partner_management/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+14.0.1.0.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pedrobaeza: No need of this file, but if included, use proper underlining of the same length as the text

pedrobaeza (changes requested):
Don't include company logo in the module

mrp_subcontracting_partner_management/models/res_partner.py
elvise: @pedrobaeza That's required from standard flow. 
It's used when you have the semifinished products

pedrobaeza: I mean to use the same picking type for all

mrp_subcontracting_partner_management/models/res_partner.py
elvise: @pedrobaeza what happens when the partner is no longer a subcontractor?
In this case when it's =0 then archive locations, operation type. (and viceversa)

pedrobaeza: That's something that you can do without this check. Just do on write if data is not default, archive it.

ivs-cetmix: This field is also useful for filters or domain rules. 
E.g. I would like some of my user to be able to see only Subcontractors

pedrobaeza: Ok, as you prefer. This is not blocking

mrp_subcontracting_partner_management/models/res_partner.py
ivs-cetmix: Because this field always contains a record and we need to find some way to distinguish the default location from the one we have created ourselves. Here is an example of why it is needed for:
1. Enable "Subcontractor". A new location is created.
2. Now disable "Subcontractor". Location is archived.
3. Now let's re-enable "Subcontractor". At this point we need to unarchive an existing location. Otherwise a new location will be created again. This is what this field is used for.

I understand your concerns regarding "too much fields in partner". But we need to store the information somewhere. Another option is to store in the stock.location We can opt for this if you prefer

mrp_subcontracting_partner_management/readme/DESCRIPTION.rst
elvise: @isserver1 can you fix it ?

mrp_subcontracting_partner_management/readme/HISTORY.rst
elvise: @isserver1  can you fix it ?

mrp_subcontracting_partner_management/models/res_partner.py
elvise: @pedrobaeza what do you think ?

pedrobaeza: You are already adding `is_subcontractor_partner` for distinguishing if the partner has already been setup or not, so no need of these extra fields.

ivs-cetmix: > You are already adding `is_subcontractor_partner` for distinguishing if the partner has already been setup or not, so no need of these extra fields.

It distinguishes only the current status of the partner. If we remove these fields then next time when the subcontractor is re-enabled new records will be created instead of unarchiving the existing ones.

pedrobaeza: I understand what you mean, but being a corner case that you play with activating and deactivating the field, I would go with:

- Using standard fields.
- If you unmark the check, you deactivate the records whatever they have linked.
- If you mark (again or not), check if the linked location is active. If not, you activate if. If active, then you create a new one, as it can be deduced that is the standard one which is linked.

What do you think?

ivs-cetmix: Yes, you are absolutely right. However in some cases we cannot distinguish the location that was archived by our module from the location that was archived by user explicitly.

Example:
- Archive the default "Subcontracting" location. 
- Install this module.
- Enable the "Subcontractor" for  some partner.

Expected behavior:
- A new location is created "Subcontracting/PartnerName"

Actual behavior without special fields:
- The default "Subcontracting" location will be unarchived, no location will be created for the partner

elvise: @pedrobaeza looks reasonable? :)

pedrobaeza: I know, but as said, it's a corner case that is not going to happen a lot, and easily fixable. I prefer to not add too much fields, but I don't want to impose my criteria. Go the way you prefer.

ivs-cetmix: Yes, I totally agree with you. Do not like this fields too) But this approach is actually based on our case.
We have some seasonal and part-time available subcontractors. And we disable/enable them based on their availability. So they do not appear in search when purchase folks decide where to place a new order.

pedrobaeza (approved):

elvise:
@pedrobaeza Can you check the bot ? 
Looks tired today:)

pedrobaeza:
That GitHub action is broken right now. Nothing that I can fix.

PR #798, opened 2022-05-12
[12.0][ADD] Sections and Notes on BoM

quentinDupont:
Inspired by code https://github.com/BADEP/addons/tree/14.0/mrp_bom_section
In a similar way to invoices, this module adds the possibility to add sections ands notes in Bills of Materials. It adds a page in BoM form view and hide the orignal page (you can see it as admin in debug mode)

mrp_bom_widget_add_section_and_note

:warning: Depends on account module since widget 'section_and_note' is defined on section_and_note_fields_backend.js in account module.


robinkeunen (commented):
Looks clean. I don't understand how/where the original `bom_line_ids` are created since you display `bom_line_with_sectionnote_ids` in your new page.

quentinDupont:
> 

While writing this code, I understand that if you have two fields (fieldA & fieldB) on One2many to the same model (lines of ModelA), fieldA and fieldB will share the lines of ModelA.
The trick is to use `display_type` here to handle notes&sections or not.

So `bom_line_ids` are "created" when you save `bom_line_with_sectionnote_ids`


Here is a gif where we see the two pages as admin :

Peek 15-06-2022 10-51
`

mrp_bom_widget_section_and_note_one2many/models/mrp_bom_line.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import fields, models
+
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    # To handle bom_line section & note, we have to make these fiels not required
+    product_id = fields.Many2one(required=False)
+    product_qty = fields.Float(required=False)
```
remi-filament: I guess you should then add a constraint that these fields are at least present when you are not entering either section or note ?

As done here : https://github.com/odoo/odoo/blob/f31fcc7e72e8e1c73b7f9b38fdb41a2df9e529a2/addons/sale/models/sale.py#L1315-L1322

Also you end up with product_qty = 1 since this is the default for this field, so either you should set it to False when writing a new section / note, or you can leave it required.

remi-filament (changes requested):
Thanks @quentinDupont for this PR !
I have one comment on the fact you are removing the required attribute from field which should be replaced by a constraint I guess.
Also, while testing with Runboat, I see that the menu added by mrp_bom_component_menu module will then show section and notes lines without product.

robinkeunen (approved):
Thanks for the explanation. Then it looks good to me.

mrp_bom_widget_section_and_note_one2many/models/mrp_bom_line.py
quentinDupont: Thanks fo your review ! I agree with you !

quentinDupont:
@remi-filament "Also, while testing with Runboat, I see that the menu added by mrp_bom_component_menu module will then show section and notes lines without product." Indeed. Out of scope of this PR no ? But i will write it on roadmap

mrp_bom_widget_section_and_note_one2many/models/mrp_bom_line.py
quentinDupont: I let product_qty required
Remove product_id required on models but add sql_constraints + view also requires it

mrp_bom_widget_section_and_note_one2many/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2022 - Today: GRAP (http://www.grap.coop)
+# @author: BADEP
+# @author: Quentin Dupont (quentin.dupont@grap.coop)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "MRP Widget Section and Note in BoM",
+    "summary": "Add section and note in Bills of Materials",
+    "version": "12.0.1.1.0",
+    "category": "GRAP - Incubator",
```
remi-filament: I suppose this one is internal GRAP category and Manufacturing/Manufacturing should go there instead

remi-filament (approved):
One minor comment on category in manifest, otherwise LGTM !

mrp_bom_widget_section_and_note_one2many/__manifest__.py
quentinDupont: :pray:  :heavy_check_mark:

ivantodorovich (approved):

bosd:
@dreispt Can this one be merged?

legalsylvain:
hi @dreispt : could you try to merge again ? Thanks !

PR #802, opened 2022-05-30
[14.0][ADD] mrp subcontracting partner management mto mts

aayartsev:
The goal of this module is to simplify the management of the partner properties used in MRP Subcontracting.
It adds a new selection "Subcontractor" which have two parts:
* Buy - enable mrp_subcontracting_partner_management module logic
* MTO+MTS - create MTO+MTS rule


elvise:
@aayartsev can you rebase and move in open state ?

PR #807, opened 2022-06-15
[14.0][FIX]mrp_multi_level: avoid mixing str with boolean

LaiaTForgeFlow:
Prevents error during procurement from the MRP

### DIFF
--- a/mrp_multi_level/wizards/mrp_inventory_procure.py
+++ b/mrp_multi_level/wizards/mrp_inventory_procure.py
@@ -83,8 +83,8 @@ class MrpInventoryProcure(models.TransientModel):
                     item.qty,
                     item.uom_id,
                     item.location_id,
-                    "MRP: " + item.planned_order_id.name or str(self.env.user.login),
-                    "MRP: " + item.planned_order_id.origin or str(self.env.user.login),
+                    "MRP: " + (item.planned_order_id.name or self.env.user.login),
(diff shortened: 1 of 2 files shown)


mrp_multi_level/wizards/mrp_inventory_procure.py
```diff
@@ -83,8 +83,8 @@ def make_procurement(self):
                     item.qty,
                     item.uom_id,
                     item.location_id,
-                    "MRP: " + item.planned_order_id.name or str(self.env.user.login),
-                    "MRP: " + item.planned_order_id.origin or str(self.env.user.login),
+                    "MRP: " + str(item.planned_order_id.name) or str(self.env.user.login),
+                    "MRP: " + str(item.planned_order_id.origin) or str(self.env.user.login),
```
MiquelRForgeFlow:
```suggestion
                    "MRP: " + (item.planned_order_id.name or "") str(self.env.user.login),
                    "MRP: " + (item.planned_order_id.origin or "") or str(self.env.user.login),
```
str(None) = 'None'
str(False) = 'False'

I suppose you don't want a string like "MRP: False" :)

MiquelRForgeFlow: Also you need to execute `pre-commit run -a` in your command window

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):

PR #816, opened 2022-06-20
[15.0][MIG] mrp_bom_line_sequence

JuanyDForgeflow:
Migration to 15.0

Provide a new field sequence on bom components, which allows to manage the
order of moves in a picking.


MiquelRForgeFlow:
please rebase

PR #825, opened 2022-07-11
[15.0][ADD] add module mrp_mass_produce_allow_components_serial_number

damdam-s:
Allow to "mass produce" when finished product is tracked by lot and
its components are tracked by unique serial number

### DIFF
--- a/mrp_mass_produce_allow_components_serial_number/__init__.py
+++ b/mrp_mass_produce_allow_components_serial_number/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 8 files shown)


ivantodorovich (commented):
Test coverage would be good, specially with these large overrides

yankinmax:
@damdam-s can you pls rebase?
@simahawk @LoisRForgeFlow can we pls open this PR?

yankinmax (commented):
LG, functionally it works as expected already several months on the project

LoisRForgeFlow:
> Test coverage would be good, specially with these large overrides

@yankinmax I agree with Ivan here, these is missing a small test case.

LoisRForgeFlow (changes requested):
@damdam-s Could you add a small test case?

PR #831, opened 2022-08-04
[14.0][FIX] repair_stock_move: do not change repair line default location. Update moves

JordiBForgeFlow:
* Do not change repair line default location: because this interferes with other modules and
standard behaviour.
* Update moves: when you change the repair lines the corresponding stock moves should be updated
as well.


JordiBForgeFlow:
@DavidBForgeFlow Can you review?

PR #842, opened 2022-09-05
[12.0][FIX] mrp_multi_level: Manage Kits in MRP Multi Level

BernatPForgeFlow:
When running MRP Multi Level, if a product BoM Type is _Kit_, it won't create Planned Orders for the product and it won't neither create an MRP Inventory row. It will only create the MRP Moves to continue the explosion down.

For MRP Parameters that are Kits, they will have _Kit_ supply method.


LoisRForgeFlow (approved):
LGTM :+1:

JoanSForgeFlow (approved):
This is great! Thank you @BernatPForgeFlow

PR #843, opened 2022-09-07
[15.0][ADD] mrp_lot_number_propagation

sebalix:
Allow to propagate a lot number from a component to a finished product.


mrp_lot_number_propagation/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "maintainers": ["sebalix"],
+    "summary": "Propagate a serial number from a component to a finished product",
+    "website": "https://github.com/OCA/manufacture",
+    "category": "Manufacturing",
+    "depends": ["mrp"],
+    "data": [
+        "views/mrp_bom.xml",
+        "views/mrp_production.xml",
+    ],
+    "demo": [],
```
TDu: Could be removed if empty.

mrp_lot_number_propagation/models/mrp_bom.py
```diff
@@ -0,0 +1,70 @@
...
+                and tools.float_compare(
+                    line.product_qty, 1, precision_rounding=line.product_uom_id.rounding
+                )
+                == 0
+                and line.product_uom_id == uom_unit
+            ):
+                return True
+        return False
+
+    def onchange_display_lot_number_propagation(self):
```
TDu, line 73: No `@api.onchange` decorator needed ?

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,76 @@
...
+    propagated_lot_producing = fields.Char(
+        compute="_compute_propagated_lot_producing",
+        help=(
+            "The BoM used on this manufacturing order is set to propagate "
+            "lot number from one of its components. The value will be "
+            "computed once the corresponding component is selected."
+        ),
+    )
+
+    def _compute_propagated_lot_producing(self):
```
TDu, line 42: Missing an `api.depends` ?

mrp_lot_number_propagation/models/mrp_bom_line.py
```diff
@@ -0,0 +1,75 @@
...
+    _inherit = "mrp.bom.line"
+
+    propagate_lot_number = fields.Boolean(
+        default=False,
+    )
+    display_propagate_lot_number = fields.Boolean(
+        compute="_compute_display_propagate_lot_number"
+    )
+
+    @api.depends()
```
TDu: Why empty ?

sebalix: Fixed

mrp_lot_number_propagation/views/mrp_bom.xml
```diff
@@ -0,0 +1,28 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+  <record id="mrp_bom_form_view" model="ir.ui.view">
+    <field name="name">mrp.bom.form.inherit</field>
+    <field name="model">mrp.bom</field>
+    <field name="inherit_id" ref="mrp.mrp_bom_form_view" />
+    <field name="arch" type="xml">
+      <field name="company_id" position="after">
+        <field name="display_lot_number_propagation" invisible="0" />
```
grindtildeath:
```suggestion
        <field name="display_lot_number_propagation" invisible="1" />
```

mrp_lot_number_propagation/models/mrp_bom_line.py
```diff
@@ -0,0 +1,78 @@
...
+                if len(lines_to_propagate) > 1:
+                    raise ValidationError(
+                        _(
+                            "Only one BoM line can propagate its lot/serial number "
+                            "to the finished product."
+                        )
+                    )
+                if line.propagate_lot_number and line.product_id.tracking != "serial":
+                    raise ValidationError(
+                        _("%s is not tracked serial number.")
```
grindtildeath:
```suggestion
                        _("Only components tracked by serial number can propagate its lot/serial number to the finished product.")
```

mrp_lot_number_propagation/models/mrp_bom_line.py
```diff
@@ -0,0 +1,78 @@
...
+                    )
+                if (
+                    line.propagate_lot_number
+                    and line.product_id.tracking != line.bom_id.product_tmpl_id.tracking
+                    and (
+                        line.bom_id.product_tmpl_id.tracking == "serial"
+                        or line.product_id.tracking == "serial"
+                    )
+                ):
+                    raise ValidationError(
```
grindtildeath: Could we ever have this error instead of the one above?

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,81 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    is_lot_number_propagated = fields.Boolean(
+        related="bom_id.lot_number_propagation",
```
grindtildeath: After testing, I feel like this field should not be related to the BOM, but set through `action_confirm`. So that the MOs that were confirmed won't be affected by changes to the BOM (as is the case with components).

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,81 @@
...
+        compute="_compute_propagated_lot_producing",
+        help=(
+            "The BoM used on this manufacturing order is set to propagate "
+            "lot number from one of its components. The value will be "
+            "computed once the corresponding component is selected."
+        ),
+    )
+
+    @api.depends(
+        "move_raw_ids.bom_line_id.propagate_lot_number",
```
grindtildeath: Same as for the comment above, I feel like this should not be related to the BOM. Can we instead add a readonly `propagate_lot_number` Boolean field on `stock.move` and set it through `action_confirm`?

mrp_lot_number_propagation/views/mrp_production.xml
```diff
@@ -0,0 +1,22 @@
...
+<!-- Copyright 2022 Camptocamp SA
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+  <record id="mrp_production_form_view" model="ir.ui.view">
+    <field name="name">mrp.production.form.inherit</field>
+    <field name="model">mrp.production</field>
+    <field name="inherit_id" ref="mrp.mrp_production_form_view" />
+    <field name="arch" type="xml">
+      <!-- Place new fields in the first group while being compatible with OE  -->
```
grindtildeath, line 11: Can you please also set the `lot_producing_id` and the button next to it as invisible if `is_lot_number_propagated = True` and `state 'not in' (done,to_close)`?

mrp_lot_number_propagation/views/mrp_production.xml
```diff
@@ -0,0 +1,22 @@
...
+      <xpath expr="//field[@name='id']/.." position="inside">
+        <field name="is_lot_number_propagated" />
+        <field
+                    name="propagated_lot_producing"
```
grindtildeath, line 17:
```suggestion
                    name="propagated_lot_producing" string="Lot/Serial Number"
```
So that it kinds of replaces `lot_producing_id` while the MO is open

grindtildeath (commented):
Thanks, looks quite good overall :+1: 

There's one issue I noticed when testing on runbot:
After creating a MO for 3 units, and registering the components consumption of the first unit, clicking on Validate will open the "create backorder" wizard. Then, when clicking on create backorder I get following error:
> The operation cannot be completed:
> - Create/update: a mandatory field is not set.
> - Delete: another model requires the record being deleted. If possible, archive it instead.
> 
> Model: Lot/Serial (stock.production.lot), Field: Lot/Serial Number (name)

mrp_lot_number_propagation/views/mrp_production.xml
sebalix: There is already some `attrs` on these elements and didn't want to break the view by copying it, that's why I put the check in the `write` method to be sure this field won't be updated by mistake.
But if you think we can replace the whole `attrs` value I can do the change.

mrp_lot_number_propagation/views/mrp_production.xml
sebalix: Make sense :thinking:

mrp_lot_number_propagation/models/mrp_bom_line.py
sebalix: Indeed, could be removed

mrp_lot_number_propagation/views/mrp_production.xml
sebalix: Done through `fields_view_get` to not break the inheritance.

mrp_lot_number_propagation/models/mrp_production.py
sebalix: Stored on `action_confirm` + added an `onchange` on `bom_id` to have the value even if the MO is not yet validated

mrp_lot_number_propagation/views/mrp_bom.xml
sebalix: fixed

sebalix:
Still have to check your last comment + other fixes to apply

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,142 @@
...
+    @api.depends(
+        "move_raw_ids.propagate_lot_number",
+        "move_raw_ids.move_line_ids.qty_done",
+        "move_raw_ids.move_line_ids.lot_id",
+    )
+    def _compute_propagated_lot_producing(self):
+        for order in self:
+            order.propagated_lot_producing = False
+            move_with_lot = order.move_raw_ids.filtered(
+                lambda o: o.bom_line_id.propagate_lot_number
```
grindtildeath:
```suggestion
                lambda o: o.propagate_lot_number
```

sebalix:
@grindtildeath the issue you encountered is fixed. Also, we should display a warning/info message to invite the user to consume 1 qty of the reserved SN because the finished product is tracked by SN. Right now he have to guess this by himself.

mrp_lot_number_propagation/models/mrp_bom_line.py
```diff
@@ -0,0 +1,62 @@
...
+
+        - if the bom has lot_number_propagation marked, there is one and
+          only one line of this bom with propagate_lot_number marked.
+        - the bom line being marked with lot_number_propagation is of the same
+          tracking type as the finished product
+        """
+        for line in self:
+            lines_to_propagate = line.bom_id.bom_line_ids.filtered(
+                lambda o: o.propagate_lot_number
+            )
```
i-vyshnevska, line 44: this makes me think that constrain should work mrp.bom level not at line level, the order is responsible for lines, and changes in one line should not trigger changes in another line (they may but that brings complexity)

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,148 @@
...
+    def write(self, vals):
+        for order in self:
+            if (
+                order.is_lot_number_propagated
+                and "lot_producing_id" in vals
+                and not self.env.context.get("lot_propagation")
+            ):
+                raise UserError(
+                    _(
+                        "Lot/Serial number is propagated from a component, "
```
i-vyshnevska, line 108:
```suggestion
                        "Serial number is propagated from a component, "
```

i-vyshnevska (approved):

mrp_lot_number_propagation/models/mrp_bom.py
```diff
@@ -0,0 +1,71 @@
...
+        - the finished product is tracked by serial number
+        - the quantity of the finished product is 1 and its UoM is unit
+        - there is at least one bom line, with a component tracked by serial,
+          having a quantity of 1 and its UoM is unit
+        """
+        uom_unit = self.env.ref("uom.product_uom_unit")
+        for bom in self:
+            bom.display_lot_number_propagation = (
+                bom.type == "normal"
+                and bom.product_tmpl_id.tracking == "serial"
```
i-vyshnevska, line 45: we narrow here only for "serial" tracking would be good to rid of the "lot" notion in naming as implementation only about the propagation of a serial number

mrp_lot_number_propagation/models/mrp_bom_line.py
sebalix: Yes, I should move this to BoM level indeed

mrp_lot_number_propagation/models/mrp_bom.py
sebalix: I can't say about that, maybe the support for lot numbers could be added later? cc @grindtildeath

mrp_lot_number_propagation/models/mrp_bom_line.py
sebalix: I couldn't make it working by moving it on BoM, and I prefer to not spend too much time on this ATM so I'll let this part like this.

mrp_lot_number_propagation/models/mrp_bom.py
grindtildeath: That's the idea. cf ROADMAP.rst

mrp_lot_number_propagation/models/mrp_bom_line.py
```diff
@@ -0,0 +1,62 @@
...
+            line.display_propagate_lot_number = (
+                line.bom_id.display_lot_number_propagation
+                and line.bom_id.lot_number_propagation
+            )
+
+    @api.constrains(
+        "propagate_lot_number",
+        "bom_id.lot_number_propagation",
+        "product_id.tracking",
+        "bom_id.product_tmpl_id.tracking",
```
grindtildeath: This raises following warning:
```
2022-10-10 15:38:58,454 1 WARNING demo_mrp_propagation odoo.models: method mrp.bom.line._check_propagate_lot_number: @constrains parameter 'bom_id.lot_number_propagation' is not a field name 
2022-10-10 15:38:58,454 1 WARNING demo_mrp_propagation odoo.models: method mrp.bom.line._check_propagate_lot_number: @constrains parameter 'product_id.tracking' is not a field name 
2022-10-10 15:38:58,454 1 WARNING demo_mrp_propagation odoo.models: method mrp.bom.line._check_propagate_lot_number: @constrains parameter 'bom_id.product_tmpl_id.tracking' is not a field name 
```
I guess we must define constraints functions on both `mrp.bom` and `product.product` (or `product.template`?) that will call this function on the BOM lines otherwise we can end up with this situation:
image

mrp_lot_number_propagation/models/mrp_production.py
sebalix: I named it this way to stick with the std lot/sn form: 
image
And if later we add the support for production lot we won't need to change the string+translations.

ivantodorovich (approved):

mrp_lot_number_propagation/models/mrp_bom_line.py
sebalix: I added some constraints on `product.product`/`product.template`/`mrp.bom` to avoid this situation.
If we tick "Lot Number Propagation" on a BoM, it's mandatory to tick "Propagate Lot Number" on a BoM line to not let the BoM in a wobbly state. Also, if a product is related to a BoM or BoM line, its tracking can't be unset (the user will have to update/delete/archive related BoM before).

mrp_lot_number_propagation/models/mrp_bom_line.py
sebalix: With the last changes, we can't anymore half configure the BoM, constraints have been added also at BoM level (+ products level as well to avoid updates on `tracking` field).

PR #847, opened 2022-09-08
[REF] Run pre-commit autoupdate

dreispt:
- [REF] Run pre-commit autoupdate
- [REF] quality_control_mrp_oca: fix pre-commit


dreispt:
cc @OCA/manufacturing-maintainers 
Trigger for this was to avoid pre-commit dependency errors when using Python 3.10

PR #861, opened 2022-09-28
[12.0][IMP] mrp_sale_info: Smartlink on sale order to mrp.production related records. A direct copy from sale_mrp_link

ntsirintanis:


mrp_sale_info/models/sale_order.py
```diff
@@ -0,0 +1,28 @@
...
+    @api.multi
+    def action_view_production(self):
+        action = self.env.ref('mrp.mrp_production_action').read()[0]
+        if self.production_count > 1:
+            action['domain'] = [('id', 'in', self.production_ids.ids)]
+        else:
+            action['views'] = [
+                (self.env.ref('mrp.mrp_production_form_view').id, 'form')]
+            action['res_id'] = self.production_ids.id
+        return action
```
thomaspaulb, line 29: For the sake of coverage, could you add a simple assert of `sale.production_count == 1` and a call to [action_view_production](https://github.com/OCA/manufacture/blob/12.0/mrp_sale_info/tests/test_mrp_sale_info.py#L39) in the tests here, asserting that the response in `res_id` is the production order we are expecting? It won't cover the `> 1` case, but it should already be a better coverage.

thomaspaulb:
@ntsirintanis Could you change tile to `[12.0][IMP] mrp_sale_info: ....` format

thomaspaulb (approved):

NL66278 (changes requested):
Looks mostly good to me,  but black the added `mrp_sale_info/models/sale_order.py` to minimize the diff when forward porting. And because blacked code is nicer anyway.

thomaspaulb:
@NL66278 Makes sense, blacked it

NL66278 (approved):
:+1: LGTM

gfcapalbo:
:+1: LGTM

thomaspaulb:
@pedrobaeza This is the 12.0 version of https://github.com/OCA/manufacture/pull/943, can it be merged too?

PR #865, opened 2022-10-01
[14.0][IMP] mrp_workcenter_hierarchical: Parents level identification

mymage:
Hi all,
I've a suggestion to the mrp_workcenter_hierarchical module.
Now the parents level 1, 2, 3 are filled with same value if there isn't other dependencies and uses its own value as parent.
before

With this change references are set to false if they already exist in the hierarchy, as well as by itself.
after

It remains a "problem" if the stratification is over 4 levels.
I don't know if is opportune to add an indicator of "there are other levels..."

Surely my code isn't elegant but I'm not a Python Cheff so thanks for review.

### DIFF
--- a/mrp_workcenter_hierarchical/models/mrp_workcenter.py
+++ b/mrp_workcenter_hierarchical/models/mrp_workcenter.py
@@ -45,7 +45,7 @@ class MrpWorkcenter(models.Model):
             ids = []
         return ids
 
-    @api.depends("parent_id.parent_id.parent_id", "child_ids")
+    @api.depends("parent_id.parent_id.parent_id.parent_id", "child_ids")
     def _compute_parent_level(self):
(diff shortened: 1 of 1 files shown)


mymage:
@andreampiovesana @marcelofrare @primes2h @TheMule71

mymage:
Ping @andreampiovesana @marcelofrare @primes2h @TheMule71

andreampiovesana (approved):
LGTM

marcelofrare (approved):
functional improvement is correct for me.

marcelofrare:
Also I think it would be better to match the sort order with the hierarchy. Example

WC = 1
PARENT = 12
PARENT LEVEL 1 = 123
PARENT LEVEL 2 = 1234
PARENT LEVEL 3 = 12345

current situation
immagine

desirable situation
immagine

PR #869, opened 2022-10-04
[14.0] Add module mrp_lot_on_hand_first

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/stock-logistics-workflow/pull/1097
 - [x] https://github.com/OCA/stock-logistics-workflow/pull/1096

PR #870, opened 2022-10-04
[15.0][MIG] mrp_lot_on_hand_first

grindtildeath:
Migrated from:
 - [ ] #869 

Depends on:
 - [x] https://github.com/OCA/stock-logistics-workflow/pull/1098
 - [x] https://github.com/OCA/stock-logistics-workflow/pull/1099

### DIFF
--- a/mrp_lot_on_hand_first/views/mrp_production.xml
+++ b/mrp_lot_on_hand_first/views/mrp_production.xml
@@ -12,10 +12,11 @@
                 expr="//field[@name='move_raw_ids']//field[@name='lot_ids']"
                 position="attributes"
             >
-                <attribute name="context">{
-                    'name_search_qty_on_hand_first': parent.display_lots_on_hand_first,
-                    'default_company_id': company_id,
-                    'default_product_id': product_id,}</attribute>
+            <attribute
+                    name="context"
+                    operation="python_dict"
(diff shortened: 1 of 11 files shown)


ajaniszewska-dev (approved):

ivantodorovich (approved):

mrp_lot_on_hand_first/views/mrp_production.xml
```diff
@@ -0,0 +1,23 @@
...
+            </field>
+            <xpath
+                expr="//field[@name='move_raw_ids']//field[@name='lot_ids']"
+                position="attributes"
+            >
+            <attribute
+                    name="context"
+                    operation="python_dict"
+                    key="name_search_qty_on_hand_first"
+                >parent.display_lots_on_hand_first</attribute>
```
ajaniszewska-dev: just being curious, i see also lot of times construction with update - like below, why `python_dict` then?
```
<attribute name="context" operation="update">
          {"some_key": some_value}
</attribute>
```

grindtildeath: It used to be so in version 14.0, but you're right that it has to use update in 15.0

ivantodorovich:
Build is red 😢

PR #871, opened 2022-10-05
[14.0][IMP] mrp_subcontracting_partner_management

geomer198:
Functionality added by this module:
* When name of subcontractor is updated, names of entities are updated automatically.
* It is also possible to check inventory locations using **Subcontractor Location Stock** smart button on partner.
When name of subcontractor is updated, names of entities (subcontracting location, operation type, rules) are updated automatically.


mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -41,6 +41,11 @@ def unlink(self):
         return result
 
     def write(self, values):
+        name = values.get("name")
+        if name:
+            self.filtered(lambda p: p.is_subcontractor_partner).mapped(
+                "property_stock_subcontractor"
+            ).write({"name": name})
         for record in self:
```
ivs-cetmix: This should be refactored as well. Use one write for both cases

ivs-cetmix (changes requested):
We should also rename all other entities invoked. Like routes, rules etc

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
-        if self.subcontracted_created_location_id:
-            self.subcontracted_created_location_id.active = active
-        if self.partner_picking_type_id:
-            self.partner_picking_type_id.active = active
-        if self.partner_buy_rule_id:
-            self.partner_buy_rule_id.active = active
-        if self.partner_resupply_rule_id:
-            self.partner_resupply_rule_id.active = active
+        for rec in self:
+            for record in filter(bool, map(lambda f: getattr(rec, f), self.CHECK_DATA)):
```
GabbasovDinar: please, check this way
```python
# get field names
field_names = self.CHECK_DATA.keys()

for record in self:
    result = record.read([field_names])[0]
    for field in field_names:
        if result.get(field):
             setattr(record, field, active)
```

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
-        for record in self:
-            is_subcontractor_partner = values.get("is_subcontractor_partner")
-            active = values.get("active")
-            if is_subcontractor_partner is not None:
-                values.update(record._update_subcontractor_entities_for_record(values))
-            if active is not None:
-                record._set_subcontracting_values_active(active)
-        super(ResPartner, self).write(values)
+        is_subcontractor_partner = values.get("is_subcontractor_partner")
+        if is_subcontractor_partner is not None:
```
GabbasovDinar:
```python
if "is_subcontractor_partner" in values:
    self._update_subcontractor_entities_for_record(values["is_subcontractor_partner"])
...
```

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
     def write(self, values):
-        for record in self:
-            is_subcontractor_partner = values.get("is_subcontractor_partner")
-            active = values.get("active")
-            if is_subcontractor_partner is not None:
-                values.update(record._update_subcontractor_entities_for_record(values))
-            if active is not None:
-                record._set_subcontracting_values_active(active)
-        super(ResPartner, self).write(values)
+        is_subcontractor_partner = values.get("is_subcontractor_partner")
```
GabbasovDinar: why separate variable here?

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
-            "company_id": company.id,
-            "active": True,
-        }
         location_rec = self.subcontracted_created_location_id
-        if not location_rec:
-            location_rec = self.env["stock.location"].create(location_vals)
-        return location_rec
+        if location_rec:
+            return location_rec
+        return self.env["stock.location"].create(
```
GabbasovDinar: we can just return exist or create new record
```
return self.subcontracted_created_location_id or self.env["stock.location"].create ....
```

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
+        )
 
     def _create_subcontracted_operation_type(self, warehouse, location):
         """Creating Operation Type for Subcontracting"""
-        name = self._compose_entity_name()
-        operation_type_name = "{}: {}".format(name, " IN")
-        sequence_code = ""
-        for code in list(filter(None, operation_type_name.split(" "))):
-            sequence_code += code[0]
         operation_type_rec = self.partner_picking_type_id
```
GabbasovDinar: why separate variable here?

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,125 +6,125 @@ class ResPartner(models.Model):
...
+        rule = self.env["stock.rule"].create(
+            {
+                "name": self._compose_entity_name(),
+                "action": "buy",
+                "picking_type_id": operation_type_rec.id,
+                "location_id": location.id,
+                "route_id": buy_route.id,
+            }
+        )
+        self.partner_buy_rule_id = rule
```
GabbasovDinar, line 120: метод который предпологает создать что то или вернуть существующую записи который еще делает write не очень очевидно по названию метода (

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -157,74 +155,57 @@ def _create_subcontractor_entities(self):
...
-
-        return values
+            rec.write(rec._create_subcontractor_entities_for_record())
+
+    def _update_subcontractor_entities_for_record(self, is_subcontractor_partner):
+        data = self.CHECK_DATA
+        if is_subcontractor_partner:
+            for rec in self:
+                vals = {}
+                records = map(lambda f: (getattr(rec, f[0]), f[1]), data.items())
```
GabbasovDinar: we need move the data.items() before sycle

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -157,74 +155,57 @@ def _create_subcontractor_entities(self):
...
-        company = self.company_id or default_company
-        warehouse = (
-            self.env["stock.warehouse"].search([("company_id", "=", company.id)])[0]
-            if self.company_id
-            else default_warehouse
-        )  # noqa
-        return warehouse
+        StockWarehouse = self.env["stock.warehouse"]
+        company = self.company_id or self.env.company
+        return StockWarehouse.search([("company_id", "=", company.id)], limit=1)
```
GabbasovDinar: here only search and nothing more, why need we to use the separate method?

GabbasovDinar (changes requested):

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Wrong solution(

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: I removed this variable

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: I removed this variable

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved

mrp_subcontracting_partner_management/readme/CONFIGURE.rst
```diff
@@ -1 +1 @@
-* No configuration is required
+To see newly created rules, go to **Settings** > **Inventory** and activate "Multi step routes".
```
francesco-ooops, line 1: It is also possible to allow a user to check inventory locations in Sales & Purchases tab of Vendor without activating dev mode. To do so, enable technical extra right "Display Inventory section on Vendor" in user.

geomer198: Added

francesco-ooops:
@geomer198 rebase please

@isserver1 @GabbasovDinar can you review? thanks!

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -318,11 +318,31 @@ def _mrp_cleanup(self, mrp_areas):
...
     @api.model
     def _low_level_code_calculation(self):
         logger.info("Start low level code calculation")
         counter = 999999
         llc = 0
+        llc_recursion_limit = (
+            int(
+                self.env["ir.config_parameter"]
+                .sudo()
+                .get_param("mrp_multi_level.llc_calculation_recursion_limit")
```
GabbasovDinar: we can use default value for get_param method:

get_param("mrp_multi_level.llc_calculation_recursion_limit", 1000)

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,258 +8,219 @@ class ResPartner(models.Model):
...
-    def _create_subcontracted_resupply_rule(self, location):
-        """# Creating Route Rule for Subcontracting resupply on order starts here"""
-        resupply_on_order_route = self.env.ref(
-            "mrp_subcontracting.route_resupply_subcontractor_mto",
-            raise_if_not_found=False,
+        location_id = self._get_location_id_for_record(vals)
+        if "partner_picking_type_id" in vals:
+            return vals.get("partner_picking_type_id"), location_id
+        if self.partner_picking_type_id:
+            return self.partner_picking_type_id.id, location_id
```
GabbasovDinar, line 143: by this method name: the method should create a new record but you also return some values (not related with creation), maybe we can split this method? (the original method has the correct name and logic of the method, after changes the consistency of the name-logic is broken)

GabbasovDinar (changes requested):

ivs-cetmix (approved):

francesco-ooops (approved):
functional review ok 

@geomer198 can you squash commits and update [icon](https://github.com/OCA/manufacture/blob/14.0/mrp_subcontracting_partner_management/static/description/icon.png) with the OCA one?

---

IMP as per configure/usage files: 

1) It is also possible to allow a user to check inventory locations in **Sales & Purchases** tab of **Vendor** without activating dev mode.
To do so, enable technical extra right **Display Inventory section on Vendor** in user.

2) When name of subcontractor is updated, names of entities (subcontracting location, operation type, rules) are updated automatically.

GabbasovDinar (approved):
LGTM!

francesco-ooops:
@pedrobaeza @OCA/manufacturing-maintainers could you please review? thanks!

pedrobaeza:
Can you summarize the changes/improvements of this PR?

francesco-ooops:
> IMP as per configure/usage files:
> 
> 1. It is also possible to allow a user to check inventory locations in **Sales & Purchases** tab of **Vendor** without activating dev mode.
>    To do so, enable technical extra right **Display Inventory section on Vendor** in user.
> 2. When name of subcontractor is updated, names of entities (subcontracting location, operation type, rules) are updated automatically.

Yes sorry, I did it here

francesco-ooops:
> > IMP as per configure/usage files:
> > 
> > 1. It is also possible to allow a user to check inventory locations in **Sales & Purchases** tab of **Vendor** without activating dev mode.
> >    To do so, enable technical extra right **Display Inventory section on Vendor** in user.
> > 2. When name of subcontractor is updated, names of entities (subcontracting location, operation type, rules) are updated automatically.
> 
> Yes sorry, I did it here

@pedrobaeza

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,258 +8,219 @@ class ResPartner(models.Model):
...
-    )
-    partner_resupply_rule_id = fields.Many2one(
-        copy=False,
-        comodel_name="stock.rule",
+        comodel_name="stock.picking.type", copy=False
     )
+    partner_buy_rule_id = fields.Many2one(comodel_name="stock.rule", copy=False)
+    partner_resupply_rule_id = fields.Many2one(comodel_name="stock.rule", copy=False)
+
+    CHECK_DATA = {
```
pedrobaeza: A good practice is that this is returned by a hook that can be inherited.

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,258 +8,219 @@ class ResPartner(models.Model):
...
-            self.partner_resupply_rule_id.active = active
+        """Set subcontracting values active/inactive by argument key"""
+        for key in self.CHECK_DATA:
+            self.mapped(key).write({"active": active})
 
-    def unlink(self):
+    @api.model
+    def _update_name_translation(self, records, name):
+        """Update name field translation for records"""
+        self.env["ir.translation"].search(
```
pedrobaeza, line 50: Better to simply remove translations, isn't it?

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,258 +8,219 @@ class ResPartner(models.Model):
...
+            self._update_subcontractor_values_name(vals.get("name"))
+        return result
 
     @api.model
-    def create(self, values):
-        partner = super(ResPartner, self).create(values)
-        if values.get("is_subcontractor_partner", False):
-            partner._create_subcontractor_entities()
-        return partner
+    def create(self, vals):
```
pedrobaeza: Do it create_multi for not losing performance in this critical object.

mrp_subcontracting_partner_management/security/res_partner_security.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_display_inventory" model="res.groups">
+        <field name="name">Display Inventory section on Vendor</field>
```
pedrobaeza: This extra group is overkilling for such tiny feature. There are too much technical groups. You can use instead `base_technical_features` group to have technical fields without developer mode.

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: @pedrobaeza What do you mean?

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: @pedrobaeza  When copying a vendor record or changing the vendor name then the translated name of related records doesn't change. That is why I added a translation method .

pedrobaeza: AFAIK, standard copy method handles translations.

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved. Wrapped CHECK_DATA in function.

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: Resolved. Added model_create_multi decorator and optimize a functions code.

mrp_subcontracting_partner_management/security/res_partner_security.xml
francesco-ooops: Hi @pedrobaeza , our goal is to display a read-only "Subcontractor Location" field to inventory user, in order to allow user to check the subcontracting location is correct (but not change it)

A possible solution is to add a read-only field with the subcontractor location name set in the field, in order to achieve the goal above and to maintain the possibility to change the value for user with dev mode access

would this solution be ok?

pedrobaeza: It's a bit of nonsense if this module is automating all the handling of these locations, to have such field to be manually reviewed.

francesco-ooops: @pedrobaeza I checked again - the goal is to allow user to easily check stock in subcontractor's location, so we added a smart button in partner (same smart button as in location). 

Would that work?

pedrobaeza: Yeah, that can be an interesting added value, but without extra security group.

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: @pedrobaeza If the name of a subcontractor changes, translations for the related operations types are not updated.
This function fix this behaviour for related operations types.

francesco-ooops:
@pedrobaeza could you take a look at latest changes?

mrp_subcontracting_partner_management/models/res_partner.py
pedrobaeza: Please use correct nomenclature. The argument should be `vals_list`, and can't this be done inline for not having to create + write?

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,258 +8,234 @@ class ResPartner(models.Model):
...
+        check_data = self.get_data_struct().items()
+        for record in records.filtered(
+            lambda r: r.is_subcontractor_partner and r.is_company
+        ):
+            values = {}
+            for key, func in check_data:
+                if not getattr(record, key) or not values.get(key):
+                    values.update(**getattr(record, func)(values) or {})
+            if values:
+                record.write(values)
```
pedrobaeza: This can be done before calling super iterating on each dictionary and seeing if `is_subcontractor_partner` and `is_company` vals are set.

mrp_subcontracting_partner_management/models/res_partner.py
geomer198: @pedrobaeza After creation record created related records that is why used create + write.

pedrobaeza: But that's the question: Why can't you create them before calling super, and pre-assign them in the dictionary values?

geomer198: @pedrobaeza Resolved.

francesco-ooops:
@pedrobaeza ready to be approved? thanks a lot

pedrobaeza (approved):

PR #876, opened 2022-10-17
[15.0][IMP] mrp_multi_level: avoid recursion on LLC calculation.

LoisRForgeFlow:
Introduce a safe and configurable LLC depth limit to avoid infinite recursion.

Also, improve extensibility of BoM finding and fix it to not consider archived BoMs.

@ForgeFlow


AaronHForgeFlow (approved):
Code review LG!

PR #878, opened 2022-10-19
[15.0][ADD] mrp_subcontracting_no_negative

sebalix:
Same feature than [`stock_no_negative`](https://github.com/OCA/stock-logistics-workflow/tree/15.0/stock_no_negative) but dedicated to subcontracting. It doesn't allow the record of components on a subcontractor receipt if they can't be reserved, so we can't get negative stock levels in subcontractor locations.


ivantodorovich (approved):

i-vyshnevska (approved):

grindtildeath:
@OCA/manufacturing-maintainers Can we please keep this one open as well?

AungKoKoLin1997:
I tried to use this module.Then, I noticed this module's function is covered by `stock_no_negative`, isn't it? I knew there is a issue when validating subcontracting receipts in `stock_no_negative`. If we use these two modules together, the issue is still exit. If we fix the issue in stock_no_negative for subcontracting receipts, we don't need to use mrp_subcontracting_no_negative and cover this module's function.
May be I am missing about the background reason of adding this new module. 
I am planning to fix the issue in stock_no_negative. Please let me know if there is the anything I missed.

sebalix:
@AungKoKoLin1997 Can we fix the subcontracting stock level issue in `stock_no_negative` without adding a dependency on `mrp_subcontracting`? (we want to avoid that)
Also, `stock_negative_qty` constraint is triggered when the `quantity` field of the quant is updated, in a context of subcontracting the error message arrives too late to the user I guess (I don't remember exactly, but it's probably why `stock_no_negative` is not yet triggered when recording the components on the stock transfer).

Another use case is we want to allow negative stock here and there, but disallow it only for subcontractor locations (only `mrp_subcontracting_no_negative` installed).

Of course if you find a way to make `stock_no_negative` working when we are recording components on a subcontractor stock transfer, it is very welcome so we won't need to maintain this one anymore (if we forget the other use case I mentioned, I saw we can configure locations in this regard).

pedrobaeza:
The root problem is described in https://github.com/odoo/odoo/pull/75065

AungKoKoLin1997:
@sebalix @pedrobaeza 

> Can we fix the subcontracting stock level issue in stock_no_negative without adding a dependency on mrp_subcontracting?

Yes. I know we can’t fix without adding a dependency on `mrp_subcontracting`. What I mentioned was we need to fix `stock_no_negative` for subcontracting receipts. So, my thought was instead of adding this feature in this module, we can create new module that depends on `stock_no_negative` to fix subcontracting stock level issue. If we create that new module, the function of `mrp_subcontracting_no_negative` will be covered automatically.

> I don't remember exactly, but it's probably why stock_no_negative is not yet triggered when recording the components on the stock transfer

If I am not wrong, `stock_no_negative` is not triggered when recording the components on the stock transfer because it will not create any negative stock.quant during record components. It will create negative stock.quant when we validate subcontracting receipts. So, it will be shown validation error message for component products when we click validate button.

So, I was thinking we should skip no negative check for finished product in subcontracting receipts. But the workflow will be the same for component products.
Here is my outcome:

- [[16.0][IMP] stock_no_negative: add option to skip negative qty check #1284](https://github.com/OCA/stock-logistics-workflow/pull/1284)
- [[16.0][ADD] mrp_subcontracting_skip_no_negative #1025](https://github.com/OCA/manufacture/pull/1025)

PR #882, opened 2022-10-25
[14.0][BACKPORT] mrp_multi_level: avoid recursion on LLC calculation.

BernatPForgeFlow:
Backport from: https://github.com/OCA/manufacture/pull/876

Introduce a safe and configurable LLC depth limit to avoid infinite recursion.

Also, improve extensibility of BoM finding and fix it to not consider archived BoMs.


LoisRForgeFlow (approved):
Thanks

PR #902, opened 2022-11-22
[15.0][FIX] mrp_multi_level: demo data was failing on installation.

LoisRForgeFlow:
`product_uom_id` is not in the allowed fields for stock.quant when in inventory mode. See `_get_inventory_fields_create` method in stock.quant for more details.

This was making some test fail.


LoisRForgeFlow:
Fast tracking to fix CI.

/ocabot merge patch

PR #913, opened 2022-12-12
[15.0] [ADD] `mrp_production_component_availability_search`

ivantodorovich:
Allows to search or filter manufacturing orders by their Components Availability State.
To do so, a search method is implemented on that computed field, to compute and filter records on the fly.
For convenience, a filter named **Components Availability Issues** is added to the search view.

Somewhat related to:
- https://github.com/OCA/stock-logistics-workflow/pull/1153


jcoux (approved):
Technically seems good :+1:

yankinmax (approved):

PR #914, opened 2022-12-13
[15.0][ADD] mrp_production_date_planned_finished

ivantodorovich:
Shows the Scheduled End Date on Manufacturing Orders, and allows to change the start date from it, through an onchange.

It's useful for companies wanting to plan their manufacturing orders from the target finish date. Usually these companies have long manufacturing times (days or weeks).


mrp_production_date_planned_finished/models/mrp_production.py
```diff
@@ -0,0 +1,26 @@
...
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    @api.onchange("date_planned_finished")
+    def _onchange_date_planned_finished_set_date_planned_start(self):
```
yankinmax, line 14: But, why not
```
@api.depends("date_planned_finished", "is_planned"):
def _compute_date_planned_start(self):
```

yankinmax: @ivantodorovich I remember you showed me examples with: `_compute_`, `store=True`, `readonly=False`.
Can we apply here the same?

ivantodorovich: It would be possible indeed, but I wanted to be consistent with core `onchange` method for `date_planned_start`.

yankinmax: ok, NP for me, thanks for the work

yankinmax (approved):

jcoux (approved):

LoisRForgeFlow (approved):
Functional test

PR #916, opened 2022-12-15
[14.0][ADD] mrp_unbuild_subcontracting: unbuild created when is returned a product subcontracted

ThiagoMForgeFlow:
This module automatically creates a unbuild in draft state when a subcontracting picking return is created. In addition, when the picking is validated, the unbuild is also validated.

@ForgeFlow


mrp_unbuild_subcontracting/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Unbuild orders with return subcontracting",
+    "version": "14.0.1.0.1",
```
MiquelRForgeFlow:
```suggestion
    "version": "14.0.1.0.0",
```

JuanyDForgeflow (commented):
LGTM

mrp_unbuild_subcontracting/__manifest__.py
ThiagoMForgeFlow: Done

JordiBForgeFlow (approved):

mrp_unbuild_subcontracting/views/mrp_unbuild_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        </field>
+    </record>
+
+    <menuitem
+        id="menu_mrp_unbuild_subcontracted"
+        name="Subcontracted Unbuild Orders"
+        parent="mrp.menu_mrp_manufacturing"
+        action="mrp_unbuild_subcontracted"
+        sequence="21"
+        groups="base.group_no_one"
```
BernatPForgeFlow, line 28: Why is this menu only available in 'Developer' mode?

BernatPForgeFlow:
Functional review OK!

mrp_unbuild_subcontracting/models/stock_picking.py
```diff
@@ -0,0 +1,110 @@
...
+            unbuild.date_planned_finished = (
+                move.date
+            )  # Avoid to have the picking late depending on the unbuild order
+            self.subcontracted_unbuild_ids |= unbuild
+
+    def _action_done(self):
+        res = super(StockPicking, self)._action_done()
+        for picking in self:
+            unbuilds_to_done = (
+                picking.subcontracted_unbuild_ids._subcontracting_filter_to_done()
```
LoisRForgeFlow: Why don't just filter on the go with a lambda function? I only see this function used here.

mrp_unbuild_subcontracting/models/mrp_unbuild.py
```diff
@@ -0,0 +1,79 @@
...
+        "stock.picking.type",
+        "Operation Type",
+        domain="[('code', '=', 'mrp_operation'), ('company_id', '=', company_id)]",
+        default=_get_default_picking_type,
+        required=True,
+        check_company=True,
+        readonly=True,
+        states={"draft": [("readonly", False)]},
+    )
+    picking_id = fields.Many2one("stock.picking", "Operation id", readonly=True)
```
LoisRForgeFlow: "Operation id" -> "Transfer" (it is the generic name used for stock.pickings, operation id can be misleading for users)

mrp_unbuild_subcontracting/models/mrp_unbuild.py
```diff
@@ -0,0 +1,79 @@
...
+        help="Date at which you plan to start the unbuild.",
+        index=True,
+        required=True,
+    )
+    date_planned_finished = fields.Datetime(
+        "Scheduled End Date",
+        default=_get_default_date_planned_finished,
+        help="Date at which you plan to finish the unbuild.",
+        copy=False,
+    )
```
LoisRForgeFlow: Why do you need this date fields? In any case, this looks like something that can be in  generic module, not just attached to subcontracting.

mrp_unbuild_subcontracting/models/mrp_unbuild.py
```diff
@@ -0,0 +1,79 @@
...
+    picking_type_id = fields.Many2one(
```
LoisRForgeFlow: Same here, this could be generic.

mrp_unbuild_subcontracting/models/mrp_unbuild.py
```diff
@@ -0,0 +1,79 @@
...
+            self.env["stock.picking.type"]
+            .search(
+                [
+                    ("code", "=", "mrp_operation"),
+                    ("warehouse_id.company_id", "=", company_id),
+                ],
+                limit=1,
+            )
+            .id
+        )
```
LoisRForgeFlow: I think it would be great to use the same operation type used for subcontracting in standard odoo (warehouse_id.subcontracting_type_id), it will be more consistent.

mrp_unbuild_subcontracting/models/stock_move.py
```diff
@@ -0,0 +1,55 @@
...
+            res |= super(StockMove, self - move_to_not_merge)._action_confirm(
+                merge=merge, merge_into=merge_into
+            )
+            if subcontract_details_per_picking:
+                self.env["stock.picking"].concat(
+                    *list(subcontract_details_per_picking.keys())
+                ).action_assign()
+            return res
+        resultat = super(StockMove, self)._action_confirm(merge=True, merge_into=False)
+        return resultat
```
LoisRForgeFlow: Be careful with the arguments, I think here you should not override them: merge=merge, merge_into=merge_into. Plus, use English for variable names.

LoisRForgeFlow (changes requested):
Some comments below

mrp_unbuild_subcontracting/views/mrp_unbuild_views.xml
ThiagoMForgeFlow: The same functionality but in the mrp you cannot directly see the generated mrp. I have decided that the unbuilds are hidden like mrps but can be located in developer mode

JoanSForgeFlow (approved):
LGTM :+1:

mrp_unbuild_subcontracting/models/mrp_unbuild.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class MrpUnbuild(models.Model):
+    _inherit = "mrp.unbuild"
+
+    picking_id = fields.Many2one("stock.picking", "Transfer", readonly=True)
+    procurement_group_id = fields.Many2one(
+        "procurement.group", "Procurement Group", copy=False
+    )
```
LoisRForgeFlow: This also looks to be added without a purpose. Could you check?

LoisRForgeFlow (commented):
One last comment, before I do a functional review

LoisRForgeFlow (changes requested):
I have a problem in my functional test with serial numbers:

image

This is a retunr of a subcontracting order of a product tracked by serials in which there are 2 components (one tracked by serials and other by lots)

LoisRForgeFlow (approved):
I update my review, it was an issue with the initial reservation (not related with this module), but as soon as you provide a serial number, it works perfectly

PR #920, opened 2022-12-20
[REF] Fix broken website links

SirTakobi:
Fixed a few website links.

In the last commit I updated the repo template (see https://github.com/OCA/oca-addons-repo-template) because it was failing.

### DIFF
--- a/mrp_production_filter_lot/__manifest__.py
+++ b/mrp_production_filter_lot/__manifest__.py
@@ -7,7 +7,7 @@
     'version': '12.0.1.0.1',
     'category': 'Manufacturing',
     'website': "https://github.com/OCA/manufacture"
-               "12.0/mrp_production_filter_lot",
+               "/tree/12.0/mrp_production_filter_lot",
     'author': "Takobi, "
               "Odoo Community Association (OCA)",
     'license': 'AGPL-3',
--- a/quality_control_mrp/__manifest__.py
+++ b/quality_control_mrp/__manifest__.py
@@ -13,7 +13,7 @@
               "Agile Business Group, "
               "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/manufacture"
-               "quality_control_mrp",
+               "/tree/12.0/quality_control_mrp",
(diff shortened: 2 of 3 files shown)


SirTakobi:
@OCA/manufacturing-maintainers can someone please have a look?
It should be quick, this is only fixing broken links and updating CI configuration.
If you want me to split the commits in many PRs it is not a problem.

mrp_production_filter_lot/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     'version': '12.0.1.0.1',
     'category': 'Manufacturing',
     'website': "https://github.com/OCA/manufacture"
-               "12.0/mrp_production_filter_lot",
+               "/tree/12.0/mrp_production_filter_lot",
```
LoisRForgeFlow, line 10: The correct link should be only `https://github.com/OCA/manufacture`

SirTakobi: Thanks for having a look!
In https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules, both forms are accepted:
> The website key must be [https://github.com/OCA/<repo](https://github.com/OCA/%3Crepo)>, so as to provide the most relevant link to discover more information about the addon. That link shows the repository README. Alternatively [https://github.com/OCA/<repo>/tree/<branch>/<addon](https://github.com/OCA/%3Crepo%3E/tree/%3Cbranch%3E/%3Caddon)> may be used, to provide a direct link to the addon README, which includes proper credits (authors, contributors and theirs companies), and links to the relevant information on the OCA website.

And personally, I prefer the link to the module because I think that if a user is clicking on a link in a module it is expecting information about the module itself and not information about a repository that happens to contain the module.

Since both links should be accepted, can you update your review?

SirTakobi:
@OCA/manufacturing-maintainers can someone please have a look? Thanks!

gurneyalex (approved):

PR #921, opened 2022-12-20
[15.0][IMP] mrp_workorder_sequence: Multiple improvements

grindtildeath:


mrp_workorder_sequence/hooks.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import SUPERUSER_ID, api, tools
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    all_workorders = env["mrp.workorder"].search([], order="production_id ASC, id ASC")
+    for _, workorders in tools.groupby(all_workorders, lambda w: w.production_id):
+        for seq, wo in enumerate(workorders, 1):
+            wo.sequence = seq
```
ivantodorovich: I'm worried this might cause a `MemoryError` on large databases.
It's common with this kind of scripts.

The best solution to this is to use raw sql whenever possible, as it will perfom much faster.
Maybe something like this (⚠️ may need fixing, untested!)

```suggestion
    cr.execute(
        """
            UPDATE mrp_workorder w
            SET w.sequence = n.sequence
            FROM (
                SELECT
                    id,
                    ROW_NUMBER() OVER (PARTITION BY production_id) AS sequence
                FROM mrp_workorder
                ORDER BY production_id, id
            ) AS n
            WHERE w.id = n.id
        """
    )
```

----

Alternatively, look into openupgradelib's `chunked` method

mrp_workorder_sequence/models/mrp_workorder.py
```diff
@@ -1,10 +1,24 @@
...
+        for values in values_list:
+            if values.get("sequence"):
+                continue
+            last_production_wo = self.search(
+                [("production_id", "=", values.get("production_id"))],
+                order="sequence DESC",
+                limit=1,
+            )
+            if last_production_wo:
+                values["sequence"] = last_production_wo.sequence + 1
```
ivantodorovich: NITPICKING

It seems this code will give you more than 1 record with the exact same sequence in case you create multiple workorders for the same `production_id` at once.

Not sure this is acceptable or not (out of context).

Also, performance wise it's not a good idea to have a `search` within a loop.

Maybe you could identify all the `vals` that need a new sequence before-hand, extract their `production_id` and search for all of them at once using `read_group` and the `sequence:max` aggregator.

This also allows you to set unique sequences in the case described above, as you'd have everything in memory by the time you enter the loop, you can maintain a dict mapping with `{production_id: last_sequence}`, and increment that number in-memory as you process them.

```py
    for values in vals_list:
        if values.get("sequence"):
            continue
        production_id = values["production_id"]
        last_sequence_by_production.setdefault(production_id, 0)
        last_sequence_by_production[production_id] += 1
        values["sequence"] = last_sequence_by_production[production_id]
```

ivantodorovich (approved):

mrp_workorder_sequence/hooks.py
grindtildeath: @ivantodorovich That looks quite good actually, I agree it's less risky to raise a MemoryError

mrp_workorder_sequence/models/mrp_workorder.py
grindtildeath: Still working on this one

LoisRForgeFlow:
@grindtildeath Thanks for the proposal, it looks in general correct. From your comment above I understand it is still a work in progress, I will label it as such.

Please, ping me when ready to review.

mrp_workorder_sequence/models/mrp_workorder.py
grindtildeath: Solved in fixup commit below.

grindtildeath:
@LoisRForgeFlow Thanks for checking in. The override of `create` function is improved in latest commit, I'll rebase and squash when ready to merge after reviews.

ivantodorovich (approved):

LoisRForgeFlow:
@grindtildeath functional review LGTM :+1: 

Could you squash the commits? we should be good to merge.

sebalix (approved):

grindtildeath:
@LoisRForgeFlow Thank you.

Just wondering, before we merge, shouldn't we do this change in this module? 

https://github.com/OCA/manufacture/pull/922/files#diff-0b88d6e9569b2683f2777f452068029449a562f57c984924af27f423eb80af75R10

I didn't do it here since it wasn't strictly needed and could be a breaking change, but IMO it would make more sense to have it in this module. (but I also understand one could prefer keeping the original `_order`)

LoisRForgeFlow:
> @LoisRForgeFlow Thank you.
> 
> Just wondering, before we merge, shouldn't we do this change in this module?
> 
> https://github.com/OCA/manufacture/pull/922/files#diff-0b88d6e9569b2683f2777f452068029449a562f57c984924af27f423eb80af75R10
> 
> I didn't do it here since it wasn't strictly needed and could be a breaking change, but IMO it would make more sense to have it in this module. (but I also understand one could prefer keeping the original `_order`)

I actually thought it when testing, so yes, if you don't mind I think it makes sense to include it here.

grindtildeath:
@LoisRForgeFlow done :+1:

LoisRForgeFlow:
@grindtildeath thanks! time to squash then.

LoisRForgeFlow (approved):
LGTM :+1:

PR #922, opened 2022-12-20
[15.0][ADD] Create module mrp_production_operation_injection

grindtildeath:
This module provides a wizard to add extra workorders based on existing BOM
operations, and to select where such workorder should be added.

Depends on:
 - [x] #921

### DIFF
--- a/mrp_production_operation_injection/__manifest__.py
+++ b/mrp_production_operation_injection/__manifest__.py
@@ -1,0 +1,24 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "MRP Production Inject Operation",
+    "summary": "Adds an existing operation from the Bill of Material",
+    "version": "15.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Manufacturing",
+    "website": "https://github.com/OCA/manufacture",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["grindtildeath"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "mrp_workorder_sequence",
+    ],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/mrp_production.xml",
+        "views/mrp_workorder.xml",
+        "wizard/mrp_workorder_injector.xml",
+    ],
+}
--- a/mrp_production_operation_injection/models/mrp_production.py
+++ b/mrp_production_operation_injection/models/mrp_production.py
@@ -1,0 +1,72 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    display_inject_workorder = fields.Boolean(
+        compute="_compute_display_inject_workorder"
+    )
+
+    @api.depends("state", "bom_id", "bom_id.operation_ids", "workorder_ids")
+    def _compute_display_inject_workorder(self):
+        for production in self:
+            production.display_inject_workorder = (
+                production.state in ["confirmed", "progress", "to_close"]
+                and production.bom_id.operation_ids
+                and production.workorder_ids
+            )
+
+    def action_open_workorder_injector(self):
+        self.ensure_one()
+        action = self.env["ir.actions.act_window"]._for_xml_id(
+            "mrp_production_operation_injection.mrp_workorder_injector_action"
+        )
+        ctx = self.env.context.copy()
+        ctx["default_production_id"] = self.id
+        action.update({"context": ctx})
+        return action
+
+    def _prepare_injected_workorder_values(self, operation):
+        self.ensure_one()
+        return {
+            "name": operation.name,
+            "production_id": self.id,
+            "workcenter_id": operation.workcenter_id.id,
+            "product_uom_id": self.product_uom_id.id,
+            "operation_id": operation.id,
+            "state": "pending",
+        }
+
+    def _add_workorder(self, operation, previous_workorder):
+        self.ensure_one()
+        following_workorders = self.workorder_ids.filtered(
+            lambda w: w.sequence > previous_workorder.sequence
+        )
+        next_workorder = fields.first(following_workorders)
+        # Prepare creation of new workorder
+        workorder_values = self._prepare_injected_workorder_values(operation)
+        workorder_values["sequence"] = previous_workorder.sequence + 1
+        workorder_values["next_work_order_id"] = next_workorder.id
+        # FIXME: state computation is not good in Odoo anyway so handle
+        #  only most 'probable' cases only
+        if next_workorder.state in ["ready", "progress"]:
+            workorder_values["state"] = "ready"
+        # Update following workorders sequence before create to make sure workorders
+        #  can be ordered properly for _action_confirm (cf override in mrp_workorder)
+        for wo in following_workorders:
+            wo.sequence += 1
+        new_workorder = self.env["mrp.workorder"].create(workorder_values)
+        # Update next workorder
+        # FIXME: state computation is not good in Odoo anyway so handle
+        #  only most 'probable' cases only
+        if next_workorder.state == "ready":
+            next_workorder.state = "pending"
+        new_workorder.duration_expected = new_workorder._get_duration_expected()
+        # Replan if needed after cache invalidation to make sure all workorders are considered
+        self.invalidate_cache()
+        if self.is_planned:
+            self._plan_workorders(replan=True)
+        return True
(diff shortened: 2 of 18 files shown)


mrp_production_operation_injection/wizard/mrp_workorder_injector.xml
```diff
@@ -0,0 +1,44 @@
...
+        <field name="name">mrp.workorder.injector.form.view</field>
+        <field name="model">mrp.workorder.injector</field>
+        <field name="arch" type="xml">
+            <form string="Add operation to workorders">
+                <group>
+                    <field name="production_id" invisible="1" />
+                    <field name="bom_id" invisible="1" />
+                    <field name="allowed_bom_operation_ids" invisible="1" />
+                    <field name="production_workorder_ids" invisible="1" />
+                    <!-- TODO: Add domains -->
```
sebalix:
```suggestion
```
It is done now, right?

mrp_production_operation_injection/wizard/mrp_workorder_injector.xml
```diff
@@ -0,0 +1,44 @@
...
+                        name="operation_id"
+                        domain="[('id', 'in', allowed_bom_operation_ids)]"
+                        options="{'no_create': True, 'no_quick_create': True, 'no_create_edit':True}"
+                    />
+                    <field
+                        name="workorder_id"
+                        domain="[('id', 'in', production_workorder_ids)]"
+                        options="{'no_create': True, 'no_quick_create': True, 'no_create_edit':True}"
+                    />
+                    <!-- <field name="position" /> -->
```
sebalix: What to do with this? To remove?

mrp_production_operation_injection/tests/test_mrp_production_inject_operation.py
```diff
@@ -0,0 +1,256 @@
...
+    def _get_new_workorder(cls, previously_existing_wos, existing_wos):
+        return previously_existing_wos - existing_wos
+
+    @classmethod
+    def _record_time_tracking(cls, workorder, duration, productivity):
+        workorder_form = Form(workorder)
+        with workorder_form.time_ids.new() as time_tracking_form:
+            time_tracking_form.date_end = time_tracking_form.date_start + relativedelta(
+                seconds=duration
+            )
```
sebalix: Nitpick to avoid the direct import of third module and use Odoo API:
```suggestion
            time_tracking_form.date_end = fields.Datetime.add(
                time_tracking_form.date_start, seconds=duration
            )
```

grindtildeath: Didn't know about this one, nice :+1:

sebalix (approved):

mrp_production_operation_injection/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022 Camptocamp SA
```
JuMiSanAr, line 1: If we're adding this module now, shouldn't this be 2023?

mrp_production_operation_injection/models/mrp_production.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    display_inject_workorder = fields.Boolean(
+        compute="_compute_display_inject_workorder"
+    )
+
+    @api.depends("state", "bom_id", "bom_id.operation_ids", "workorder_ids")
```
JuMiSanAr, line 13: Isn't `bom_id` unnecessary here? We should only need `bom_id.operations_ids` right?

mrp_production_operation_injection/models/mrp_production.py
```diff
@@ -0,0 +1,72 @@
...
+            "operation_id": operation.id,
+            "state": "pending",
+        }
+
+    def _add_workorder(self, operation, previous_workorder):
+        self.ensure_one()
+        following_workorders = self.workorder_ids.filtered(
+            lambda w: w.sequence > previous_workorder.sequence
+        )
+        next_workorder = fields.first(following_workorders)
```
mmequignon, line 48: <s>do we have any guarantee that `first()` will return the next workorder based on its sequence?</s>
Explained afterwards

mrp_production_operation_injection/views/mrp_workorder.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="mrp_production_workorder_tree_view_inherit" model="ir.ui.view">
+        <field name="name">mrp.workorder.tree</field>
```
JuMiSanAr, line 4: The name for the mrp production form is `<field name="name">mrp.production.form.inherit</field>`, maybe add `.inherit` here as well so both name structures are consistent.

mrp_production_operation_injection/models/mrp_production.py
```diff
@@ -0,0 +1,72 @@
...
+        # Prepare creation of new workorder
+        workorder_values = self._prepare_injected_workorder_values(operation)
+        workorder_values["sequence"] = previous_workorder.sequence + 1
```
mmequignon, line 51: <s>what if `previous_workorder.sequence + 1 == next_workorder.sequence` ?</s>
nvm, it's right after this

JuMiSanAr (approved):

mrp_production_operation_injection/models/mrp_production.py
```diff
@@ -0,0 +1,72 @@
...
+        workorder_values["next_work_order_id"] = next_workorder.id
+        # FIXME: state computation is not good in Odoo anyway so handle
+        #  only most 'probable' cases only
+        if next_workorder.state in ["ready", "progress"]:
+            workorder_values["state"] = "ready"
+        # Update following workorders sequence before create to make sure workorders
+        #  can be ordered properly for _action_confirm (cf override in mrp_workorder)
+        for wo in following_workorders:
+            wo.sequence += 1
```
mmequignon, line 60: can it happen that previous_workorder.sequence == next_workorder.sequence?
In such case after this, you would have new_workorder.sequence == next_workorder.sequence.
Is that ok?
Otherwise smth like that maybe?
```python
next_sequence = previous_workorder.sequence + 1
next_sequences = range(next_sequence, next_sequence + len(following_workorders))
for wo, wo_sequence in zip(following_workorders, next_sequences):
    wo.sequence = wo_sequence
```

mrp_production_operation_injection/__manifest__.py
grindtildeath: It was written and submitted last year, so it must be 2022.

mrp_production_operation_injection/wizard/mrp_workorder_injector.py
```diff
@@ -0,0 +1,67 @@
...
+    bom_id = fields.Many2one("mrp.bom", related="production_id.bom_id")
+    allowed_bom_operation_ids = fields.Many2many(
+        "mrp.routing.workcenter", compute="_compute_allowed_bom_operation_ids"
+    )
+    production_workorder_ids = fields.Many2many(
+        "mrp.workorder", compute="_compute_production_workorder_ids"
+    )
+    operation_id = fields.Many2one(
+        "mrp.routing.workcenter", "New operation", required=True
+    )
```
mmequignon, line 22: isn't it possible to set a domain here, instead of computing the allowed operations?

mrp_production_operation_injection/wizard/mrp_workorder_injector.py
```diff
@@ -0,0 +1,67 @@
...
+    workorder_id = fields.Many2one("mrp.workorder", "Previous workorder", required=True)
```
mmequignon, line 23: same here

mrp_production_operation_injection/views/mrp_workorder.xml
grindtildeath: The view I'm inheriting here is from `mrp.workorder`, we can nitpick on the name but IMO there's no need to

mrp_production_operation_injection/models/mrp_production.py
grindtildeath: I'm not sure if it's still needed in the newer versions, but it won't harm.

mrp_production_operation_injection/wizard/mrp_workorder_injector.py
grindtildeath: As the domain must be dynamic (according to the `mrp.production` we're injecting an operation on), we need to compute what's allowed anyway.

mrp_production_operation_injection/wizard/mrp_workorder_injector.py
grindtildeath: same answer as above :grin:

mrp_production_operation_injection/models/mrp_production.py
grindtildeath: I would say we don't care because this case would have to be handled by the module providing workorder sequence: https://github.com/OCA/manufacture/tree/15.0/mrp_workorder_sequence

grindtildeath:
CI error must be fixed by #1128

PR #934, opened 2023-01-04
[14.0][ADD] mrp_lot_production_date

sebalix:
This module automatically set a production date on lots/SN produced through manufacturing orders.

Depends on:
- https://github.com/OCA/stock-logistics-workflow/pull/1162

Ref. 3859


sebalix:
Can't succeed to make the `test-requirements.txt` working, I'll wait for the dependency to be merged.

mrp_lot_production_date/models/mrp_production.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def action_generate_serial(self):
+        res = super().action_generate_serial()
+        if self.lot_producing_id.product_id.use_production_date:
+            self.lot_producing_id.production_date = fields.Datetime.now()
```
grindtildeath: This will set the production to when the user clicks on the button in the manufacturing order. IMO that's good as default if the field is required (cf my comment https://github.com/OCA/stock-logistics-workflow/pull/1162#pullrequestreview-1237598249 ).
But depending on each company's process the lot number could be set before the the manufacturing is processed, in which case the production date probably needs to be overwritten later in their process.

sebalix: In our use-case we let the MO creating the lot automatically (`action_generate_serial` is called by the production wizard).
I could move this in the wizard process method instead, could be a better default behavior...
thanks!

sebalix: NOTE to myself: indicate clearly in the readme that the date will be set at the validation of the MO (and not at lot creation).

ivantodorovich:
Is there a reason to have a "Production Date" on lots, without the "MRP" module installed?
Is it to manually fill in the value for purchased goods where we still want to know the manufactured date? 🤔 

If not, I would merge the two modules into this one: `mrp_lot_production_date`.


---

Also, but independently from the previous point..

Is there a reason for not wanting the production date in some products?
If not, then I suggest to drop the `product.template` setting and simply store it everytime.
Then the user can choose to use it or not.

sebalix:
> Is there a reason to have a "Production Date" on lots, without the "MRP" module installed? Is it to manually fill in the value for purchased goods where we still want to know the manufactured date?

This, you got it. I prefer to separate concerns. I worked one day for a pharmaceutical company and you want to track everything (like the production date) even for goods you don't produce yourself, for traceability reasons.

> Is there a reason for not wanting the production date in some products? If not, then I suggest to drop the `product.template` setting and simply store it everytime. Then the user can choose to use it or not.

It could be, I simply align this on `product_expiry` with its `use_expiration_date`/`expiration_date`. Maybe no need to enable this feature on `product.template` indeed :+1:

ivantodorovich (approved):

ivantodorovich:
`stock_production_lot_date` got merged. 
You can rebase this one ;)

grindtildeath (approved):

mrp_lot_production_date/models/mrp_production.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def _post_inventory(self, cancel_backorder=False):
+        res = super()._post_inventory(cancel_backorder=cancel_backorder)
+        for order in self:
+            if order.lot_producing_id:
```
JuMiSanAr:
```suggestion
            if order.lot_producing_id and not order.lot_producing_id.production_date:
```

If we need to set the production_date before, at least this won't overwrite the existing value.

sebalix: done, thanks!

grindtildeath:
@OCA/manufacturing-maintainers can we please reopen this one? :pray:

PR #949, opened 2023-01-18
[15.0] [ADD] `mrp_production_split`

ivantodorovich:
This module adds a "Split" button to Manufacturing Orders.

Manufacturing Orders can be split as long as they haven't been completed yet.

For products tracked by "Serial Number", it allows to choose the Quantity to extract from the original MO, and it'll create one MO per single unit.

For other products, more options are available, that will let you do things like:

* Extract 10 units from a MO into 5 MOs of 2 units each.
* Extract 10 units from a MO into a single new MOs.
* Extract 10 units from a MO into multiple MOs of different quantities.

### DIFF
--- a/mrp_production_split/models/mrp_production.py
+++ b/mrp_production_split/models/mrp_production.py
@@ -1,0 +1,34 @@
+# Copyright 2023 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def copy_data(self, default=None):
+        # OVERRIDE copy the date_planned_start and date_planned_end when splitting
+        # productions, as they are not copied by default (copy=False).
+        [data] = super().copy_data(default=default)
+        data.setdefault("date_planned_start", self.date_planned_start)
+        data.setdefault("date_planned_finished", self.date_planned_finished)
+        return [data]
+
+    def action_split(self):
+        self.ensure_one()
+        self._check_company()
+        if self.state in ("draft", "done", "to_close", "cancel"):
+            raise UserError(
+                _(
+                    "Cannot split a manufacturing order that is in '%s' state.",
+                    self._fields["state"].convert_to_export(self.state, self),
+                )
+            )
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "mrp_production_split.action_mrp_production_split_wizard"
+        )
+        action["context"] = {"default_production_id": self.id}
+        return action
--- a/mrp_production_split/tests/common.py
+++ b/mrp_production_split/tests/common.py
@@ -1,0 +1,92 @@
+# Copyright 2023 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import Command
+from odoo.tests import Form, TransactionCase
+
+
+class CommonCase(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        # Create bom, product and components
+        cls.component = cls.env["product.product"].create(
+            {
+                "name": "Component",
+                "detailed_type": "product",
+            }
+        )
+        cls.product = cls.env["product.product"].create(
+            {
+                "name": "Product",
+                "detailed_type": "product",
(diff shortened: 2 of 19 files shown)


mrp_production_split/models/mrp_production.py
```diff
@@ -0,0 +1,23 @@
...
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def action_split(self):
+        self.ensure_one()
+        self._check_company()
+        if self.state in ("draft", "done", "to_close", "cancel"):
+            raise UserError(
+                _("Cannot split a manufacturing order that is done or cancelled.")
+            )
```
yankinmax: if can't split in done or canceled, why draft is in states?

yankinmax (approved):
Thanks for the work done

mrp_production_split/models/mrp_production.py
ivantodorovich: Oops, I forgot to update the message string.
It's changed now ;) thanks!

SilvioC2C (approved):
LGTM

yankinmax:
@rvalyi @simahawk can you pls merge it?

mrp_production_split/models/mrp_production.py
```diff
@@ -0,0 +1,34 @@
...
+from odoo.exceptions import UserError
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def copy_data(self, default=None):
+        # OVERRIDE copy the date_planned_start and date_planned_end when splitting
+        # productions, as they are not copied by default (copy=False).
+        [data] = super().copy_data(default=default)
```
simahawk, line 15: curious: why this notation `[data]` ?

mrp_production_split/tests/common.py
```diff
@@ -0,0 +1,91 @@
...
+
+from odoo import Command
+from odoo.tests import Form, TransactionCase
+
+
+class CommonCase(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Create bom, product and components
```
simahawk, line 14: can you disable tracking pls?

mrp_production_split/wizards/mrp_production_split_wizard.py
```diff
@@ -0,0 +1,167 @@
...
+            return quantities
+        elif self.split_mode == "custom":
+            quantities = self._parse_float_list(self.custom_quantities)
+            split_total = sum(quantities)
+            split_rest = production.product_qty - split_total
+            if not fields.Float.is_zero(split_rest, precision_rounding=rounding):
+                quantities = [split_rest] + quantities
+            return quantities
+        else:  # pragma: no cover
+            raise UserError(_("Invalid Split Mode: '%s'", self.split_mode))
```
simahawk, line 135:
```suggestion
            raise UserError(_("Invalid Split Mode: '%(mode)s'" % {"mode": self.split_mode}))
```

mrp_production_split/wizards/mrp_production_split_wizard.py
```diff
@@ -0,0 +1,167 @@
...
+from typing import List, Union
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class MrpProductionSplitWizard(models.TransientModel):
+    _name = "mrp.production.split.wizard"
+
+    production_id = fields.Many2one(
```
simahawk, line 14: readonly?

ivantodorovich: Not needed in fact, the field is not shown. And making it readonly would probably require `force_save="1"`
But also doesn't make much sense for a wizard, IMO.

mrp_production_split/tests/common.py
ivantodorovich: Absolutely

mrp_production_split/models/mrp_production.py
ivantodorovich: Beats me. It's like that in core

https://github.com/odoo/odoo/blob/eb3650f64e393e0b4b0353c903b8b10e70c21602/odoo/models.py#L4791

mrp_production_split/wizards/mrp_production_split_wizard.py
ivantodorovich: I do tend to do this when `len(args) > 1`, definitely a must for translations, but not if `len(args) == 1` as it seems redundant.

mrp_production_split/models/mrp_production.py
simahawk: makes no sense to me but yeah....

mrp_production_split/wizards/mrp_production_split_wizard.py
simahawk: fact is: in v16 you'll have to change it ;)
Not a blocker, we can merge if you don't care

ivantodorovich: 😱 what am I missing?

PR #954, opened 2023-01-26
[15.0][IMP] mrp_multi_level: better group operators

LoisRForgeFlow:
For initial OH, final OH and planned availability grouping with sum does not provide any value, specially when grouping by product. And avg though can be more interesting.

For demand and supply we could have more doubts and the sum can make sense, so we keep it as is.


BernatPForgeFlow (approved):
LGTM!

PR #960, opened 2023-01-31
[15.0] [ADD] mrp_workorder_ref

ivantodorovich:


ivantodorovich:
ping @OCA/manufacturing-maintainers : could you add the 'no stale' label please?

mrp_workorder_ref/models/mrp_workorder.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # OVERRIDE to auto-complete the reference from the operation
+        for vals in vals_list:
+            if "ref" not in vals and vals.get("operation_id"):
+                operation_id = vals["operation_id"]
+                operation = self.env["mrp.routing.workcenter"].browse(operation_id)
+                if operation.ref:
+                    vals["ref"] = operation.ref
```
grindtildeath, line 29: IMO would be cleaner to have this in a dedicated function.

grindtildeath (commented):
Won't there be any issue with the SQL constraints at installation if the column `ref` is empty?

mrp_workorder_ref/models/mrp_workorder.py
ivantodorovich: Hm, I'm not sure I agree. I tend to keep patterns that manipulate `vals` directly in `create` / `write` overrides, with the only exception where they are too complex or shared among them. But in this case it's neither

ivantodorovich (commented):
> Won't there be any issue with the SQL constraints at installation if the column ref is empty?

Not at all, the `UNIQUE` constraint treats `NULL` values as different, so it only really plays for rows having both values set

gurneyalex (approved):

bosd (approved):

LoisRForgeFlow (approved):
LGTM

/ocabot merge nobump

PR #967, opened 2023-02-08
[14.0][IMP] mrp_multi_level: Supply method computation for MRP Parameters

BernatPForgeFlow:
For each MRP Parameter, we will calculate its supply method based on the procurement rules path. We will go back in the rules until we find that the action is "buy" or "manufacture", or until the action is "Pull From" or "Pull & Push" and the supply method is "Take from Stock".

This means we will show what the latest rule is if we were to do a procurement. Although it would be possible that the final action does not end up being executed if stock already exists in one of the intermediate locations.

### DIFF
--- a/mrp_multi_level/demo/product_product_demo.xml
+++ b/mrp_multi_level/demo/product_product_demo.xml
@@ -36,6 +36,14 @@
             eval="[(6, 0, [ref('mrp.route_warehouse0_manufacture')])]"
         />
     </record>
+    <record id="product_product_fp_4" model="product.product">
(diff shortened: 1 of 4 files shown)


JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
CI issues have been fixed with https://github.com/OCA/manufacture/pull/984.

/ocabot merge patch

LoisRForgeFlow:
@BernatPForgeFlow Can you port this to 13, 15 and 16?

PR #975, opened 2023-02-13
[15.0][MIG] mrp_bom_line_sequence

JuanyDForgeflow:


AaronHForgeFlow:
Edit: well, actually it works, because the sequence is not propagated to any other place in Odoo. Actually the description of the module is misleading because the sequence is not propagated anywhere, but standard Odoo will take the same order when creating an MO for the BOM. The refactor is not needed in this case.

AaronHForgeFlow (commented):
Fuctional tests ok!

AaronHForgeFlow (approved):

JasminSForgeFlow (approved):
Functional Test

LoisRForgeFlow (approved):
/ocabot merge nobump

PR #978, opened 2023-02-17
[15.0][MIG] quality_control_stock_oca

dalonsod:
Standard migration


dalonsod:
@jonerikceberio @CarlosRoca13 could you review? Thanks!

CarlosRoca13 (approved):
Code review

quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -97,7 +97,7 @@ def setUp(self):
     def _create_user(self, login, groups, company):
```
pedrobaeza: @victoralmau, how is the existing helper method in core for doing this?

victoralmau: https://github.com/odoo/odoo/blob/15.0/odoo/tests/common.py#L110-L154

pedrobaeza: @dalonsod please use this method instead.

pedrobaeza (approved):
But there's a failing unrelated test

@LoisRForgeFlow is with `mrp_multi_level`. Can you check it?

LoisRForgeFlow:
@dalonsod @pedrobaeza There have been some merges in the last days in v15 without issues with mrp_multi_level tests, could you rebase this PR to be sure it wasn't something punctual?

Thanks,

dalonsod:
Rebase made, still in error 😟

LoisRForgeFlow:
@dalonsod Then there must be some conflicting functions somewhere (it can be a third module in combination with these 2) or some issue in OCA CI. I quickly run the tests in my local with both quality_contro_stock_oca and mrp_multi_level isntalled, and they all passed.

dalonsod:
Thanks @LoisRForgeFlow .
Which should be the next step in these cases?

LoisRForgeFlow:
@dalonsod You could try to install all OCA/manfuacture addons available in your local to try to be able to reproduce the error. If with that you still don't get the error, you will need to run tests using the OCA CI image.

dalonsod:
@LoisRForgeFlow added this addon as rebel module, but test error persists in the same way. Maybe am I doing something wrong with this?

LoisRForgeFlow:
> @LoisRForgeFlow added this addon as rebel module, but test error persists in the same way. Maybe am I doing something wrong with this?

After editing the copier answers like this, you need to run the following command for the change to be applied to the relevant files: `copier -f update`

LoisRForgeFlow:
Also, could you do this in a separate commit from migration one? using that commit to also add a good explanation of why it is neede to mark it as rebel.

dalonsod:
> After editing the copier answers like this, you need to run the following command for the change to be applied to the relevant files: `copier -f update`

Sorry, I'm newbie with copier... `copier update -f` returns `Template uses unsafe feature: tasks` error and no changes have been made.

quality_control_stock_oca/models/stock_picking_type.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import api, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    def _create_qc_trigger(self):
+        for picking_type in self:
+            qc_trigger = {
+                "name": picking_type.name,
```
Saran440:
```suggestion
                "name": picking_type.name,
```

if we have 2 warehouse it will show same name. Is it posible to add hook function or change it to uniqe?

dalonsod: Yes, what do you think about using `display_name` instead of `name` for `picking_type`? It'll be shown e.g. "Warehouse: operation"

Saran440: @dalonsod I agree to use `display_name`

dalonsod: Done! Could you approve this? 😇

dalonsod:
@LoisRForgeFlow test errors are gone! Is this ready to be merged?

Saran440 (approved):

.copier-answers.yml
```diff
@@ -14,7 +14,8 @@ include_wkhtmltopdf: false
 odoo_version: 15.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
```
Saran440: This file should not change manual

pedrobaeza:
@dalonsod why do you need to put it as rebel module?

dalonsod:
Tests fail for `mrp_multi_level` and couldn't find why, you can see PR history. After a talk with @LoisRForgeFlow I've changed to this approach. Is that correct from your poing of view?

pedrobaeza:
Please remove that part to let me see which is the error in the CI, so I can tell you.

dalonsod:
@pedrobaeza removed rebel module configuration, today it works 🤔

If it's OK, I'll remove unnecessary commits and should it be merged?

pedrobaeza:
Yes, please

dalonsod:
Done!

PR #984, opened 2023-02-23
[14.0][FIX] mrp_component_operation: fix test and small issues

DavidJForgeFlow:
The test were not passing when also installing mrp_multi_level, the pickings weren't in the desired order. 
Also there was a problem when only 1 picking was in the system.

@BernatPForgeFlow @LoisRForgeFlow


LoisRForgeFlow (approved):
LGTM

BernatPForgeFlow (approved):

PR #985, opened 2023-02-23
[14.0][FIX] mrp_subcontracting_purchase_link: link to PO line, not PO

ThiagoMForgeFlow:
The mrps related with a subcontracting process, there are linked to the purchase line associated, not the purchase order

@ForgeFlow

### DIFF
--- a/mrp_subcontracting_purchase_link/models/mrp_production.py
+++ b/mrp_subcontracting_purchase_link/models/mrp_production.py
@@ -8,5 +8,13 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     purchase_order_id = fields.Many2one(
-        "purchase.order", "Subcontract Purchase Order", readonly=True
+        "purchase.order",
+        "Subcontract Purchase Order",
+        readonly=True,
+        related="purchase_line_id.order_id",
+        store=True,
+    )
+
+    purchase_line_id = fields.Many2one(
+        "purchase.order.line", "Subcontract Purchase Order Line", readonly=True
     )
--- a/mrp_subcontracting_purchase_link/__manifest__.py
+++ b/mrp_subcontracting_purchase_link/__manifest__.py
@@ -1,7 +1,7 @@
 # Copyright 2020 Quartile Limited
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 {
-    "name": "Link Purchase Order to Subcontract Productions",
+    "name": "Link Purchase Order Line to Subcontract Productions",
     "version": "14.0.1.0.1",
     "category": "Manufacturing",
     "license": "LGPL-3",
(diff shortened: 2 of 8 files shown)


JordiBForgeFlow:
@yostashiro do you agree with this?

MiquelRForgeFlow:
It would need migration scripts...

JordiBForgeFlow (approved):
@MiquelRForgeFlow After the changes made by @ThiagoMForgeFlow no migration is required. I just tested that. I installed the module in the previous version, created a subcontracting PO and then I upgraded the module to this PR, and the link is maintained.

JordiBForgeFlow:
cc @LoisRForgeFlow

MiquelRForgeFlow:
@JordiBForgeFlow Yes, I see. But, another thing is bothering me. Imagine having a mrp.production record with purchase_order_id filled. Then implement this change. The field purchase_order_id will be still be filled, because it's stored, but it will be not making much sense as it will be related to an empty purchase_line_id. I think, shouldn't we execute an script to fill the purchase_line_id?

JordiBForgeFlow:
@MiquelRForgeFlow your desires have been accomplished.

image

JordiBForgeFlow:
To merge use major, otherwise the migration will not run

MiquelRForgeFlow (approved):
Thank you, Genie.

JordiBForgeFlow:
@LoisRForgeFlow This one is ready to merge

mrp_subcontracting_purchase_link/models/mrp_production.py
```diff
@@ -8,5 +8,13 @@ class MrpProduction(models.Model):
...
     purchase_order_id = fields.Many2one(
-        "purchase.order", "Subcontract Purchase Order", readonly=True
+        "purchase.order",
+        "Subcontract Purchase Order",
+        readonly=True,
+        related="purchase_line_id.order_id",
+        store=True,
+    )
+
+    purchase_line_id = fields.Many2one(
```
yostashiro, line 18: Another approach could be to make this a stored compute field with a compute method like below, removing the extension of `_prepare_subcontract_mo_vals()`.

```
@api.depends("move_dest_ids")
def _compute_purchase_line_id(self):
    for production in self:
        if production.move_dest_ids:
            production.purchase_line_id = production.move_dest_ids[0].purchase_line_id
```

If we take this approach, though, we may need to prepare a pre-init hook script for newly installing this module in an existing database with many subcontracting transactions.

yostashiro (approved):
The proposed change makes sense to me. I'm curious - did you face some inconveniences for not having the link between manufacturing order and purchase order line?

mrp_subcontracting_purchase_link/models/mrp_production.py
ThiagoMForgeFlow: In the case of subcontracting, it makes sense to have the purchase order line in the manufacturing order, as it provides coherence when passed in _prepare_subcontract_mo_vals because you have it directly

JordiBForgeFlow:
@LoisRForgeFlow can you merge this one?

LoisRForgeFlow (approved):

PR #991, opened 2023-03-01
[14.0][IMP] mrp_multi_level: Get BoM to explode 

BernatPForgeFlow:
When exploding requirements or calculating supply method, we will consider the first active BoM taking into account the routes.

If the last supply rule is 'Manufacture' and the first active BoM other than 'Subcontracting' is a 'Kit', the supply method for the  MRP Parameter will be 'Kit'. Otherwise, if the last supply rule is 'Manufacture' and there is no BoMs or the first active BoM other than 'Subcontracting' is 'Manufactured', the supply method for the  MRP Parameter will be 'Produce'.

In the image below we have an example of a product with multiple active BoMs. If the route is 'Manufacture', the 'Kit' BoM is the one that will be considered.
image

### DIFF
--- a/mrp_multi_level/wizards/mrp_multi_level.py
+++ b/mrp_multi_level/wizards/mrp_multi_level.py
@@ -8,6 +8,7 @@ import logging
 from datetime import date, timedelta
 
 from odoo import _, api, exceptions, fields, models
+from odoo.tools import float_is_zero
 
 logger = logging.getLogger(__name__)
 
@@ -193,6 +194,17 @@ class MultiLevelMrp(models.TransientModel):
         return mrp_action_date, mrp_date_supply
 
     @api.model
+    def _get_bom_to_explode(self, product_mrp_area_id):
+        boms = self.env["mrp.bom"]
+        if product_mrp_area_id.supply_method in ["manufacture", "phantom"]:
+            boms = product_mrp_area_id.product_id.bom_ids.filtered(
+                lambda x: x.type in ["normal", "phantom"]
+            )
+        if not boms:
+            return False
+        return boms[0]
+
+    @api.model
     def explode_action(
         self, product_mrp_area_id, mrp_action_date, name, qty, action, values=None
     ):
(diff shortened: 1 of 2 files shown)


mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -200,45 +200,48 @@ def explode_action(
         mrp_date_demand = mrp_action_date
         if mrp_date_demand < date.today():
             mrp_date_demand = date.today()
-        if not product_mrp_area_id.product_id.bom_ids:
+        boms = self.env["mrp.bom"]
+        if product_mrp_area_id.supply_method == "buy":
+            boms = product_mrp_area_id.product_id.bom_ids.filtered(
+                lambda x: x.type == "subcontract"
```
LoisRForgeFlow: this is added in `mrp_subcontracting`, you should consider create a new glue module `mrp_multi_level_subcontracting`

BernatPForgeFlow: Done!

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -200,45 +211,39 @@ def explode_action(
...
         if mrp_date_demand < date.today():
             mrp_date_demand = date.today()
-        if not product_mrp_area_id.product_id.bom_ids:
+        bom = self._get_bom_to_explode(product_mrp_area_id)
+        if not bom:
             return False
-        bomcount = 0
-        for bom in product_mrp_area_id.product_id.bom_ids:
-            if not bom.active or not bom.bom_line_ids:
+        for bomline in bom.bom_line_ids:
```
hieulucky111, line 219: we can use `bom.bom_line_ids.filtered(lambda r: r.product_qty < 0 or r.product_id.type != "product")`

hieulucky111 (approved):
LGTM
Code review

mrp_multi_level/wizards/mrp_multi_level.py
BernatPForgeFlow: Yes, that's true! 

Well, it really is `bom.bom_line_ids.filtered(lambda r: r.product_qty >= 0 or r.product_id.type != "product")`, but thanks!

hieulucky111 (approved):

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -200,45 +211,39 @@ def explode_action(
...
+            ):
+                # Stop explosion.
                 continue
-            bomcount += 1
-            if bomcount != 1:
+            if bomline._skip_bom_line(product_mrp_area_id.product_id):
                 continue
-            for bomline in bom.bom_line_ids:
-                if bomline.product_qty <= 0.00 or bomline.product_id.type != "product":
-                    continue
```
LoisRForgeFlow, line 214: @BernatPForgeFlow Can you keep this inside the loop instead of applying it with a filtered? with the "for bom in boms.filtered()" you are iterating twice.

BernatPForgeFlow: That's true! I'll go back to the initial version then

LoisRForgeFlow (approved):
LGTM

PR #992, opened 2023-03-01
[14.0][MIG] product_quick_bom: Migration to 14.0

cuongnmtm:


hieulucky111 (approved):
LGTM
Code review

product_quick_bom/models/product.py
```diff
@@ -0,0 +1,34 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.depends("bom_ids", "bom_ids.active")
+    def _compute_bom_id(self):
+        for rec in self:
+            if len(rec.bom_ids.ids) == 1:
+                rec.bom_id = rec.bom_ids[0]
+            else:
+                rec.bom_id = self.env["mrp.bom"]
```
DucTruongKomit, line 16:
```suggestion
            rec.bom_id = rec.bom_ids[0] if len(rec.bom_ids.ids) == 1 else self.env["mrp.bom"]
```

Leonidas-VII: Hi Mr @DucTruongKomit 
It seems `.ids` is redundant, how about only `rec.bom_ids`?

PR #998, opened 2023-03-07
[14.0][ADD] Mrp Workorder Cancel

Hardik-OSI:


mrp_workorder_cancel/models/mrp_workorder.py
```diff
@@ -0,0 +1,27 @@
...
+
+class MrpWorkorder(models.Model):
+    _inherit = "mrp.workorder"
+
+    def button_cancel(self):
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "mrp_workorder_cancel.action_work_order_cancel_confirm"
+        )
+        return action
```
dreispt: FYI this is redundant - a Button can directly call a Window Action, without the need of a model method in the middle.

mrp_workorder_cancel/models/mrp_workorder.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def button_workorder_cancel_undo(self):
+        work_order_ids = self.env["mrp.workorder"].browse(
+            self.env.context.get("active_ids")
+        )
+        for wo in work_order_ids:
```
dreispt: Can't you just use self?
```suggestion
        for wo in self:
```

mrp_workorder_cancel/views/mrp_workorder_view.xml
```diff
@@ -0,0 +1,36 @@
...
+            name="inherit_id"
+            ref="mrp.mrp_production_workorder_tree_editable_view"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//tree" position="attributes">
+                <attribute name="decoration-muted">state=='cancel'</attribute>
+            </xpath>
+            <button name="button_start" position="after">
+                <button
+                    name="button_cancel"
```
dreispt: This should call action_cancel and have `confirm="True"` to popup the confirmation on click.
The wizard is not really needed.

dreispt (commented):
Thank you, this looks good.
I believe the goal can be achieved with significantly less code.
Can you try simplifying it?

mrp_workorder_cancel/views/mrp_workorder_view.xml
Hardik-OSI: Hello @dreispt 
When we add Button in List View Odoo List View dont support confirm Attribute.
So we need to add Wizard.

PR #1000, opened 2023-03-14
[14.0][IMP] mrp_subcontracting_purchase_link: add post init hook to populate the PO line on previous manufacturing orders

JordiBForgeFlow:
Otherwise when you install the module the previous subcontracting MO's will not have the link with the purchase order
cc @ThiagoMForgeFlow

### DIFF
--- a/mrp_subcontracting_purchase_link/hooks.py
+++ b/mrp_subcontracting_purchase_link/hooks.py
@@ -1,5 +1,6 @@
 # Copyright 2023 ForgeFlow, S.L.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
 import logging
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


mrp_subcontracting_purchase_link/hooks.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2023 ForgeFlow, S.L.
+# License OPL-1
```
yostashiro: Should be LGPL, I suppose.

JordiBForgeFlow: @yostashiro You are right! Fixed ;)

yostashiro (approved):
Code review. LGTM. 👍

LoisRForgeFlow (approved):

LoisRForgeFlow:
Good morning ocabot, time to get up! :)

/ocabot merge minor

PR #904, opened 2022-11-22
[16][MIG] mrp_restrict_lot

florian-dacosta:
Standard migration
Depends on https://github.com/OCA/stock-logistics-workflow/pull/1137

PR #1010, opened 2023-03-31
[14.0][FIX] mrp_restrict_lot avoid duplicate mo name in case of cancellation qty changes ...

bguillot:
Hello,

In some cases odoo will create a second mo with the same name.

In that case the mo name constrains raise an error.
With this PR, I try to avoi that.

Thanks


florian-dacosta (approved):
Code review LGTM

If you can review/port this here it would be nice https://github.com/OCA/manufacture/pull/904

sebastienbeau (approved):
LGTM (code review and test)

PR #625, opened 2021-03-13
[14.0][MIG] mrp_production_request -> mrp_request: Migration to V14

alan196:
### Proposed changes

Migration PR for module mrp_production_request

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

### Further comments

The module was renamed to [mrp_request.](https://github.com/OCA/manufacture/issues/534#issuecomment-734850047)
I have changed the dependency from stock_available_unreserved to stock_free_qty as [commented.](https://github.com/OCA/stock-logistics-warehouse/issues/981#issuecomment-751362489)

PR #1018, opened 2023-04-21
[14.0][MIG] mrp_production_request

ypapouin:
This PR is an alternative to #625, I'm not against the refactoring proposal but I don't have time to refactor all modules that depends on this one at the moment. The refactoring could be done in a second PR.

- I've added more tests to ensure that the migration does not break the existing workflow.
- The list view has been updated to the 14.0 design
- A basic kanban view has been added.

image
image


elvise:
Hi @ypapouin thanks for this PR!
I tested it but looks doesn't work "trigger manufacture request" please check this [video](http://recordit.co/nRLySqalX6)

ypapouin:
@elvise, you are right. The problem happens when the `mrp_sale_info` module is installed.
It shows up that the current implementation was invalid (by removing fields from a static list).
An exception is raised when running the scheduler (but silenced because the scheduler run in parallel).

Its working now, but please note that since MTO rule is not enabled, you have to create a « minimum stock rule » and the product must have a BoM (even empty). (This restriction could be removed in another PR since a BoM is no more necessary for a production order)

elvise:
@ypapouin thanks for your work! 
i’m testing your PR and why not add an action in rule ?

IMG_1274

for complex operations it will be more scalable…

ypapouin:
@elvise, replacing the boolean option with a route should be possible but I would prefer having this module migrated first.

PR #1019, opened 2023-04-26
[13.0][BACKPORT] mrp_multi_level: Get BoM to explode

BernatPForgeFlow:
When exploding requirements or calculating supply method, we will consider the first active BoM taking into account the routes.

Origin: https://github.com/OCA/manufacture/pull/991


LoisRForgeFlow (approved):
LGTM

PR #1036, opened 2023-05-11
[14.0][FIX] mrp_attachment_mgmt: Change to the correct inherit view

victoralmau:
Change to the correct inherit view

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT43261


pedrobaeza (approved):
Fast-tracking as obvious problem:

/ocabot merge minor

PR #1038, opened 2023-05-16
[16.0][ADD] mrp_workcenter_cost

sebalix:
Controls how to compute the workcenter cost of a manufactured product (effective vs theoretical duration).


ivantodorovich (approved):

sebalix:
I cannot reproduce the failing tests locally, even with the last version of Odoo 16.0 :thinking:

mrp_workcenter_cost/tests/test_mrp_workcenter_cost.py
```diff
@@ -0,0 +1,75 @@
...
+
+from odoo.tests import Form, TransactionCase
+
+
+class TestMrp(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.bom = cls.env.ref("mrp.mrp_bom_table_top")
```
grindtildeath, line 13:
```suggestion
        cls.bom = cls.env.ref("mrp.mrp_bom_table_top")
        # Activate operations disabled by default in demo data
        cls.bom.with_context(active_test=False).operation_ids.active = True
```
Cf https://github.com/odoo/odoo/commit/aaba4eb4deb3f9e10fd611924c5a24680d633537

sebalix: Thank you! It's working now :+1:

TDu (approved):

grindtildeath (approved):

mrp_workcenter_cost/models/mrp_production.py
```diff
@@ -0,0 +1,28 @@
...
+        if should_overwrite_duration:
+            workorders = self.workorder_ids.filtered("duration_expected")
+            for workorder in workorders:
+                duration_by_workorder[workorder.id] = workorder.duration
+                workorder.duration = workorder.duration_expected
+        res = super()._cal_price(consumed_moves)
+        # Restore the durations set by users
+        if should_overwrite_duration:
+            for workorder in workorders:
+                workorder.duration = duration_by_workorder[workorder.id]
```
sebalix, line 27: Sadly we are breaking the product cost report by doing this:
- the stored cost is the one computed thanks to the duration set above
- but the product cost report is still relying on the times duration  recorded by the user (linked to the workorder)

This PR can't be merged as-is.

mrp_workcenter_cost/models/product_template.py
```diff
@@ -0,0 +1,19 @@
...
+
+    mrp_workcenter_cost = fields.Selection(
+        selection=[("theoretical", "Theoretical"), ("effective", "Effective")],
+        string="Workcenter Cost Duration",
+        help="Controls how to compute the workcenter cost for manufactured products.\n"
+        "* Theoretical: The cost is computed based on the theoretical duration of the "
+        "workcenter.\n"
+        "* Effective: The cost is computed based on the real duration of the workcenter.",
+        default="effective",
+        required=True,
```
grindtildeath, line 18: If we have only two possible values, would be great to avoid having a new field set as required because it can break CIs.
What about defining a Boolean with a default value matching the std Odoo?

grindtildeath:
@sebalix Can you please attend my last comment? We might also need a migration script though

PR #1042, opened 2023-05-25
[14.0][IMP] mrp_multi_level_estimate: Stock moves from MO should always be considered

BernatPForgeFlow:
When we consider forecasts as the 'Direct' demand, we should always keep considering the demand coming from Manufacturing Orders, as they are 'Indirect' demand.

### DIFF
--- a/mrp_multi_level_estimate/wizards/mrp_multi_level.py
+++ b/mrp_multi_level_estimate/wizards/mrp_multi_level.py
@@ -122,8 +122,11 @@ class MultiLevelMrp(models.TransientModel):
         )
         if direction == "out":
             mrp_date = res.get("mrp_date")
-            if self._exclude_considering_estimate_demand_and_other_sources_strat(
-                product_mrp_area, mrp_date
+            if (
+                self._exclude_considering_estimate_demand_and_other_sources_strat(
+                    product_mrp_area, mrp_date
+                )
+                and not res["production_id"]
             ):
                 return False
         return res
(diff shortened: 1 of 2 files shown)


mrp_multi_level_estimate/wizards/mrp_multi_level.py
```diff
@@ -122,8 +122,11 @@ def _prepare_mrp_move_data_from_stock_move(
         )
         if direction == "out":
             mrp_date = res.get("mrp_date")
-            if self._exclude_considering_estimate_demand_and_other_sources_strat(
-                product_mrp_area, mrp_date
+            if (
+                self._exclude_considering_estimate_demand_and_other_sources_strat(
+                    product_mrp_area, mrp_date
+                )
+                and not res["production_id"]
```
LoisRForgeFlow, line 129: ~I think it should be raw_material_production_id to refer to indirect demand.~ Please, add a test case too.

EDIT: this is actually correct because is the field of the mrp.move not the stock.move, my bad.

LoisRForgeFlow (changes requested):

gurneyalex (approved):

LoisRForgeFlow:
Hi @gurneyalex 

Thanks for reviewing! However, I would like to ask you to refrain for merging mrp_multi_level* related PRs that I have not explicitly approved, I would like to have control of the changes done and maintain it at my own pace.

I try to answer to all the pings that I receive, and if you find I don't answer a github mention, just drop me a direct email.

Hope this sounds reasonable to you :)

PR #1043, opened 2023-05-29
[14.0][IMP] mrp_component_operation: enhance extensibility of scrap vals

LoisRForgeFlow:
aditionally some adjustments have been done:
- create_date is a magic field, no need to include it in the vals.
- production_id is the inverse name of mo_id.scrap_ids.

cc @DavidJForgeFlow


DavidJForgeFlow (approved):
Thank you Lois, LGTM :+1:

PR #1054, opened 2023-06-13
[16.0][mig] mrp_production_grouped_by_product: Migration to 16.0

dsolanki-initos:


alihashaam:
@dsolanki-initos review is done

PR #1056, opened 2023-06-13
[15.0][IMP] quality_control_oca: inspection access buttons in product forms

dalonsod:
This improvement adds form buttons for product and product variants, so inspections for product variants can be directly accessed. Also, when an inspection is created within that form/tree views, where possible, product is auto-filled.


dalonsod:
@cubells @enriquemartin @angelmoya could you review?

When merged, I'll port it to v16

dalonsod:
@lk-eska @gurneyalex could you review? Thanks!

quality_control_oca/models/product_product.py
```diff
@@ -16,3 +18,62 @@ class ProductProduct(models.Model):
...
+        qc_type = self.env.context.get("qc_type", "all")
+        if qc_type == "done":
+            domain.append(("state", "not in", ["draft", "waiting"]))
+        elif qc_type == "passed":
+            domain.append(("state", "=", "success"))
+        elif qc_type == "failed":
+            domain.append(("state", "=", "failed"))
+        action["domain"] = domain
+        if len(product_ids) == 1:
+            context = ast.literal_eval(action.get("context"))
```
pedrobaeza: Please don't do literal eval. That's very counter-productive. `context` should be already a dictionary.

dalonsod: In my case it's not:

imagen

Am I doing something wrong?

pedrobaeza: The problem is that `literal_eval` will crash on a lot of cases. Isn't another way to do this? What about having an action including already such context?

dalonsod: Ok, I'll take a look

dalonsod: @pedrobaeza done, move to XML actions, so python action coding (and `ast.literal_eval`) is no longer required at this point. I've lost `default_object_id` management, but I think it's not a key feature.

dalonsod:
Tests seem to fail due to `stock_picking_product_kit_helper` addon. I've recently done a rebase, what could I do to solve it?

pedrobaeza:
It seems one module is doing `move_line_raw_ids` non stored, or there's a non-appropriate search on that field.

About the `default_object_id`, maybe you can use `active_id`.

dalonsod:
> It seems one module is doing `move_line_raw_ids` non stored, or there's a non-appropriate search on that field.

This test error is reproduced in #1149 , an addon with no tests. How could we solve it? I've looked for any search across `move_line_raw_ids` in other addons and I haven't found anything. This error should stop merges in this repo in general?

> About the `default_object_id`, maybe you can use `active_id`.

Yes, I tried it before updating this PR, but it didn't work. Thanks for the point anyway 😀

PR #1079, opened 2023-08-16
[ADD][16.0] module mrp_bom_configurable

FranzPoize:
This merges #1076 and #1077 into one module


mrp_bom_configurable/models/demo_input_mixin.py
```diff
@@ -0,0 +1,65 @@
+from odoo import api, fields, models
+
+
+class InputMixin(models.AbstractModel):
+    _name = "input.mixin"
+    _description = "Input Configuration mixin"
+
+    test_config = fields.Boolean(default=False, required=True)
```
bealdav: TODO (no emergency) : add a demo_view.xml with this field

mrp_bom_configurable/models/__init__.py
```diff
@@ -0,0 +1,10 @@
+from . import input_config
+from . import input_line
+from . import input_constraint
+from . import mrp_bom
+from . import mrp_bom_line
+
+from odoo.tools import config
+
+if not config["without_demo"] or config["test_enable"]:
```
rvalyi: we were using without_demo to test demo mode in several OCA/l10n-brazil modules and switch back to testing with a select demo from ir_module where... because it was more resilient to shity config values like empty vs False for without_demo as it was reported to us in several issues.

bealdav: Hi, do you have an example ?

rvalyi: @bealdav @FranzPoize sorry I missed your message. So one way we fixed it is by looking if some demo record is loaded, for instance:
https://github.com/OCA/l10n-brazil/commit/a79caf2d211ef7efb84f7d913c137116a5851b05

and another way is looking at the demo column in the ir_module_module table:
https://github.com/OCA/l10n-brazil/commit/eece8ca1b84040699ef5658cfcb3f087c2276425#diff-50fe6815141d1f784405627ca86d85e38989b6be8ccc48e479f51480b94e27b1R28

again, simply using config["without_demo"] has real issues, we had to fix all the modules using this approach.

quentinDupont:
I'm not sure to understand the usage of this module but i'm interested in testing it when it will be ready to do so :)

FranzPoize:
@quentinDupont Thanks we're still developing it and I'll keep you posted once it's ready

mrp_bom_configurable/models/__init__.py
bealdav: I confirm it's a bad idea to use without_demo key to guess if we can install a demo model.
We have to apply one of these alternate methods

bealdav:
Variable BOM

image

Quantity computation
image

Technical parameters
image

Configuration linked to sale
image

image

image

LoisRForgeFlow:
@FranzPoize is this ready to review?

FranzPoize:
Hi @LoisRForgeFlow ! No it's not we're still in a development and it may still change. It will probably be stable and ready in september.

Issue #1088, opened 2023-09-01
[15.0] mrp_account_bom_attribute_match: error in tests


ChristianSantamaria:
In PR #1048 the following error is occurring in Runboat, could you take a look at it @ivandodorovich?:
```
2023-09-01 06:21:38,598 252 ERROR odoo unittest.suite: ERROR: setUpClass (odoo.addons.mrp_bom_attribute_match.tests.test_mrp_bom_attribute_match.TestMrpBomAttributeMatch)
Traceback (most recent call last):
  File "/__w/manufacture/manufacture/mrp_bom_attribute_match/tests/common.py", line 142, in setUpClass
    cls.p3_bom_id = cls._create_bom(
  File "/__w/manufacture/manufacture/mrp_bom_attribute_match/tests/common.py", line 162, in _create_bom
    for vals in line_form_vals:
  File "/opt/odoo/odoo/tests/common.py", line 2185, in __exit__
    self.save()
  File "/opt/odoo/odoo/tests/common.py", line 2204, in save
    r = self._model.create(values)
  File "<decorator-gen-123>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 412, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/addons/mail/models/mail_thread.py", line 261, in create
    threads = super(MailThread, self).create(vals_list)
  File "<decorator-gen-67>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 413, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/odoo/addons/base/models/ir_fields.py", line 613, in create
    recs = super().create(vals_list)
  File "<decorator-gen-13>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 413, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/odoo/models.py", line 4088, in create
    records = self._create(data_list)
  File "/opt/odoo/odoo/models.py", line 4253, in _create
    records._validate_fields(name for data in data_list for name in data['stored'])
  File "/opt/odoo/odoo/models.py", line 1371, in _validate_fields
    check(self)
  File "/opt/odoo/addons/mrp/models/mrp_bom.py", line 152, in _check_bom_cycle
    _check_cycle(bom.bom_line_ids.product_id, finished_products)
  File "/opt/odoo/addons/mrp/models/mrp_bom.py", line 131, in _check_cycle
    _check_cycle(subcomponents, finished_products | component)
[...]

PR #1090, opened 2023-09-04
[15.0][FIX] mrp_bom_attribute_match: cyclic references in bom

ValentinVinagre:
Setup 'test common' is changed because have cyclic references in bom.
issue: https://github.com/OCA/manufacture/issues/1088

### DIFF
--- a/mrp_bom_attribute_match/tests/common.py
+++ b/mrp_bom_attribute_match/tests/common.py
@@ -143,7 +143,7 @@ class TestMrpBomAttributeMatchBase(TransactionCase):
             cls.p3,
             [
                 dict(
-                    product_id=cls.p1.product_variant_ids[0],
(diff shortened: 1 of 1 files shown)


ValentinVinagre:
@HaraldPanten

oca_dependencies.txt
```diff
@@ -1,3 +1,7 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
 
 stock-logistics-warehouse
+account-analytic
```
etobella: oca_dependancies are no longer necessary, as we should be using pip  for installing

ValentinVinagre:
@etobella now 😄

etobella (approved):
/ocabot merge patch

Fast tracking as it is failing

PR #1091, opened 2023-09-06
[16.0][MIG] mrp_unbuild_subcontracting: Migration to 16.0

PauBForgeFlow:


mariadforgeflow (approved):
LGTM!

BernatPForgeFlow (approved):
LGTM!

AungKoKoLin1997 (approved):
LGTM

rinaldifirdaus (approved):
LGTM!

rinaldifirdaus:
I did some test in runboat using a component with a route set as "Resupply Subcontractor on Order" and found that when we Return the receipt transfer, new resupply transfer from warehouse to subcontractor location and another subcontract manufacturing order are created. without do anything to the transfer and MO, then validate the return transfer, the Subcontract manufacture order will automatically done and the resupply transfer remain in ready state.

IMO, these scenarios result in the following issues:
1. The resupply transfer remains open until the user is aware and manually cancels the transfer.
2. The finished goods, which were unbuild upon validating the return transfer, are automatically reproduced and stored in the subcontractor location.  (I believe that with the unbuild function, the stock that should be in the subcontractor location is the component, not the finished goods.)

Expected result:
1. Once a return transfer is created, it should not generate a new resupply transfer and the subcontract MO.
2. Once the return transfer is validated, and the unbuild order is automatically completed, the finished product should be located in the Production location, and the component should be located in the Subcontractor location. (no need reproduction scenario at this stage)

Could you please review this @PauBForgeFlow ?
I also compared the behavior with its v14 in runboat, and in v14 the behavior is aligned with the expected results mentioned above.  Please CMIIW

BernatPForgeFlow:
This test will fail until https://github.com/OCA/manufacture/pull/1179 is merged.

Also I added this PR that allowed me to simplify the logic: https://github.com/odoo/odoo/pull/148262

ChrisOForgeFlow:
@PauBForgeFlow please include changes made by this PR https://github.com/OCA/manufacture/pull/1266

PR #1092, opened 2023-09-07
[15.0][ADD] mrp_production_force_bom

AungKoKoLin1997:
This module restricts adding components in MO if the bom_id is not set.

@qrtl


mrp_bom_enforcement/views/mrp_production_views.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">mrp.production</field>
+        <field name="inherit_id" ref="mrp.mrp_production_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='move_raw_ids']" position="attributes">
+                <attribute name="attrs">{
+                    'readonly': ['|', '|', ('bom_id', '=', False),
+                    ('state', '=', 'cancel'),
+                    '&amp;', ('state', '=', 'done'),
+                    ('is_locked', '=', True)]
+                }</attribute>
```
yostashiro: Not comfortable with this domain override.  If we are to do this, we may consider using base_view_inheritance_extension, however I actually prefer not to change anything on the view.  Can we achieve the goal by adding a constraint method instead?

kanda999 (approved):
Functional review. it was working as expected.

yostashiro:
@AungKoKoLin1997 Contrary to my previous advice, have you thought about extending `_get_consumption_issues()` of mrp.production?  I now think that the design will be more in line with the standard feature that way.

The questions that I have with the current code are:

- Is it a good design to force the checking even when the order state is still draft?
- Should we ignore the cases where there is a BoM selected for the order but the user adds some random products as components (not checked in the standard function).

AungKoKoLin1997:
I did some adjustments. I didn't use `_get_consumption_issues()` because this function is called after an MO is confirmed. So, we can't remove or add products if the error is raised. So, I extended the `action_confirm()` instead.

mrp_bom_enforcement/models/mrp_production.py
```diff
@@ -0,0 +1,43 @@
...
+                production.bom_id.mapped("bom_line_ids.product_id.id")
+            )
+            move_raw_product_ids = set(production.move_raw_ids.mapped("product_id.id"))
+            # Check for unexpected products
+            unexpected_products = move_raw_product_ids - bom_product_ids
+            if unexpected_products:
+                error_messages.append(
+                    _("For order '%s': Some components are not part of the BOM.")
+                    % production.name
+                )
```
yostashiro: Shall we split this part into a method for extendability.

mrp_bom_enforcement/models/mrp_production.py
```diff
@@ -0,0 +1,43 @@
...
+            # Check for missing products
+            missing_products = bom_product_ids - move_raw_product_ids
+            if missing_products:
+                error_messages.append(
+                    _(
+                        "For order '%s': Some components from the BOM are missing"
+                        " in the raw materials."
+                    )
+                    % production.name
+                )
```
yostashiro: Shall we split this part into a method for extendability.

mrp_bom_enforcement/models/mrp_production.py
```diff
@@ -0,0 +1,43 @@
...
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def action_confirm(self):
+        error_messages = []
+        for production in self:
```
yostashiro:
```suggestion
        for production in self:
            if not self._force_bom:
                return super().action_confirm()
```

For extendability, can we do like this upon adding the `_force_bom()` method which returns True/False based on a config parameter?

yostashiro (commented):
Name suggestion: mrp_production_force_bom

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,71 @@
...
+            error_messages.append(
+                _(
+                    "For order '%s': Some components from the BOM are missing"
+                    " in the raw materials."
+                )
+                % self.name
+            )
+        return error_messages
+
+    def _force_bom(self):
```
yostashiro:
```suggestion
    @api.model
    def _force_bom(self):
```

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,71 @@
...
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    def _check_unexpected_products(
+        self, move_raw_product_ids, bom_product_ids, error_messages
+    ):
+        unexpected_products = move_raw_product_ids - bom_product_ids
+        if unexpected_products:
+            error_messages.append(
+                _("For order '%s': Some components are not part of the BOM.")
```
yostashiro: To make the message slightly simpler. Please do the same for other messages.

```suggestion
                _("%s: Some components are not part of the BOM.")
```

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        if param == "1":
+            return True
+        return False
+
+    def action_confirm(self):
+        if not self._force_bom():
+            return super().action_confirm()
+        error_messages = []
+        for production in self:
```
yostashiro, line 75:
```suggestion
        for production in self:
            if production._skip_bom_check():
                continue
```
Can we do this to provide a hook point.

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,77 @@
...
+            .get_param("mrp_production_force_bom.force_bom")
+        )
+        if param == "1":
+            return True
+        return False
+
+    # Intend to inherit as necessary
+    @api.model
+    def _skip_bom_check(self):
+        return False
```
yostashiro:
```suggestion
    def _skip_bom_check(self):
    """Extend this method as necessary to control whether or not to check the components against BoM."""
        self.ensure_one()
        return False
```

yostashiro (changes requested):
@AungKoKoLin1997 Please follow up on the suggestion on `_force_bom()` as well.

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,82 @@
...
+        if param == "1":
+            return True
+        return False
+
+    @api.model
+    def _skip_bom_check(self):
+        """
+        Extend this method as necessary to control whether or not to check the components
+        against BoM.
+        """
```
yostashiro: `@api.model` should not be added here.
```suggestion
    def _skip_bom_check(self):
        """Extend this method as necessary to control whether or not to check the
        components against BoM.
        """
```

AungKoKoLin1997: Yes, you right. I confused.

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,80 @@
...
+            if not production.bom_id:
+                error_messages.append(
+                    _("'%s': You cannot proceed without setting a BOM.")
+                    % production.name
+                )
+                continue
+            bom_product_ids = set(
+                production.bom_id.mapped("bom_line_ids.product_id.id")
+            )
+            move_raw_product_ids = set(production.move_raw_ids.mapped("product_id.id"))
```
yostashiro: I suppose this does not prevent users from adding an existing component product in the MO as another line, does it? How should we handle this case? Check number of lines?

mrp_production_force_bom/models/mrp_production.py
```diff
@@ -0,0 +1,80 @@
...
+                continue
+            if not production.bom_id:
+                error_messages.append(
+                    _("'%s': You cannot proceed without setting a BOM.")
+                    % production.name
+                )
+                continue
+            bom_product_ids = set(
+                production.bom_id.mapped("bom_line_ids.product_id.id")
+            )
```
yostashiro: How about consideration on 'Apply on Variants' in the BoM line? I guess we should use `explode()` (or `_skip_bom_line()`) of mrp.bom as appropriate.

PR #1096, opened 2023-09-18
[16.0][IMP] mrp_bom_location: delete unused view

SergiCForgeFlow:


LoisRForgeFlow (changes requested):
It is needed, for the PDF report:

image

PR #1101, opened 2023-10-06
[13.0][IMP] quality_control_stock_oca: sudo the process of creating the test and calculated fields, so no permissions have to be assigned to the sotck user

CarlosRoca13:
cherry-picked from https://github.com/OCA/manufacture/commit/9eb29d6d678efa85460771496a4438e2aa41712f#diff-74fb8dfbe6ed5b9479f2438da6ae44c9f1a56a92e048c2ccb2012eec23c3e24d

cc @Tecnativa

ping @pedrobaeza @victoralmau

### DIFF
--- a/quality_control_stock_oca/models/stock_picking.py
+++ b/quality_control_stock_oca/models/stock_picking.py
@@ -54,9 +54,11 @@ class StockPicking(models.Model):
 
     def action_done(self):
         res = super().action_done()
-        inspection_model = self.env["qc.inspection"]
-        qc_trigger = self.env["qc.trigger"].search(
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
It's not strictly correct, as only the sudo in `action_done` is needed. The rest should be avoided restricting full views to the corresponding groups, but it's not a big problem and this is a relatively old version, so let's go this way.

victoralmau (approved):

pedrobaeza:
Uhm, there's seem to still be an error in the CI. Please check @CarlosRoca13

CarlosRoca13:
I have left only the changes related to the action_done method, now it should not fail.

PR #1102, opened 2023-10-09
[FIX][14.0] mrp_routing reuse sequence from mrp module

sergiocorato:


ChrisOForgeFlow (approved):
LGTM+

sergiocorato:
This change does not preserve the current sequence and customization, so I'll do it by hand.

PR #1104, opened 2023-10-13
[15.0][FIX] Fix edge case date bug in addon `mrp_production_grouped_by_product`

Sibert-Aerts:
Edge case: If `date` is on the last day of a month, `date.replace(day=date.day +1)` obviously fails.


LoisRForgeFlow (approved):
Thanks for the fix. LGTM

gurneyalex (approved):

gurneyalex (commented):
@Sibert-Aerts It seems that the fix breaks some tests (see the logs of oca-bot merge attempt). Can you check it?

PR #1109, opened 2023-10-16
[13.0][FIX] mrp_multi_level: Move LLC to a separate table

BernatPForgeFlow:


mrp_multi_level/models/mrp_product_llc.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+from odoo import fields, models
+
+
+class MrpProductLlc(models.Model):
+    _name = "mrp.product.llc"
+    _description = "MRP Product LLC"
+
+    product_id = fields.Many2one('product.product', "Product", required=True, ondelete='cascade')
```
JordiBForgeFlow, line 12: Please add index=True

mrp_multi_level/models/product_product.py
```diff
@@ -46,6 +47,14 @@ def write(self, values):
                 .search([("product_id", "in", self.ids)])
             )
             parameters.write({"active": False})
+        if not values.get("mrp_product_llc_id", True):
```
JordiBForgeFlow, line 50: What about the create method? When you create a product the mrp_product_llc table should also be updated, right?

mrp_multi_level/models/product_product.py
```diff
@@ -4,13 +4,14 @@
...
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class Product(models.Model):
     _inherit = "product.product"
 
-    llc = fields.Integer(string="Low Level Code", default=0, index=True)
+    llc = fields.Integer(related="mrp_product_llc_id.llc")
+    mrp_product_llc_id = fields.Many2one(comodel_name="mrp.product.llc")
```
JordiBForgeFlow, line 14: This is not a Many2one, but a One2many.

PR #1123, opened 2023-11-06
[15.0][IMP] mrp_multi_level: safety stock

gurneyalex:
When a product has a safety stock in an area, and the current stock is below safety, but there are moves in the future, mrp_multi_level does not compute an immediate action to get back to the safety stock.

This PR changes this behavior: when the mrp_moves are considered, before processing the first move in the future, we insert a resupply action to rebuild the safety stock.


LoisRForgeFlow:
@gurneyalex Sounds good, I will try to find some time to review this ASAP.

Thanks for the contribution!

SilvioC2C (approved):
LGTM

gurneyalex:
@LoisRForgeFlow I updated the PR with some improvements to handle "low ressource" period for a MRP area (maybe missing employees, broken equipment...) when we could need to consume the safety stock without rebuilding it. 

Using this is optional.

LoisRForgeFlow (commented):
Hi @gurneyalex 

Finally I'm reviewing this. The general idea of the prioritize safety stock is a good addition (as that it is very standard in other MRP implementations). 

With the secondary settings added, I'm more hesitant because they feel a bit ad-hoc, some thoughts:
- forces a specific date logic similar to a freezing/planning horizon but only for safety stock.
- Considers only weekly buckets planning.
- The date computed is the date to schedule the finish date, the trigger dates can be today if the LT is long enough. In other words, you are throwing procurement recommendations before that safety stock lead date.

The 2 first points are more comments than other thing, we can merge and let them evolve and improve if needed. For the 3rd one is more a warning and I want to know if that was intended, see below a couple of screenshots highlighting the problem:

image

image

gurneyalex:
@LoisRForgeFlow thanks for the review :) 

Point 3: Good catch, I forgot to take the lead time into account. I'll fix this. 

Point 1 and 2: yes this is a bit of a hack and comes straight from my current customer's requirements. They do give some flexibility by being at the product.mrp.area level. I had originally proposed to have a single date on the mrp.multi.level wizard screen, but the customer has a case where for instance one specific machine is unavailable for 2 weeks and this makes a set of products "harder to produce" during the period, so we chose to move things to the mrp.area which gives a finer control. 

If you are more at ease with a single date (not computed) on the mrp.area I can change this PR to go in this direction and then possibly change the logic of computing this date in a customer specific module. What do you think?

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -537,6 +537,29 @@ def _mrp_initialisation(self, mrp_areas):
...
+    def _get_qty_to_order(self, product_mrp_area, date, move_qty, onhand):
+        """compute the qty to order at a given date, for a product MRP area, given an
+        mrp.move quantity and an onhand quantity.
+
+        The method takes into account the safety stock priorization and safety
+        stock target date.
+        """
+        if (
+            product_mrp_area.mrp_area_id.priorize_safety_stock
+            and date < product_mrp_area.mrp_area_id.safety_stock_target_date
```
SilvioC2C:
```suggestion
            product_mrp_area.mrp_area_id.priorize_safety_stock
            and date
            and product_mrp_area.mrp_area_id.safety_stock_target_date
            and date < product_mrp_area.mrp_area_id.safety_stock_target_date
```

Will fix: https://github.com/OCA/manufacture/actions/runs/7100988379/job/19328179577?pr=1123#step:8:587

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -615,6 +642,103 @@ def _init_mrp_move_grouped_demand(self, nbr_create, product_mrp_area):
...
+            # if we are after the safety stock target date, and the forecast qty is
+            # going to be below the minimum, then we must
+            # resupply today to rebuild the safety stock.
+            #
+            # This is only done if the area has the
+            # priorize_safety_stock flag set
+            if (
+                product_mrp_area.mrp_area_id.priorize_safety_stock
+                and move.mrp_date
+                > product_mrp_area.mrp_area_id.safety_stock_target_date
```
SilvioC2C:
```suggestion
                and move.mrp_date
                and product_mrp_area.mrp_area_id.safety_stock_target_date
                and move.mrp_date > product_mrp_area.mrp_area_id.safety_stock_target_date
```

SilvioC2C (approved):
LGTM, tests are red because of failing dates comparison (some of them are not properly set so we get `date <operator> bool` which triggers a `TypeError`).

I'm suggesting changes but they may not be functionally ok (maybe we want to handle cases differently if one of the dates is not set?).

LoisRForgeFlow:
> @LoisRForgeFlow thanks for the review :)
> 
> Point 3: Good catch, I forgot to take the lead time into account. I'll fix this.
> 
> Point 1 and 2: yes this is a bit of a hack and comes straight from my current customer's requirements. They do give some flexibility by being at the product.mrp.area level. I had originally proposed to have a single date on the mrp.multi.level wizard screen, but the customer has a case where for instance one specific machine is unavailable for 2 weeks and this makes a set of products "harder to produce" during the period, so we chose to move things to the mrp.area which gives a finer control.
> 
> If you are more at ease with a single date (not computed) on the mrp.area I can change this PR to go in this direction and then possibly change the logic of computing this date in a customer specific module. What do you think?

Since it is optinal to use it, I'm fine with it as it is now. If you just fix point 3, I think we could move forward

LoisRForgeFlow:
@gurneyalex Also as a reminder, please do not merge PRs of mrp_multi_level, let us control the merges. I hope you understand, I might have approved a PR but maybe give a final thought before merging or maybe @JordiBForgeFlow is involved in a final review too.

gurneyalex:
> please do not merge PRs of mrp_multi_leve

Understood. Sorry about this.

gurneyalex:
> if you just fix point 3

It should be fixed. I'm adding some tests for this

JordiBForgeFlow:
@gurneyalex I am confused about this feature. It's not the same as what is in the description of the PR right? You are introducing 2 features:

1. Rebuild now safety if stock level requires it -> I believe that this solves a bug. I propose that you remove the boolean and set it as the default behaviour.

2. Rebuild safety stock on a future date -> does not make sense to me, at least as part of the main module. Seems to me more like a customization. Besides, adding a specific date in the MRP area does not make sense. Because this would apply to all products in the area.

I propose that you introduce whatever hooks you need in this module that can help you to create an additional module that implements this feature.

gurneyalex:
@JordiBForgeFlow Thanks for the feedback. I'll update the PR.

gurneyalex:
The PR was updated. The code I split for the safety stock rebuilding is available on #1169

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -538,6 +538,16 @@ def _mrp_initialisation(self, mrp_areas):
                 self._init_mrp_move(product_mrp_area)
         logger.info("End MRP initialisation")
 
+    def _get_qty_to_order(self, product_mrp_area, date, move_qty, onhand):
+        """compute the qty to order at a given date, for a product MRP area, given an
+        mrp.move quantity and an onhand quantity.
+
+        The method takes into account the safety stock priorization and safety
```
JordiBForgeFlow: This text is not relevant any more, right?

JordiBForgeFlow (approved):
:+1: Code review. Thanks for the fix!

mrp_multi_level/wizards/mrp_multi_level.py
gurneyalex: you're right. Fixed in d45721609ced96f652d83f02ec6c53d37e36b4c9

Happy new year to the ForgeFlow team :grin:

PR #1130, opened 2023-11-15
[15.0][IMP] mrp_multi_level: access action

AaronHForgeFlow:
Steps to reproduce. User with access to stock and manufacture goes to a product and clicks on the MRP Areas smart button. Result:
image

cc @ForgeFlow

### DIFF
--- a/mrp_multi_level/models/product_product.py
+++ b/mrp_multi_level/models/product_product.py
@@ -50,8 +50,9 @@ class Product(models.Model):
 
     def action_view_mrp_area_parameters(self):
         self.ensure_one()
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow (approved):

LoisRForgeFlow (changes requested):
Can you use `action = self.env["ir.actions.actions"]._for_xml_id(` instead to avoid the use of sudo?

AaronHForgeFlow:
> Can you use `action = self.env["ir.actions.actions"]._for_xml_id(` instead to avoid the use of sudo?

@LoisRForgeFlow  done that. Thanks!

LoisRForgeFlow (approved):
:+1:

PR #1137, opened 2023-11-21
[14.0][FIX] repair_sale_order, keep shipping address 

ChrisOForgeFlow:
Keep shipping address  after trigger partner onchange on sale order creation

@ForgeFlow

### DIFF
--- a/repair_sale_order/models/repair_order.py
+++ b/repair_sale_order/models/repair_order.py
@@ -62,7 +62,12 @@ class RepairOrder(models.Model):
             sale_order_data = rec._get_sale_order_data()
             sale_order = order_model.create(sale_order_data)
             orders |= sale_order
+            partner_shipping_id = False
+            if sale_order.partner_shipping_id != sale_order.partner_id:
+                partner_shipping_id = sale_order.partner_shipping_id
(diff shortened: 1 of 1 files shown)


repair_sale_order/models/repair_order.py
```diff
@@ -62,7 +62,12 @@ def action_create_sale_order(self):
             sale_order_data = rec._get_sale_order_data()
             sale_order = order_model.create(sale_order_data)
             orders |= sale_order
+            partner_shipping_id = False
```
MiquelRForgeFlow, line 66: why not do the same with partner_invoice_id?

ChrisOForgeFlow: You're right, I added same behavior, thank you

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM

mariadforgeflow (approved):
LGTM!

grindtildeath:
While working on this module, would be good if you can add a test there, because the existing one doesn't bring anything: https://github.com/OCA/manufacture/blob/14.0/repair_sale_order/tests/test_repair_sale_order.py#L12

PR #1143, opened 2023-11-27
[13.0][FIX] mrp_multi_level: Correct main supplier when changing sequence in Vendor Pricelist

BernatPForgeFlow:
When we change the sequence order of the Vendor Pricelist for a product, we firstly fetch the vendors and then we change the sequence. Then, when changing a sequence we try to get the first supplier but we do not get the desired result. Reordering the suppliers allows to always get the first supplier in the list.

### DIFF
--- a/mrp_multi_level/models/product_mrp_area.py
+++ b/mrp_multi_level/models/product_mrp_area.py
@@ -219,7 +219,7 @@ class ProductMRPArea(models.Model):
             suppliers = rec.product_id.seller_ids.filtered(
                 lambda r: (not r.product_id or r.product_id == rec.product_id)
                 and (not r.company_id or r.company_id == rec.company_id)
-            )
+            ).sorted(lambda s: (s.sequence, -s.min_qty, s.price, s.id))
             if not suppliers:
                 rec.main_supplierinfo_id = False
                 rec.main_supplier_id = False


JordiBForgeFlow:
@BernatPForgeFlow But the table 'product.supplierinfo' is already sorted by https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L737. So when accessing to product.seller_ids, the first item should be the one with minimum sequence.

On the other side I found that the current code may not be selecting the correct supplierinfo record because it is not filtering the seller_ids by company or checking if there's a variant specific record. I think that you should use the same as:
https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L485, up to https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L492

BernatPForgeFlow:
> @BernatPForgeFlow But the table 'product.supplierinfo' is already sorted by https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L737. So when accessing to product.seller_ids, the first item should be the one with minimum sequence.

I think the problem is that the search is done first, in which the seller_ids are ordered by sequence and stored in memory in python objects. The sequences of the objects are then changed and written to memory, but at no point is the database looked up again, so they are no longer ordered. Therefore, once filtered by company I believe that yes they should be reordered.

It is also possible that I am wrong, but I am attaching a video that shows that when debugging the seller_ids are not ordered by sequence.

https://github.com/OCA/manufacture/assets/90243017/2b252507-cdc5-43e1-a95d-a9fff7dad3eb

> On the other side I found that the current code may not be selecting the correct supplierinfo record because it is not filtering the seller_ids by company or checking if there's a variant specific record. I think that you should use the same as: https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L485, up to https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L492

I think it is filtered by company in this part of the code:
https://github.com/OCA/manufacture/blob/13.0/mrp_multi_level/models/product_mrp_area.py#L219
```
suppliers = rec.product_id.seller_ids.filtered(
                lambda r: (not r.product_id or r.product_id == rec.product_id)
                and (not r.company_id or r.company_id == rec.company_id)
            )
```

JordiBForgeFlow:
@BernatPForgeFlow what about using similar to https://github.com/odoo/odoo/blob/13.0/addons/sale_purchase/models/sale_order.py#L265C1-L265C15?

BernatPForgeFlow:
> @BernatPForgeFlow what about using similar to https://github.com/odoo/odoo/blob/13.0/addons/sale_purchase/models/sale_order.py#L265C1-L265C15?

@JordiBForgeFlow Inspired by the base code you gave me, I changed the 'sorted' method to a 'sorted' method from Odoo core, as I didn't remember that is also available. 

I don't use _select_seller for two reasons:
1. It does a database search and is therefore slower.
2. If the supplier in the vendor pricelist does not have a quantity of 0, it will never be selected.

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1147, opened 2023-11-30
[16.0][PORT] mrp_multi_level: Correct main supplier when changing sequence in Vendor Pricelist

BernatPForgeFlow:
Forward Port from: https://github.com/OCA/manufacture/pull/1143

When we change the sequence order of the Vendor Pricelist for a product, we firstly fetch the vendors and then we change the sequence. Then, when changing a sequence we try to get the first supplier but we do not get the desired result. Reordering the suppliers allows to always get the first supplier in the list.


LoisRForgeFlow (commented):
/ocabot merge patch

PR #1148, opened 2023-11-30
[14.0][FIX] mrp_sale_info direct procurement without move dests

sergiocorato:
How to replicate:

1. create a product with a bom and produce route
2. replenish the product
3. an error appears as `move_dest_ids` is False


thomaspaulb:
Can be closed, merged already in https://github.com/OCA/manufacture/pull/1189

PR #1150, opened 2023-12-04
[15.0][REF] mrp_multi_level: location management

gurneyalex:
Small refactoring adding a `_get_locations` method on `product.mrp.area` which by defaults delegates the computation to the related `mrp.area`.

This enables extending a few things related to locations at the `product.mrp.area` level (so on a per-product basis)

### DIFF
--- a/mrp_multi_level/models/mrp_area.py
+++ b/mrp_multi_level/models/mrp_area.py
@@ -39,6 +39,4 @@ class MrpArea(models.Model):
 
     def _get_locations(self):
         self.ensure_one()
-        return self.env["stock.location"].search(
-            [("id", "child_of", self.location_id.id)]
-        )
+        return self.location_id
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow:
@gurneyalex LGTM. Same as in the other, we have failing tests, probably unrelated.

gurneyalex:
@LoisRForgeFlow I opened https://github.com/OCA/manufacture/issues/1152 for the failing tests

gurneyalex:
I'm wondering about https://github.com/OCA/manufacture/pull/1150/commits/965b3d03f1f489f9bb067370ae2f2e12fba81d2f (I put it as a separate commit here to see if it changes the test results). 

@LoisRForgeFlow  Was there a need to use list of ids instead of the child_of operator in the first place? For large warehouses using many sublocations, I think this change will improve performance.

SilvioC2C (approved):
LGTM

LoisRForgeFlow:
> I'm wondering about [965b3d0](https://github.com/OCA/manufacture/commit/965b3d03f1f489f9bb067370ae2f2e12fba81d2f) (I put it as a separate commit here to see if it changes the test results).
> 
> @LoisRForgeFlow Was there a need to use list of ids instead of the child_of operator in the first place? For large warehouses using many sublocations, I think this change will improve performance.

Not really, probably unchanged from the first version until now.

LoisRForgeFlow (approved):
LGTM

/ocabot merge patch

PR #1151, opened 2023-12-04
[15.0][IMP] mrp_multi_level: reduce some logging

gurneyalex:
When running the multi level planner, the wizard will delete the records from the previous run (`mrp.inventory`, `mrp.move`, `mrp.planned.order`). There can be hundreds of such records, and the call to `unlink()` will create a log entry which is not really informative. We disable this log during the call to the wizard (but keep if for manual unlinking of the records which we want to trace).

### DIFF
--- a/mrp_multi_level/wizards/mrp_multi_level.py
+++ b/mrp_multi_level/wizards/mrp_multi_level.py
@@ -8,7 +8,7 @@ import logging
 from datetime import date, timedelta
 
 from odoo import _, api, exceptions, fields, models
-from odoo.tools import float_is_zero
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):

JordiBForgeFlow:
@gurneyalex Thanks. Looks good.

LoisRForgeFlow:
@gurneyalex LGTM, but test are failing, unrelated?

gurneyalex:
@LoisRForgeFlow Tests are broken on the 15.0 branch, see https://github.com/OCA/manufacture/issues/1152

SilvioC2C (approved):

PR #1153, opened 2023-12-06
[14.0][ADD] mrp_bom_attribute_match_duplicate_kit

geomer198:
This module allows to duplicate a product with variants, at the same time creating a BoM with type "Kit" for the new product mapping variants of the original product.

This is useful when managing products that only exist as a "front" for a product with another name, to avoid having to manually map variants in BOM.


angelinaanaki (approved):
LGTM!

francesco-ooops (approved):
Functional ok!

mrp_bom_attribute_match_duplicate_kit/tests/common.py
```diff
@@ -0,0 +1,62 @@
...
+            "product.product_attribute_1", raise_if_not_found=False
+        )
+
+        # Attribute Legs Values
+        cls.attribute_legs_steel = cls.env.ref(
+            "product.product_attribute_value_1", raise_if_not_found=False
+        )
+        cls.attribute_legs_aluminium = cls.env.ref(
+            "product.product_attribute_value_2", raise_if_not_found=False
+        )
```
HekkiMelody: **quibble(non-blocking)**: I don't think we should have `raise_if_not_found=False` here because if they are not found, you'll have `False` and then the tests would (or should) fail, and it'll be harder to troubleshoot why. If we simply have the error raised immediately when doing `cls.env.ref()` it's easier to realize that maybe the XML ID of an attribute changed.

mrp_bom_attribute_match_duplicate_kit/tests/test_product_template.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual(
+            context["default_new_product_template_name"],
+            product.name + " 1",
+            "Default New product template name must be equal to '{} 1'".format(
+                product.name
+            ),
+        )
+
+
+# product_sword
```
HekkiMelody: **todo**: cleanup leftover line

HekkiMelody (changes requested):
Code review, a couple of really minor points, but generally LGTM

mrp_bom_attribute_match_duplicate_kit/tests/common.py
geomer198: @aleuffre  Agree with you!

GabbasovDinar (approved):

HekkiMelody (approved):

francesco-ooops:
@pedrobaeza good to go?

pedrobaeza:
Blindly merging

/ocabot merge nobump

PR #1170, opened 2023-12-22
[ADD] mrp_packaging_default: product packaging data in MRP

yajo:
This module allows and encourages the use of packaging within MRP, both to manufacture products or to create kits.

@moduon MT-4506


fcvalgar (approved):
Fuunctional review

https://www.loom.com/share/db31ae838dbf410ea2269a59ad044446?sid=e0a7841d-71a5-485e-98ea-1d1ac1bb8c30

LGTM,
thank you @yajo

rafaelbn (approved):
Functional test in runboat! 👍🏼 

imagen

Gelojr (approved):
LGTM Thanks @yajo https://www.loom.com/share/3d6b3c14f04c477b93626803810c7838?sid=d88bb856-ff31-42bd-87d3-457c7da5f496

rafaelbn:
Any @OCA/manufacturing-maintainers could please review here? @rousseldenis ? 😄 Happy new year! ❤️

yajo:
@LoisRForgeFlow this one seems ready to merge, do you think you could hit the button please? 🙏🏼

mrp_packaging_default/models/stock_move.py
```diff
@@ -0,0 +1,36 @@
...
+            bom_line = self.env["mrp.bom.line"].browse(vals["bom_line_id"])
+        except KeyError:
+            # No BoM line, nothing to do
+            return
+        vals.update(
+            {
+                "product_packaging_id": bom_line.product_packaging_id.id,
+                "product_packaging_qty": bom_line.product_packaging_qty,
+            }
+        )
```
grindtildeath, line 24:
```suggestion
        """Fill vals with packaging info from BoM line."""
        bom_line_id = vals.get("bom_line_id")
        if bom_line_id:
            bom_line = self.env["mrp.bom.line"].browse(vals["bom_line_id"])
            vals.update(
                {
                    "product_packaging_id": bom_line.product_packaging_id.id,
                    "product_packaging_qty": bom_line.product_packaging_qty,
                }
            )
```
Using an if is usually nicer than a try except block :wink:

mrp_packaging_default/models/stock_move.py
```diff
@@ -0,0 +1,36 @@
...
+    def create(self, vals_list):
+        """Inherit packaging from BoM line."""
+        for vals in vals_list:
+            self._packaging_vals_from_bom_line(vals)
+        return super().create(vals_list)
+
+    def write(self, vals):
+        """Inherit packaging from BoM line."""
+        self._packaging_vals_from_bom_line(vals)
+        return super().write(vals)
```
grindtildeath, line 36: On this I'm wondering, do we necessarily want to force the packaging on all the moves related to a BOM if module is installed? Shouldn't we have an option on the BOM to activate/deactivate the behaviour?

grindtildeath (commented):
Looks good overall, just two small things

mrp_packaging_default/models/stock_move.py
yajo: This is the classical EAFP vs LBYL battle. People coming from langs such as C are more used to LBYL style. However, Python docs encourage EAFP. Thus, I'll leave as it was, because the end result is just the same but EAFP is generally considered more pythonic.

See https://stackoverflow.com/a/11360880/1468388 for a brief explanation.

Thanks for the review anyway! 🙂

mrp_packaging_default/models/stock_move.py
yajo: If you don't want packaging info to be transferred from BoM lines to stock moves, then you should either:
- Don't enable packagings in your DB.
- Don't install this module.
- Don't add packaging info in the BoM line.

At first sight, I don't see a real world use case for that option, so for starting, I prefer to have a simpler MVP implementation.

mrp_packaging_default/models/stock_move.py
grindtildeath: Thanks for the pointer, didn't know about this. Anyway I found [this article](https://realpython.com/python-lbyl-vs-eafp/) more telling than the stackoverflow thread, so linking it here FTR.

mrp_packaging_default/models/stock_move.py
grindtildeath: As your module is Alpha, that's fine as long as it's not messing with CI (which seems to be the case).

PR #1172, opened 2023-12-27
[14.0] [FIX] mrp_bom_attribute_match: display proper BoM structure

renda-dev:
Before this fix the BoM structure was wrong, for example:

A BoM setup like this:
immagine

Would have always been showing this structure:
immagine

While instead should be showing the proper structure of the BoM:
immagine

Only when the right combination of product variant is chosen.


francesco-ooops (approved):
Functional ok!

@ilyasProgrammer @ivantodorovich what do you think?

ilyasProgrammer (approved):
LGTM

francesco-ooops:
@ilyasProgrammer can you merge then? :)

ivantodorovich:
Thanks!


/ocabot merge patch

PR #1173, opened 2023-12-27
[15.0][FIX] Correct api.depends to avoid relying on non-stored computed field

### LINKED ISSUE #1156: [15.0] mrp_subcontracting_lot_number_propagation: Non-stored field mrp.production.move_line_raw_ids cannot be searched
Module mrp_subcontracting_lot_number_propagation on 15.0 has a bug with the computation _compute_propagated_lot_producing : the new dependencies involve mrp.production::move_line_raw_ids  which is not stored and cannot be searched. 

In that case, osv.expression will replace the search element by a True node (1 = 1) which likely causes a recomputation of the field on all mrp.production each time a stock.move.line is updated, which can cause performance issues.

grindtildeath:
Closes: #1156


bosd (approved):
Very quick code review, LGTM

PR #1176, opened 2024-01-09
[14.0] [FIX] mrp_bom_attribute_match_duplicate_kit: wrong UOM

renda-dev:
When creating a duplicate with attribute match kit, the new bom line UOM is set as default 'unit' while it should be product's UOM.

Also added a group to manage access to the wizard.


francesco-ooops (approved):
LGTM!

HekkiMelody (approved):
Code review, LGTM

PR #1180, opened 2024-01-18
[14.0] Add module mrp_repair_reuse_component

grindtildeath:
This module allows to bypass the limitation of Odoo, where a component
that was added and removed from a finished product through repair orders,
and built in a finished product using manufacturing order, before being
removed or unbuilt, cannot be used again in a manufacturing order.


ajaniszewska-dev (approved):
LG, pre-commit to fix

sebalix (approved):

PR #1185, opened 2024-01-24
[ADD] mrp_bom_line_formula_quantity

SirAionTech:


hbrunn (approved):
very useful.

 I just played with this, and

```python
quantity=int(production.move_dest_ids.mapped('sale_line_id.product_custom_attribute_value_ids').filtered(lambda x: x.custom_product_template_attribute_value_id.name == 'Height').custom_value or 0)
```

is a little bit unwieldy to use in formulas, but out of scope here and probably best addressed by a module depending on this one and `sale_mrp`

quentinDupont:
Hello !
Is it possible to have some screenshots in README and some use cases ?
Thanks for the work !

SirAionTech:
> Hello ! Is it possible to have some screenshots in README and some use cases ? Thanks for the work !

Hi! There you go, could you please add a review if you find this module useful? Have a look at https://odoo-community.org/resources/review if you want to know how to do that.

Note that the images will be visible only when the PR is merged because they explicitly point to OCA's `16.0` branch.
If you want to see them now, you have to make them address this PR's branch, for example this non-working image link
> https://raw.githubusercontent.com/OCA/manufacture/16.0/mrp_bom_line_formula_quantity/static/description/images/product_attribute.png

would become

> https://raw.githubusercontent.com/siraiontech/manufacture/16.0-add-mrp_bom_line_compute_quantity/mrp_bom_line_formula_quantity/static/description/images/product_attribute.png

quentinDupont:
Thanks, yes I know I can review. My request was about the fact that putting visuals allows to be more inclusive and that non-technical (and technical) people understand more easily the module and can test it in the runboat without going into the code. 
I think you could "copy and paste" your module description into the PR description to facilitate this and make the images work directly.

Having said that, I tried to review your PR but I don't understand. Maybe I'm not concerned enough with the processes you're improving. I'm not sure whether the example you give in the README is an example or the case you're trying to improve.
If it's not just a special case, I think the module name should refer more to website_sale.
Also, it seems strange to me to have to write 6 lines of code in a non-technical place (the form view of a BoM) for it to work.

Thanks for the work

SirAionTech:
> Thanks, yes I know I can review. My request was about the fact that putting visuals allows to be more inclusive and that non-technical (and technical) people understand more easily the module and can test it in the runboat without going into the code. I think you could "copy and paste" your module description into the PR description to facilitate this and make the images work directly.

I'd rather leave the PR description as is so I don't have to duplicate the README content (that means also updating both when needed); the PR name is catchy enough in my opinion and the README is the first thing in the _Files changed_ tab and can be easily read from there if someone wants to know more.

> Having said that, I tried to review your PR but I don't understand. Maybe I'm not concerned enough with the processes you're improving. I'm not sure whether the example you give in the README is an example or the case you're trying to improve. If it's not just a special case, I think the module name should refer more to website_sale.

Thanks for trying :smile: the shown use-case is not the easiest one but is very similar to what we use this module for.
There are probably other easier use-cases where additional dependencies are not needed, but I don't know them.

> Also, it seems strange to me to have to write 6 lines of code in a non-technical place (the form view of a BoM) for it to work.

Allowing the user to insert Python code is the most versatile way I found to implement the solution for my use-case; I know it is technical but it should be a one-shot configuration that is done from the technical guy.

mrp_bom_line_formula_quantity/tests/test_mrp_bom_line.py
```diff
@@ -0,0 +1,23 @@
+#  Copyright 2024 Simone Rubino - Aion Tech
+#  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.fields import first
+
+from odoo.addons.mrp.tests.common import TestMrpCommon
+
+
+class TestMRPBoMLime(TestMrpCommon):
```
MarcinSokolowski-Solvti: Typo
```suggestion
class TestMRPBoMLine(TestMrpCommon):
```

MarcinSokolowski-Solvti (approved):
Approved with typo correction suggestion

mrp_bom_line_formula_quantity/tests/test_mrp_bom_line.py
SirAionTech: :lemon: hah! Fixed it, much appreciated

PR #1187, opened 2024-01-24
[15.0][IMP] mrp_multi_level: add date to default grouping filters

LoisRForgeFlow:
It is the default column for pivot view.


JoanSForgeFlow (approved):
LGTM! :+1:

PR #1189, opened 2024-01-25
[FIX] mrp_sale_info: check moves

adrip-s73:
This fix is to solve the next error: 

image


ferran-S73 (approved):

francesco-ooops (approved):
Functional ok (it's the same solution as proposed independently in https://github.com/OCA/manufacture/pull/1215 ) 

@pedrobaeza could this be merged? thanks!

PR #1191, opened 2024-01-26
[15.0][FIX] mrp_multi_level: variable 'move' referenced before assignment

LoisRForgeFlow:
Small fix to https://github.com/OCA/manufacture/pull/1123.

PR #1192, opened 2024-01-29
[FIX] mrp_multi_level: UnboundLocalError

gurneyalex:
Fix a copy-paste error in the call to _get_qty_to_order which was not using the correct mrp_date. Using the mrp_date of the move outside the loop would generate an UnboundLocalError exception if the there was no mrp_move_ids in the mrp.area.


gurneyalex:
I'm checking the failing test.

gurneyalex:
Looks like the red CI is a transient bug as I don't get the issue locally, and restarting the jobs gets me a green build. Maybe related to day of week ?

LoisRForgeFlow:
Hi @gurneyalex this one is duplicated with https://github.com/OCA/manufacture/pull/1191. And we have the same issue with CI there

PR #1199, opened 2024-02-06
[15.0][FIX] mrp_multi_level: wrong safety stock intial order in grouped demand mode

LoisRForgeFlow:
Three tests cases modelize the issues being solved.

Fix for https://github.com/OCA/manufacture/pull/1123.

PR #1202, opened 2024-02-14
[13.0][BACKPORT] mrp_multi_level:  safety stock

BernatPForgeFlow:
When a product has a safety stock in an area, and the current stock is
below safety, but there are moves in the future, mrp_multi_level does
not compute an immediate action to get back to the safety stock.

This PR changes this behavior: when the mrp_moves are considered, before
processing the first move in the future, we insert a resupply action to
rebuild the safety stock.

We also add some refactoring in the process:

* add extension point on the wizard to compute the quantity to reorder
(so we can choose in a separate module whether to rebuild safety stock
or not)
* add extension point on the wizard to get the date at which the safety
stock must be rebuilt (defaults to today)
* make the code of the wizard symetric between the groupes and non
grouped configuration

Sources: https://github.com/OCA/manufacture/pull/1123, https://github.com/OCA/manufacture/pull/1191, https://github.com/OCA/manufacture/pull/1199

Other sources: https://github.com/OCA/manufacture/pull/1183, https://github.com/OCA/manufacture/pull/1187

Apart from the backport, I add a commit that fixes a case in which we know if we should prioritize safety stock having mrp moves today.


LoisRForgeFlow (approved):
All good, including the additional fix that need to be forward ported

PR #1209, opened 2024-02-20
[16.0][MIG] mrp_subcontracting_bom_dual_use: Migration to 16.0

victoralmau:
Migration to 16.0

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT47843


pedrobaeza (approved):

chienandalu (approved):
Tested :+1:

giarve:
As using this module is more likely to happen, you might want to check this PR

https://github.com/odoo/odoo/pull/216518

and patch accordingly. Or just wait until it happens and then apply the resolution (leaving it here to help if you ever find the same error).

TBH I did not find the bug using this module, but it is likely to happen when using this because the BoM with operations creates workorders on the subcontractor MO.

pedrobaeza:
Thanks for the warning, @giarve maybe you can propose the PR?

giarve:
> Thanks for the warning, @giarve maybe you can propose the PR?

is it unlikely to be merged upstream? I opened it in Odoo main repo directly as it is not only related to this module but for anyone that uses subcontracting with a bom that was used for internal manufacturing before

pedrobaeza:
OK, let's wait if they merge it.

PR #1213, opened 2024-02-23
[16.0][ADD] mrp_mto_analytic

AungKoKoLin1997:
This module transfers the analyic_distribution from the source MO to a newly created MO that is generated based on the stock rule.

Related: https://github.com/OCA/account-analytic/pull/536

@qrtl


mrp_mto_analytic/tests/test_mrp_mto_analytic.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        # Confirm MO to trigger child MO creation
+        mo.action_confirm()
+        # Find the child MO for the second product
+        child_mo = mo._get_children()
+        # Test if the analytic_distribution of MO and child MO are the same
+        self.assertEqual(mo.analytic_distribution, self.analytic_distribution_manual)
+        self.assertEqual(
+            mo.analytic_distribution,
+            child_mo.analytic_distribution,
```
yostashiro:
```suggestion
        self.assertEqual(
            child_mo.analytic_distribution,
            self.analytic_distribution_manual,
```
**nitpick (non-blocking):** Put the value to be tested first.

mrp_mto_analytic/models/stock_rule.py
```diff
@@ -0,0 +1,38 @@
...
+            origin,
+            company_id,
+            values,
+            bom,
+        )
+        move_dest_ids = values.get("move_dest_ids")
+        if move_dest_ids:
+            vals["analytic_distribution"] = move_dest_ids[
+                0
+            ].raw_material_production_id.analytic_distribution
```
yostashiro, line 37:
```suggestion
            vals["analytic_distribution"] = move_dest_ids[0].analytic_distribution
```
Wouldn't it technically make more sense to do this, and switch the dependency to `mrp` and `stock_analytic`, removing `mrp_stock_analytic`?

AungKoKoLin1997: The feature of this module will not work without `mrp_stock_analytic`. I mean `stock.move` that comes from MO will not have values for `analytic_distribution`. IMO, we should not remove `mrp_stock_analytic` from the dependency.

yostashiro: My previous comment was invalid as we need `mrp_stock_analytic` to have `analytic_distribution` in `mrp.production`.

PR #1214, opened 2024-02-26
[16.0][ADD] mrp_bom_line_net_qty

quentinDupont:
**The idea:** Handle gross and net quantity on BoM lines thanks to loss percentage.

**Why use this module:**  Useful in the food field, and you use BoMs as recipes, and need to manage the loss of an ingredient during the first preparation (e.g.: grating a vegetable)

**How to use:**
Fill quantity (original one is 'gross quantity') : it set automatically 'net quantity'
Fill loss percentage, Odoo will show two buttons to calculate net or 'gross quantity'
You can also set 'net quantity' and calculate 'gross quantity' with loss percentage

mrp_bom_line_net_qty

Here are the added fields

change_net_qty_before
 ↓
change_net_qty_after

:yellow_circle: Line with difference with theorical values are set in yellow


legalsylvain (approved):
Thanks ! 

@OCA/manufacturing-maintainers : could you take a look on this one ?

mrp_bom_line_net_qty/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2023-Today: GRAP (http://www.grap.coop)
+# @author: Quentin Dupont
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "MRP BoM Line Net and Gross Quantity",
+    "version": "16.0.1.0.0",
+    "category": "Manufacturing",
+    "author": "GRAP, Odoo Community Association (OCA)",
```
legalsylvain, line 8:
```suggestion
    "author": "GRAP, Odoo Community Association (OCA)",
    "maintainers": ["quentinDupont"],
```

mrp_bom_line_net_qty/models/mrp_bom_line.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    # Functions to change product fields
+    def calculate_qty_net_theoretical(self, product_qty_gross, loss_percentage):
+        return (1 - loss_percentage / 100) * product_qty_gross
+
+    @api.depends("product_qty", "product_qty_net", "loss_percentage")
+    def _compute_diff_product_qty_gross_net(self):
+        for bom_line in self:
+            _bomline_qty_net_theorical = self.calculate_qty_net_theoretical(
```
hparfr:
```suggestion
            _bom_line_qty_net_theorical = self.calculate_qty_net_theoretical(
```

be consistent

mrp_bom_line_net_qty/models/mrp_bom_line.py
```diff
@@ -0,0 +1,67 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    product_qty_net = fields.Float(
+        string="Net quantity",
+        help="Weight after any preparation, e.g. weight of carrots after they are grated.",
```
hparfr:
```suggestion
        help="Weight after any preparation, e.g. weight of carrots after they are grated.",
```

A bit too specific about food processing.

mrp_bom_line_net_qty/models/mrp_bom_line.py
```diff
@@ -0,0 +1,67 @@
...
+        digits="Product Unit of Measure",
+    )
+    loss_percentage = fields.Float(
+        string="Loss %",
+        help="Percentage loss, for example, when grating carrots",
```
hparfr:
```suggestion
        help="Percentage loss, e.g. , when grating carrots",
```

hparfr (changes requested):
Code review.
Some minor remarks.

quentinDupont:
> Code review. Some minor remarks.

thanks for the review, I make the changes in the [last commit](https://github.com/OCA/manufacture/pull/1214/commits/705153aabef9b9590d403077a579cbcea4a34858) !

hparfr (approved):
LGTM

mrp_bom_line_net_qty/models/mrp_bom_line.py
```diff
@@ -0,0 +1,67 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    product_qty_net = fields.Float(
+        string="Net quantity",
+        help="Weight after any preparation",
```
hparfr: i've been a bit quick on the review.

"Weight" and "preparation" are really linked in your domain

mrp_bom_line_net_qty/models/mrp_bom_line.py
```diff
@@ -0,0 +1,67 @@
...
+        digits="Product Unit of Measure",
+    )
+    loss_percentage = fields.Float(
+        string="Loss %",
+        help="Percentage loss, e.g., when grating carrots",
```
hparfr: here too

hparfr (changes requested):
I submitted my review to quick.

There is still minor issues with the help messages

mrp_bom_line_net_qty/models/mrp_bom_line.py
quentinDupont: do you mean that i should have a more generic help message ?
do you have a proposal ? :)

quentinDupont: Just "Net quantity" ?
Or "Quantity after process"

hparfr: yes like this. 
you may have liters or unit as input components instead of "weight"

bosd:
@quentinDupont Can you attend to the review comments?

mrp_bom_line_net_qty/models/mrp_bom_line.py
quentinDupont: I chose "Quantity after process" for the help message

mrp_bom_line_net_qty/models/mrp_bom_line.py
quentinDupont: and here → "Percentage loss during process"

quentinDupont:
Thanks for the reminder @bosd 
I made the changes, you can review :)

bosd:
Thanks, I just functionally tested it.
The flow  from filling in a unit, los and calculate the net quantity was clear and working for me.

I had some difficulty, to do the scenario the other way around.
For example, when you measure the output, and you know the input quantity. But want to calculate the loss percentage.
Took me some time to find this scenario is supported as well. But I had to click on the arrow to trigger the calculation.
As opposed to the first scenario, hitting the save button was enough.

So, my suggestion is to put a calculator symbol next to the arrows for clarity and user friendlyness.

bosd (approved):
LGTM, non blocking review comments about usability

quentinDupont:
thank for your review and point of view @bosd 
In my opinion, i can't see perfect solution.
My chosen angle of attack was to have clear arrows to understand which value will be modified, and to show with the yellow line that there is inconsistency between quantity, loose percentage and net quantity.

In the screenshoot, the two first lines are OK, but the last one needs an action to be ok ; is the net quantity not been calculated ? or the loss percentage false ?

We can also have a button to calculate loss percentage but imho it will be to much buttons and will affect UX

image

bosd:
> In my opinion, i can't see perfect solution.

Me neither, just thinking out loud..
Will this work.. When one enters two values. Just show the orange and a generic calculator symbol to compute the missing value.

quentinDupont:
> I submitted my review to quick.
> 
> There is still minor issues with the help messages

Can you review this again ? :slightly_smiling_face: 
Edit : it was "non blocking review comments" so here we go

hparfr (approved):

PR #1223, opened 2024-03-06
[16.0][FIX] mrp_packaging_default: Propagate packaging in stock move

EmilioPascual:
Before fix,If you confirm a sales order with multi-step delivery, there will be movements where packaging is not assigned. It is because in those moves in the creation dictionary the key bom_line_id is set to False.

@yajo please review.

MT-5356 @moduon


yajo (approved):
/ocabot merge patch

PR #1227, opened 2024-03-11
[16.0][ADD] mrp_bom_product_price_margin

quentinDupont:
New module that helps handling cost and sale price of product linked to bill of material.

Add a new panel in BoM with Product.
mrp_bom_product_price_margin_diff

↓ You click on `Apply this cost..` and panel displays only useful information on Sale price

mrp_bom_product_price_margin_no_diff

Also in tree view, quickly see difference between Product standard price and BoM cost.
You can also change Product standard price here `1`
Lines are red when Sale margin is negative  `2`

mrp_bom_product_price_margin_tree


polchampion (commented):
Functionnally tested, LGTM

Slight English language improvement recommended : use "theoretical" instead of frenchicized/deprecated "theorical"

quentinDupont:
> Functionnally tested, LGTM
> 
> Slight English language improvement recommended : use "theoretical" instead of frenchicized/deprecated "theorical"

Thank uuuuu for your review !

This term is not used in this module but in dependency [here on product_standard_margin](https://github.com/OCA/margin-analysis/tree/16.0/product_standard_margin).

legalsylvain:
> This term is not used in this module but in dependency [here on product_standard_margin](https://github.com/OCA/margin-analysis/tree/16.0/product_standard_margin).

Totally removed : https://github.com/OCA/margin-analysis/pull/210

mrp_bom_product_price_margin/models/product_template.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    standard_price = fields.Float(
+        tracking=True,
```
legalsylvain, line 12: I think this change should go in another OCA or custom module. 
Dont think it's related to mrp. Do you ?

quentinDupont: My idea is : as we change a value on Product being on BoM Form view, it's great for the user to have this tracking to be sure that it "works". Other use case, you missclick on Apply price, and you want to check former price

legalsylvain: makes sense.

mrp_bom_product_price_margin/models/mrp_bom.py
```diff
@@ -0,0 +1,69 @@
...
+    product_standard_price = fields.Float(compute="_compute_product_standard_price")
+    standard_price = fields.Float(
+        string="BoM Unit Cost",
+        tracking=True,
+        digits="Product Price",
+        compute="_compute_standard_price",
+        help="Calculated with raw components cost divided by the BoM quantity.",
+    )
+    diff_product_bom_standard_price = fields.Boolean(
+        default=False,
```
legalsylvain:
```suggestion
```
I think default are useless for computed field. (Or maybe I missed something ?)

mrp_bom_product_price_margin/models/mrp_bom.py
```diff
@@ -0,0 +1,69 @@
...
+        compute="_compute_diff_product_bom_standard_price",
+    )
```
legalsylvain, line 27: Nice to have / not blocking : having an helper to explain this field. something like "Technical field used to display or hide button 'Apply this cost to Product standard price' in the form view" 
(or something similar.)

mrp_bom_product_price_margin/models/mrp_bom.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Fields related to sale price
+    product_sale_price = fields.Float(
+        string="Product Sale Price", related="product_tmpl_id.list_price"
+    )
+    product_margin_rate = fields.Float(related="product_tmpl_id.standard_margin_rate")
+    product_margin_rate_percentage = fields.Float(
```
legalsylvain: Hi see this field a little redundant with ``product_margin_rate``. and a little error prone. (in fact product_margin_rate is also a percentage)

If we remove the widget="percent", and display the field product_margin_rate in the tree view, here is the result.

image

i see it enough, and avoid having two fields.

mrp_bom_product_price_margin/i18n/fr.po
```diff
@@ -0,0 +1,209 @@
...
+#. module: mrp_bom_product_price_margin
+#: model_terms:ir.ui.view,arch_db:mrp_bom_product_price_margin.view_mrp_bom_sale_product_margin_form
+#: model_terms:ir.ui.view,arch_db:mrp_bom_product_price_margin.view_mrp_bom_tree
+msgid "It will set Product's standard price with this price"
+msgstr "Cela définira le coût du Produit avec ce total."
+
+#. module: mrp_bom_product_price_margin
+#: model:ir.model.fields,field_description:mrp_bom_product_price_margin.field_mrp_bom__product_margin_rate
+msgid "Margin (%)"
+msgstr "Marge théorique (%)"
```
legalsylvain: Margin in € -> Marge en €
Margin rate in % -> taux **de Marque** en %
Markup rate in % -> taux de marge en %

Cf : https://fr.wikipedia.org/wiki/Taux_de_marque

so, here is "Taux de marque (%)"

mrp_bom_product_price_margin/models/mrp_bom_line.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.depends(
+        "standard_price_subtotal", "bom_id.bom_line_ids.standard_price_subtotal"
+    )
+    def _compute_standard_price_subtotal_percentage(self):
+        for line in self:
+            bom = line.bom_id
+            total_price = sum(line.standard_price_subtotal for line in bom.bom_line_ids)
+            line.standard_price_subtotal_percentage = (
+                line.standard_price_subtotal / total_price if total_price != 0 else 0
```
legalsylvain, line 43: percentage are between 0 and 100 in database in recent odoo version.
```suggestion
                line.standard_price_subtotal / total_price * 100 if total_price != 0 else 0
```

mrp_bom_product_price_margin/views/view_mrp_bom.xml
```diff
@@ -0,0 +1,151 @@
...
+                    icon="fa-arrow-right"
+                    string="Update Product standard price"
+                    type="object"
+                    attrs="{'invisible': [('diff_product_bom_standard_price', '=', False)]}"
+                    help="It will set Product's standard price with this price"
+                />
+                <field name="product_standard_price" widget="monetary" />
+                <field name="product_sale_price" widget="monetary" />
+                <field name="product_margin_rate_percentage" widget="percentage" />
+            </xpath>
```
legalsylvain, line 147: All that new fields are interested to manage massively all the margin. However it adds a lot of column, that can  be useless in everyday life, and that overloads the screen.

I suggest to set all the field in optional="hide". What do you think ?

legalsylvain (commented):
some remarks inline.

Otherwise, LGTM. Thanks !

polchampion (approved):
LGTM

quentinDupont:
@legalsylvain thanks for all your great remarks. I addresses them in my last commit !

legalsylvain (approved):
Thanks !

PR #1231, opened 2024-03-13
[MIG][17.0]repair_refurbish

Vandan-OSI:


LoisRForgeFlow:
Hi @Vandan-OSI 

Thanks for you contribution, the module was moved to the repair repo in v15 https://github.com/OCA/repair/tree/15.0/repair_refurbish. You should propose your migration there based on the latest version available.

Thanks!

PR #1237, opened 2024-03-19
[16.0][IMP] mrp_production_quant_manual_assign: skip _check_qty()

AungKoKoLin1997:
`assign.manual.quants._check_qty()` would show an error when you tried to consume more than you were supposed to, however this should not be applied in the manufacturing context where component losses/overcomsumption should be added to the cost of produced product.

We let the standard check logic based on the BoM configuration ('Flexible Consumption')' handle the case where discrepancy arises.

@qrtl QT4406


yostashiro (commented):
@AungKoKoLin1997 Can you please update the commit message with the details.

yostashiro (approved):
Code review.

PR #1240, opened 2024-03-22
[16.0][ADD] mrp_bom_order_by_product_name

quentinDupont:
New simple module to change BoM order in tree view.
Sequence and id order is not user friendly → change it to Product name and code
 
mrp_bom_order_by_product_name


quentinDupont:
What do you think of that @OCA/manufacturing-maintainers ? :)

rvalyi (approved):
LGTM. HoweverI have a question: what is the point in packaging such a 1 line customization as an OCA module? I Mean:

- nobody will bother searching for an OCA module when he will need a 1 line _order prroperty change
- you will probably get zero benefit from 3rd party contribution on such one liner.
- it's a lot of boiler plate code just for 1 line of feature

Eventually I'm wrong if such module is actually a basis for a larger set of modules. Did I miss something?

quentinDupont:
> LGTM. HoweverI have a question: what is the point in packaging such a 1 line customization as an OCA module? I Mean:
> 
>     * nobody will bother searching for an OCA module when he will need a 1 line _order prroperty change
> 
>     * you will probably get zero benefit from 3rd party contribution on such one liner.
> 
>     * it's a lot of boiler plate code just for 1 line of feature
> 
> 
> Eventually I'm wrong if such module is actually a basis for a larger set of modules. Did I miss something?

This code was originally in custom code. But we prefer to propose software improvements to the community, and potentially influence future views so that this code is natively integrated.

legalsylvain (approved):

legalsylvain:
Hi @rvalyi. Not fully agree with you.

> nobody will bother searching for an OCA module when he will need a 1 line _order prroperty change

One line, or 100 lines, if the features is interesting, and generic, it can be interesting to propose the module to the OCA.

> you will probably get zero benefit from 3rd party contribution on such one liner.

Totally agree on that point.

> it's a lot of boiler plate code just for 1 line of feature
> Eventually I'm wrong if such module is actually a basis for a larger set of modules. Did I miss something? 

Well, this module looks quite trivial, changing the display order of the BoM in the bom list, (and bom search, in M2o fields). Odoo introduces here a weird order. Ordering by name is a convention in many softwares, and can be expected by end users. here, Odoo is ordering by default by sequence, that is forcing user to configure order, on each BoM creation, and can be a mess if you have a lot of BoM. (And ordering by id is like ordering by nothing... ;-))

So, better to set this module in the OCA scope, specially to share it to people that don't have development knowledge, and so can not write such simple changes in a custom module.

quentinDupont:
I reopen this PR as it's approved by two people, maybe a maintainer (someone else that me ofc) could merge it ? :)

ivantodorovich (approved):
/ocabot merge nobump

PR #1241, opened 2024-03-22
[16.0][ADD] mrp_bom_produce_delay

quentinDupont:
This module is designed to better respond to the flow of certain users who think about their production time when creating a bill of materials.

It adds `Product Delay` in BoM, and changes in BoM field will affect Product `Produce Delay`.

mrp_bom_produce_delay


quentinDupont:
ping @OCA/manufacturing-maintainers what do you think of this one ?

mrp_bom_produce_delay/view/mrp_bom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <!-- ************************************************************ -->
+    <!-- MRP Bill of Materials - Form -->
+    <!-- ************************************************************ -->
+
+    <record id="view_mrp_bom_form_produce_delay" model="ir.ui.view">
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="after">
+                <label for="produce_delay" string="Product Manuf. Lead Time" />
```
bealdav:
```suggestion
                <label for="produce_delay" string="Lead Time" />
```
It avoid to have a string split in 2 lines and we know it's about product and manufacture because it's in bom view

mrp_bom_produce_delay/model/mrp_bom.py
```diff
@@ -0,0 +1,25 @@
...
+# @author: Quentin DUPONT (quentin.dupont@grap.coop)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    produce_delay = fields.Float(
```
bealdav, line 11:
```suggestion
    lead_time = fields.Float(
```
delay is "retard" !?

bealdav (commented):
String deal

mrp_bom_produce_delay/model/mrp_bom.py
quentinDupont: produce_delay is the field in MRP in Product model !
I also don't understand it's technical english name (delay is "retard" in FR) but the help message is more clear
That why i take the logic of the core field  : https://github.com/odoo/odoo/blob/16.0/addons/mrp/models/product.py#L33

In Inventory tab of Product
image

mrp_bom_produce_delay/view/mrp_bom.xml
quentinDupont: Same as 1st comment, i took the term of the core field "produce_delay". But I agree with you that i can change the string

mrp_bom_produce_delay/model/mrp_bom.py
bealdav: Yes in product form you need string 'Manuf' because there are many purposes but in bom view manufacture vocabulary is implicit, don't repeat yourself, KISS.

mrp_bom_produce_delay/view/mrp_bom.xml
bealdav: Yes in product form you need string 'Manuf' because there are many purposes but in bom view manufacture vocabulary is implicit, don't repeat yourself, KISS.

mrp_bom_produce_delay/model/mrp_bom.py
bealdav: ok for technical name

hparfr (commented):
Can you specify the behaviour when a product as multiple boms (active or not) ?

bealdav (approved):
I don't want block, but in case of many boms by product, the behavior could be unexpected for some flows.

On my side, for the same needs I would just use a related readonly field.
But it works as it

quentinDupont:
Thanks for your remarks @bealdav and @hparfr, I've detailed it all on last commit :)
Indeed, it works well on single BoM (per product) context !

hparfr (approved):

quentinDupont:
\ocabot merge nobump

PR #1248, opened 2024-04-02
[13.0][FIX] mrp_multi_level: Substract from qty_to_order the qty_ordered

BernatPForgeFlow:
If there is any other process that orders quantities before creating the planned order, we need to consider it.


LoisRForgeFlow (approved):
/ocabot merge patch

PR #1251, opened 2024-04-08
[16.0][FIX] mrp_multi_level_estimate: adapt domain to search on child locations

JordiMForgeFlow:
The fix is needed to be aligned with the changes in https://github.com/OCA/manufacture/pull/1157

CC @LoisRForgeFlow


LoisRForgeFlow (approved):
/ocabot merge patch

PR #1255, opened 2024-04-08
[16.0][ADD] mrp_product_characterisation

quentinDupont:
_Migrate from custom Grap module used in v12.0 since more than 1 year_

New module for OCA

Adds characterisation "Intermediate Products" and adjust Components characterisation :+1: 
- Intermediate products : contained in a BoM AND manufactured (has a BoM)
- Component products : contained in a BoM AND NOT manufactured (has no BoM)

For example in bakeries, leavening is a intermediate product, and buckwheat flour
a component product.

intermediate_product

Add BoM filters linked to this new characterisation

In Product Form
image


hparfr (commented):
Does it overlap with  https://github.com/OCA/manufacture/blob/14.0/mrp_bom_hierarchy/model/mrp_bom.py#L100C5-L100C19 ?

quentinDupont:
> Does it overlap with https://github.com/OCA/manufacture/blob/14.0/mrp_bom_hierarchy/model/mrp_bom.py#L100C5-L100C19 ?

Nope it doesn't seem to overlap but rather it completes it.
`mrp_bom_hierarchy` adds informations on BoM and BoM lines and permits to navigate easily between BoM child and parents, this module adds informations on Product to better classify them

hparfr (approved):

bealdav (approved):

quentinDupont:
ping @OCA/manufacturing-maintainers :-)

grindtildeath (commented):
2 things:
 - If the same product cannot be both component and intermediate, what about using a Selection field instead of two Booleans?
 - What about using "Finished", "Semi-finished" and "Component" as possible values instead of "Final", "Intermediate" and "Component"? AFAICS both wordings are widely used in the industry, so it's more nitpicking than anything else.

quentinDupont:
> 2 things:
> 
>     * If the same product cannot be both component and intermediate, what about using a Selection field instead of two Booleans?
> 
>     * What about using "Finished", "Semi-finished" and "Component" as possible values instead of "Final", "Intermediate" and "Component"? AFAICS both wordings are widely used in the industry, so it's more nitpicking than anything else.

Hello and thanks for your time :)

- Well, actually product can be neither component or semi-finished / intermediate so i think a Selection field would not be so appropriate what do you think ?
- By default, it's "Manufactured product" and not "Finished products" :) Semi-finished is not bad, i have no preference. In french, i think "intermediate" is more accurate but i have no preference for the field name and english name.

legalsylvain:
@hparfr : any feedback about naming ? see : https://github.com/OCA/manufacture/pull/1255#issuecomment-2247401490

hparfr:
Does it make sense to be at the same time an Intermediate = True and a Component = True ?
If not, use a selection.
If yes, both are valid.

About the naming, I do prefer "Finished", "Semi-finished" and "Component".

In French, semi-fini is also my prefred word at the level of the factory, it's more precise. I use "intermediaire" from a macro point of view. Like minerals, paper or wood is an intermedia product. It's more used in an economical context.

quentinDupont:
> Does it make sense to be at the same time an Intermediate = True and a Component = True ?
> If not, use a selection.
> If yes, both are valid.

The answer is : not possible
I answer earlier that actually product can be neither component or semi-finished / intermediate so i think a Selection field would not be so appropriate

> About the naming, I do prefer "Finished", "Semi-finished" and "Component".
> 
> In French, semi-fini is also my prefred word at the level of the factory, it's more precise. I use "intermediaire" from a macro point of view. Like minerals, paper or wood is an intermedia product. It's more used in an economical context.

For the choice of the word, I think it really depends on the professional field where this module would be used.
@grindtildeath said that _both wordings are widely used in the industry, so it's more nitpicking than anything else._ and @hparfr you're talking about a preference for factories, and I'm more inclined towards the food industry, where the word “semi-finished” isn't used at all.

I will wait to see if there is no veto vote or new arguments, and i'm in favor of trying it this way and merging the PR :)

grindtildeath (approved):
Let's avoid nitpicking. IMO this module can live as it is now

PR #1258, opened 2024-04-10
[16.0][FW-PORT] mrp_restrict_lot : avoid duplicate mo name in case of cancellation qty changes ...

metaminux:
Forward port of #1010


metaminux:
Hello @florian-dacosta,

Can you take a look on this PR ? This is just forward porting...

Regards,

florian-dacosta (commented):
LGTM but I believe it would be nice to add an index to the field lot_producing_id, since Odoo will make an systematic search on it...
What do you think ?

florian-dacosta (approved):
I approve it anyway since it has been this way in v14
I still think it deserves an index, but it can be added in an other PR anyway

PR #1259, opened 2024-04-10
[16.0][IMP] mrp_restrict_lot : fills Lot/Serial Number to produce with the …

metaminux:
…restricted one from finished move

This PR is linked to OCA/stock-logistics-workflow#1560 and OCA/sale-workflow#3057 

If lot/serial number restriction changes on the main product stock move of a manufacturing order, the Lot/Serial Number of the manufacturing order is updated accordingly.

### DIFF
--- a/mrp_restrict_lot/models/__init__.py
+++ b/mrp_restrict_lot/models/__init__.py
@@ -1,1 +1,2 @@
+from . import mrp_production
 from . import stock_rule
--- a/mrp_restrict_lot/models/mrp_production.py
+++ b/mrp_restrict_lot/models/mrp_production.py
@@ -1,0 +1,20 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    lot_producing_id = fields.Many2one(
+        compute="_compute_lot_producing_id", store=True, readonly=False
+    )
+
+    @api.depends("move_finished_ids.restrict_lot_id")
+    def _compute_lot_producing_id(self):
+        for order in self:
+            finish_move_lot = order.move_finished_ids.filtered(
+                lambda m: m.product_id == order.product_id
+            ).restrict_lot_id
+            if finish_move_lot:
+                order.lot_producing_id = finish_move_lot


metaminux:
Still waiting for dependencies to be merged...

metaminux:
Hi @florian-dacosta 

Now that OCA/stock-logistics-workflow#1560 has been merged, can we process this one ?

florian-dacosta:
@metaminux I am not 100% comfortable with this one for now.
For now, the `lot_producing_id` is set on the MO by the pull rule, and at this stage, there is no restrict lot on the finished move.
Meaning that if for any reason, the recompute is triggered, the `lot_production_id` will be cleared on the MO...
Well, maybe there are no reason for a recompute to happen, but it seems quite risky to me.

Maybe just overriding the `write` on the finished move to propagate the lot would be better in this case ? 
What do you think?

metaminux:
Hi @florian-dacosta 

Would you feel better if I add a **store=True** on the field ? 

I don't see clearly why there would be a recompute other than when the restrict_lot_id field is updated on move_finished_ids (like exposed in the depends...)

Not sure about a **readonly=False** or an inverse, to get an editable computed field...

florian-dacosta:
Oh yes, the `lot_producing_id` field is not stored anymore, we definitly can't have that! (I did not see it before, I though it was stored!)
The field should at least be store=True and readonly=False to keep the native functionalities.

Then, the field should not recompute unless the restrict_lot_id change on the finished move indeed, I may be a bit paranoid on this. 
But maybe we could at least add a `if` statement, to only update the lot_producing_id if the lot of the finished move is not empty?
Something like : 

```
    @api.depends("move_finished_ids.restrict_lot_id")
    def _compute_lot_producing_id(self):
        for order in self:
            finish_move_lot = order.move_finished_ids.filtered(
                lambda m: m.product_id == order.product_id
            ).restrict_lot_id
            if finish_move_lot:
                order.lot_producing_id = finish_move_lot
```

metaminux:
Thank you @florian-dacosta 

I added `store=True` and `readonly=False` to keep native behaviour, that sounds better to me too.

I also added the condition on empty lot, even if I wonder why exactly we don't want to propagate to the manufacturing order the choice to remove a previously selected lot... 
I suppose this will happen if we decide to reassign the lot/serial to another MO, so why not emptying this one ?
Anyway, as we could change it manually now (thanks to the `readonly=False`) let it be so :)

Feel free to merge 😉

emiliesoutiras (approved):

bosd (approved):
Very quick code review lgtm

florian-dacosta:
ocabot merge patch

metaminux:
> ocabot merge patch

@florian-dacosta Missing a slash ?

PR #1261, opened 2024-04-12
[14.0][IMP] mrp_subcontracting_resupply_link: add init hooks

JordiMForgeFlow:


victoralmau (approved):
Functionally tested locally.

Ping @pedrobaeza

pedrobaeza (commented):
With this method, are the FK and indexes for the m2m table created by the ORM?

JordiMForgeFlow:
@pedrobaeza yes:

image

MiquelRForgeFlow:
> are the FK and indexes for the m2m table created by the ORM?

This has been the case in Odoo for a long while, at least since v12.

PR #1263, opened 2024-04-12
[16][MIG] mrp_workcenter_hierarchical

florian-dacosta:
I also made some changes on the module.
The original purpose of the module has change because of this PR https://github.com/OCA/manufacture/pull/865
I did see this only after it was merged and it made the module useless for the use case it had been created in the first place, which is to be able to group workcenter by parent workcenters.

The change of behavior is not really explained by a use case in the above PR but I have restored the original behavior and I have added a parameter to be able to work with the behavior introduced in the above PR.

I also added a feature to be able to change the workcenter's workcenter for another one of the same group / top parent.
Also add test for both cases.

@mymage @bealdav


florian-dacosta:
@bizzappdev @andreampiovesana @marcelofrare as old contributors/reviewers, it may interest you

mymage:
@florian-dacosta Hi, sorry, I didn't see the original post.
Thanks to the integration of my PR as an option.
There's definitely something I don't understand about your way of seeing things.
I added two more levels (ok it's theory but I think it's correct to test) and the result is this:
immagine
Why the top parent-level-1 is 12345? I assume it must be 1234567.

florian-dacosta:
@mymage 
In my way of seeing things, parent level 1 is supposed to be the top parent. The top of the pyramid, no matter how much level there are.
12345 is the top parent, as it itself do not have any parent.
All other workcenters (1234567 included) are child (or great child or great great child etc) of 12345 in you example.

A more concrete example for the use case would be for instance : 
Workcenter 1 : Cutting Machines
Workcenter 2 : Cutting machines room 1
Workcenter 3 : cutting machines room 2
Workcenter 4: Cutting machine 1 (parent is Cutting machines room 1)
Workcenter 5: Cutting machine 2 (parent is Cutting machines room 1)
Workcenter 6: Cutting machine 1 (parent is Cutting machines room 2)
Workcenter 7: Cutting machine 2 (parent is Cutting machines room 2)

If you group by parent level 1, you have all your cutting machine in a group.
Grouping y parent level 2, you can see all cutting machines by room, etc...
It is used to make group of workcenter.

I hope it is clearer

mymage:
@florian-dacosta Thanks, now it is clear and I "wake up" and set the right data to test it.
Just a last doubt: on runboat now this is the result: why?
immagine

florian-dacosta:
@mymage You mean, why 1 and 12 workcenters have not 1234567 as parent level 1, right ?

My guess is that it is a "bug" and the parent level did not recompute for these 1 and 12.
This because of the invalidation condition : https://github.com/akretion/manufacture/blob/16.0-mig-mrp_workcenter_hierarchical/mrp_workcenter_hierarchical/models/mrp_workcenter.py#L49
It takes only 3 level of parent_id change. 

I am not sure what would be the best way to solve this not... if we really want/need to solve this.
Indeed, in theory, we may have an infinite number of level, but in practice, the module show only 3 levels and I doubt we would really have more than that?
In my use case, I always have 2 level max, the third beeing useless.
Also, I guess the bug must be present in v14 already (and it did not seem to be an issue to anyone.
What do you think ?

mymage:
> @mymage You mean, why 1 and 12 workcenters have not 1234567 as parent level 1, right ?

I can share your opinion. Maybe It would be useful to indicate this on some part in the readme

mymage (approved):

bealdav (approved):
LGTM, thanks to now have both behaviors and fully tested

Thanks @florian-dacosta

PR #1266, opened 2024-04-26
[14.0][FIX] mrp_unbuild_subcontracting, action_validate expect to be singleton

ChrisOForgeFlow:
@ForgeFLow

### DIFF
--- a/mrp_unbuild_subcontracting/models/stock_picking.py
+++ b/mrp_unbuild_subcontracting/models/stock_picking.py
@@ -56,14 +56,16 @@ class StockPicking(models.Model):
                 unbuild_ids_backorder = unbuilds_to_done.filtered(
                     lambda u: u.state == "draft"
                 ).ids
-            unbuilds_to_done.with_context(
-                subcontract_move_id=True, mo_ids_to_backorder=unbuild_ids_backorder
-            ).action_validate()
-            move = self.move_lines.filtered(lambda move: move.is_subcontract)
+            for unbuild in unbuilds_to_done:
+                unbuild.with_context(
+                    subcontract_move_id=True, mo_ids_to_backorder=unbuild_ids_backorder
+                ).action_validate()
+            moves = self.move_lines.filtered(lambda move: move.is_subcontract)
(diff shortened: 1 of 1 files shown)


rvalyi (approved):

mrp_unbuild_subcontracting/models/stock_picking.py
```diff
@@ -56,14 +56,16 @@ def _action_done(self):
...
                 ).ids
-            unbuilds_to_done.with_context(
-                subcontract_move_id=True, mo_ids_to_backorder=unbuild_ids_backorder
-            ).action_validate()
-            move = self.move_lines.filtered(lambda move: move.is_subcontract)
+            for unbuild in unbuilds_to_done:
+                unbuild.with_context(
+                    subcontract_move_id=True, mo_ids_to_backorder=unbuild_ids_backorder
+                ).action_validate()
+            moves = self.move_lines.filtered(lambda move: move.is_subcontract)
```
yostashiro: I think it's more appropriate to make it like this.
```suggestion
            moves = picking.move_lines.filtered(lambda move: move.is_subcontract)
```

mrp_unbuild_subcontracting/models/stock_picking.py
```diff
@@ -56,14 +56,16 @@ def _action_done(self):
...
             finished_move = unbuilds_to_done.produce_line_ids.filtered(
-                lambda m: m.product_id == move.product_id
+                lambda m: m.product_id.id in moves.mapped("product_id").ids
```
yostashiro, line 65: Non-blocking, but no need to use `mapped()` I suppose.

rvalyi (approved):

mrp_unbuild_subcontracting/models/stock_picking.py
ChrisOForgeFlow: variable move is not present before next iteration, because that is better keep mapped on original variable to avoid singleton issue

yostashiro (approved):
Code review. Thanks for the update!

PR #1272, opened 2024-05-14
[17.0][MIG] quality_control_stock_oca: Migration to 17.0

peluko00:
Module migrated to version 17.0

cc https://github.com/APSL 155133

@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall please review

- [Issue] https://github.com/OCA/manufacture/issues/1125


mpascuall (approved):
Working good, checked in runboat

lbarry-apsl (changes requested):
@peluko00 ,

This error is appearing in the view:

Uncaught Promise > Can not evaluate python expression: ([('picking_id', '=', id), ('state', 'not in', ['draft', 'waiting'])]) Error: Name 'id' is not defined

peluko00:
> @peluko00 ,
> 
> This error is appearing in the view:
> 
> Uncaught Promise > Can not evaluate python expression: ([('picking_id', '=', id), ('state', 'not in', ['draft', 'waiting'])]) Error: Name 'id' is not defined

Fix it, thanks!

lbarry-apsl (approved):
Works well, reviewed on the Runboat

quality_control_stock_oca/__manifest__.py
```diff
@@ -5,13 +5,13 @@
...
     "name": "Quality control - Stock (OCA)",
-    "version": "16.0.1.0.0",
+    "version": "17.0.1.0.0",
     "category": "Quality control",
     "license": "AGPL-3",
     "author": "OdooMRP team, AvanzOSC, Serv. Tecnol. Avanzados - Pedro M. Baeza, "
     "Agile Business Group, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/manufacture",
-    "depends": ["quality_control_oca", "stock"],
+    "depends": ["quality_control_oca", "stock", "sale_management"],
```
pedrobaeza: Why the introduction of this dependency?

pedrobaeza (changes requested):

quality_control_stock_oca/__manifest__.py
peluko00: New field sale_line_warn in product_template https://github.com/odoo/odoo/blob/c25013c79cf315856799ec009969c083d0c63965/addons/sale/models/product_template.py#L19 for tests of quality_control_oca
and needs to depends of sale module

pedrobaeza: That's because you are testing it when the module is already installed. It's not correct. Please remove the dependency, and don't add these dependencies because you retest the module being already installed. You must always uninstall it and test on installation only, or just add this dependency as "temporary hack", but finally removing it for the final shot.

peluko00: How can add a temporary depency only for testing?

pedrobaeza: I mean in your local. You add the dependency, but before committing, you remove it.

peluko00: Yes, but in runboat the test will fail no?

pedrobaeza: No, as they are tested on installation. Try it... (and it's not on runboat, but on GitHub actions).

This is not your fault though, it's a flaw of the ORM and testing system, so we have to adapt to it, but we can't put any possible dependency that adds a required field in any arbitrary model.

peluko00: Just understand, sorry for the confusion hehe

pedrobaeza: As extended information, this happens because when a module like `sale_management` is already installed, the DB not null constraint is already put in place, but on the test phase, the Python code for the default value of `sale_management` is not yet loaded. The only way to avoid this problem is to test at the same time that the module is installed, so:

- or `sale_management` is still not installed, so you don't have the DB constraint.
- or `sale_management` is installed, but both the DB constraint and the default value are in place.

quality_control_stock_oca/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2014 Oihane Crucelaegui - AvanzOSC
+# Copyright 2018 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Quality control - Stock (OCA)",
+    "version": "17.0.1.0.0",
+    "category": "Quality control",
+    "license": "AGPL-3",
+    "author": "OdooMRP team, AvanzOSC, Serv. Tecnol. Avanzados - Pedro M. Baeza, "
```
pedrobaeza: One last thing. Please change this that they are old references:

```suggestion
    "author": "Tecnativa, AvanzOSC, "
```

pedrobaeza (commented):
/ocabot migration quality_control_stock_oca

quality_control_stock_oca/__manifest__.py
peluko00: Very good information about it, many thanks for the extra help!

quality_control_stock_oca/__manifest__.py
peluko00: Done!

pedrobaeza (approved):
Thank you

/ocabot merge nobump

PR #1284, opened 2024-06-14
17.0 mig mrp bom comparison

Urvisha-OSI:


peluko00 (changes requested):
The module works good but the commits message is not correct according of OCA convention.
Check [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0)
The module works good

peluko00 (commented):
The module works good but the commits message is not correct according of OCA convention.
Check [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0)

PR #1298, opened 2024-07-02
[15.0] mrp_sale_info: Materialize link between sale order lines and created MOs

grindtildeath:
Improve test with 3 steps delivery route and use RecordCapturer

### DIFF
--- a/mrp_sale_info/models/__init__.py
+++ b/mrp_sale_info/models/__init__.py
@@ -2,4 +2,6 @@
 
 from . import mrp_production
 from . import mrp_workorder
+from . import sale_order_line
+from . import stock_move
 from . import stock_rule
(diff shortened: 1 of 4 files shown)


sebalix (approved):

ajaniszewska-dev (approved):

grindtildeath (commented):
ping @pedrobaeza @tafaRU (as you're of the few contributors)

pedrobaeza:
About the PR formal thing, please follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

About the feature itself, the module is for getting the sales info into the production. This seems bit to go in the opposite direction, isn't it?

grindtildeath:
@pedrobaeza Thx I'll fix the commit message before merging.

If the goal of the module is only to provide sale information on the MO, and not MO information on the sale, we could still add the M2m field on mrp.production as well, but as it's wasn't a requirement, I did not add it...

IMO as long as the feature doesn't harm, I tend to prefer extending the scope of an existing module providing similar features than creating a dedicated module, which in turn would probably bring more confusion to end users in what each module provides.

pedrobaeza:
OK, go ahead with the commit renaming.

PR #1299, opened 2024-07-04
[17.0][MIG] mrp_lot_number_propagation

astirpe:


mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -0,0 +1,174 @@
...
+                            "name": order.propagated_lot_producing,
+                        }
+                    )
+                order.with_context(lot_propagation=True).lot_producing_id = lot
+
+    def write(self, vals):
+        for order in self:
+            if (
+                order.is_lot_number_propagated
+                and "lot_producing_id" in vals
```
CasVissers-360ERP: @astirpe if lot_producing_id is False we should not raise the error

astirpe: @CasVissers-360ERP done here https://github.com/OCA/manufacture/pull/1299/commits/c62f997d23aaf9b98f80a10058f22c88a6317458 could you check?

CasVissers-360ERP: Thnx resolved.

CasVissers-360ERP (approved):

bosd:
@astirpe Can you please squash your last commit?

astirpe:
@bosd squashed, thanks!

bosd (approved):

bosd:
Seems like the @OCA-git-bot is still on holiday, as this one carries the approved label and can be merged.

Just noticed the "confusing" naming of this module.
Technical name is lot propagation, but the title reads serial propagation.
image

Highcooley:
> Seems like the @OCA-git-bot is still on holiday, as this one carries the approved label and can be merged.
> 
> Just noticed the "confusing" naming of this module. Technical name is lot propagation, but the title reads serial propagation. image

The idea was to leave this option open in the technical name to also propagate LOT numbers in the future and not just SN. Also Odoo std naming only talks about LOT and not about SN.

bosd:
Ok for me, Let's merge!

bosd (changes requested):
Tests are not completely run.
Please cherry pick 0762c0a9ea9cbedc25b072d8a034caae83573348 from #1437

astirpe:
@bosd done, thank you!

bosd:
@astirpe Thanks,
I just noticed the "Display Display Propagate Lot Number" field in the view of the BoM. 
Don't recall this from previous versions. Does the field do something? Is this intended?

Without Propagate box thicked:
image

with:
image

astirpe:
@bosd I think you are right, I don't see anything making use of "Display Propagate Lot Number" field. It was there since the first release of the module though.

sebalix:
@bosd @astirpe I don't remember neither, probably dead code from a first implementation. I could not find any specific code neither that could make use of this field on `mrp.bom.line`, so I would say you can remove it :+1:

bosd (approved):
Approving, as it should not block this pr. The issue is in multiple version. Better to have it in a separate commit for cherry-pciking.

StefanRijnhart (approved):
/ocabot merge nobump

StefanRijnhart:
@astirpe Please clean outdated rst readme fragments manually

grindtildeath (approved):
@astirpe Thanks for the mig. Can you please fix the build so we can merge? :pray:

PR #1302, opened 2024-07-15
[FIX] mrp_multi_level: fix mrp_area_id position on stock location for…

AlexPForgeFlow:
Fix position of mrp_area_id field on stock location form view

Issue:
image


LoisRForgeFlow (changes requested):
Actually, that field is legacy and it is now useless most likely. Can you just remove it completely from the location model and views?

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/manufacture/pull/1306

PR #1025, opened 2023-05-01
[16.0][ADD] mrp_subcontracting_skip_no_negative

AungKoKoLin1997:
This module skips the no negative check for subcontracting recepits when we use `stock_no_negative` module.
Related issue: https://github.com/OCA/stock-logistics-workflow/issues/1275
Depends on: [[16.0][IMP] stock_no_negative: add option to skip negative qty check](https://github.com/OCA/stock-logistics-workflow/pull/1284)
@qrtl

PR #1324, opened 2024-07-25
[16.0][FIX] mrp_subcontracting_skip_no_negative: consider backorder

fdesmottes:
During partial receipt of a subcontract due to the unavailability of components, then we have a bug because the negative stock message appears again.

image

After analysis with the help of @hedi_ad, the raison is we also test (by mistake) the availability status of the components on the manufacturing order left.


AungKoKoLin1997:
@fdesmottes 
> During partial receipt of a subcontract due to the unavailability of components, then we have a bug because the negative stock message appears again.

With your current approach, even we make full receipt of subcontract with some availability of components(eg, need 10 quantity for components but have only 5 quantity), the check will skip, isn't?

AungKoKoLin1997:
@fdesmottes This is the reason why we added the check for `reservation_state`. https://github.com/OCA/manufacture/pull/1025#discussion_r1361740224
After I checked the details, I assume checking `reservation_state` should not be related with no negative module. So, I think we can just remove this part in your PR.

fdesmottes:
@AungKoKoLin1997 Thank you for your feedback.

> With your current approach, even we make full receipt of subcontract with some availability of components(eg, need 10 quantity for components but have only 5 quantity), the check will skip, isn't?

Good question, but I think it's working. I will test this.

> After I checked the details, I assume checking reservation_state should not be related with no negative module. So, I think we can just remove this part in your PR.

I'm not sure I understand your answer. You want to remove the entire `for move in moves_with_no_check:` block?

AungKoKoLin1997:
> I'm not sure I understand your answer. You want to remove the entire `for move in moves_with_no_check:` block?

Yes. I mean it because this check is not related with no negative and should not be part of mrp_subcontracting_skip_no_negative. As I told you, this check was for checking `reservation_state` of MO and even the `reservation_state` of MO is not assigned, there is no negative stock in the location. So, my idea is just to pass the corresponding context for subcontract moves.

giarve:
With the current set of commits I do not see it as mergeable, they should be squashed right?


whats the relationship with PR #1467?

I see this module fails in some cases, but I am still trying to replicate when.

It indeed could involve cases with backorders

PR #1326, opened 2024-07-26
[16.0][IMP] mrp_multi_level: add hooks to improve extensibility

JordiBForgeFlow:
cc @LoisRForgeFlow

### DIFF
--- a/mrp_multi_level/wizards/mrp_multi_level.py
+++ b/mrp_multi_level/wizards/mrp_multi_level.py
@@ -291,7 +291,7 @@ class MultiLevelMrp(models.TransientModel):
             )
             # Do not create planned order for products that are Kits
             planned_order = False
-            if not product_mrp_area_id.supply_method == "phantom":
+            if product_mrp_area_id._should_create_planned_order():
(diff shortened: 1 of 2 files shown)


mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -814,6 +814,14 @@ def _init_mrp_inventory(self, product_mrp_area):
                 if invs:
                     po.mrp_inventory_id = invs[0]
 
+    def _should_not_build_time_phased_inventory(self, product_mrp_area):
```
JordiMForgeFlow: may be better to understand if we keep these hook methods as positive booleans. For example we already have 'should_create_planned_order', so this one could be 'should_build_time_phased_inventory'

JordiBForgeFlow: Fixed!

JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1335, opened 2024-08-14
[14.0][FIX] repair_picking*: fix conflicts in picking_ids field

DavidJForgeFlow:
When installed both repair_picking and repair_picking_after_done, this second module does not work as expected. With the following changes now them both work fine. However maybe we should make a common module for them both modules adding the buttons and the compute to picking_ids.

@ForgeFlow


gbrito (approved):

DavidJForgeFlow:
I've done some changes, first of all I've created a separate base module named repair_stock. Then I've removed the repeated code from both repair_picking and repair_picking_after_done modules. This fixes the compute method errors and the duplicated button for the pickings.
@ForgeFlow @gbrito

LoisRForgeFlow (approved):
LGTM :+1:

PR #1336, opened 2024-08-14
[16.0][IMP] mrp_multi_level: better logs during calculation

LoisRForgeFlow:
* Explicitly log which area and LLC is being computed.
* Correct LLC being logged.


gbrito (approved):

ChrisOForgeFlow (approved):
LGTM+

PR #1337, opened 2024-08-22
[15.0][FIX] mrp_sale_info: add measure to avoid infinite recursion error

JuMiSanAr:
There's a scenario where a move can be the origin of one of its own origin moves,
which would cause an infinite loop in recursive method `_get_orig_created_production_ids`.

We now include a measure to prevent this.

### DIFF
--- a/mrp_sale_info/models/stock_move.py
+++ b/mrp_sale_info/models/stock_move.py
@@ -6,11 +6,20 @@ from odoo import models
 class StockMove(models.Model):
     _inherit = "stock.move"
 
-    def _get_orig_created_production_ids(self):
+    def _get_orig_created_production_ids(self, already_scanned_ids=None):
         self.ensure_one()
         if self.created_production_id:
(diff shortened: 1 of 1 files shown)


mrp_sale_info/models/stock_move.py
```diff
@@ -11,6 +11,25 @@ def _get_orig_created_production_ids(self):
         if self.created_production_id:
             return self.created_production_id
         mrp_production_ids = set()
-        for move in self.move_orig_ids:
-            mrp_production_ids.update(move._get_orig_created_production_ids().ids)
+        # There's a scenario where the current move is in self.move_orig_ids.move_orig_ids,
```
sebalix: It is probably a data issue here, but in any case we need to be protected against infinite loop in the implementation.

sebalix (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1338, opened 2024-08-22
[FIX] Fix quality control view

marrasbinovo:


quality_control_oca/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 
 {
     "name": "Quality Control OCA",
-    "version": "16.0.1.2.1",
+    "version": "16.0.1.3.0",
```
pedrobaeza: Please don't change the version.

quality_control_oca/views/product_category_view.xml
```diff
@@ -1,10 +1,3 @@
-<?xml version="1.0" encoding="utf-8" ?>
```
pedrobaeza: Please don't remove all of this.

pedrobaeza (changes requested):
Follow commit message guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

and indicate in the extended message commit what is this fixing.

PR #1342, opened 2024-08-28
[14.0][IMP-FIX] mrp_subcontracting_partner_management: Define the correct picking_type (out) in the subcontracting_resupply rule

victoralmau:
Changes done:
- [x] Change the string of the `is_subcontractor_partner` field
- [x] Define the correct picking_type (out) in the subcontracting_resupply rule

Before this change the picking_type_id used to create the subcontracting_resupply rule was "fixed”, which is incorrect if you do not use the main company or multi-company.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT49962

### DIFF
--- a/mrp_subcontracting_partner_management/models/res_partner.py
+++ b/mrp_subcontracting_partner_management/models/res_partner.py
@@ -222,7 +222,11 @@ class ResPartner(models.Model):
         prop = self.env["ir.property"]._get(
             "property_stock_production", "product.template"
         )
-        picking_type = self.env.ref("stock.picking_type_out", raise_if_not_found=False)
+        company = self.company_id or self.env.company
+        warehouse = self.env["stock.warehouse"].search(
+            [("company_id", "=", company.id)], limit=1
+        )
+        picking_type = warehouse.out_type_id
         route = self.env.ref(
             "mrp_subcontracting.route_resupply_subcontractor_mto",
             raise_if_not_found=False,
--- a/mrp_subcontracting_partner_management/models/res_partner.py
+++ b/mrp_subcontracting_partner_management/models/res_partner.py
@@ -6,7 +6,7 @@ from odoo import api, fields, models
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    is_subcontractor_partner = fields.Boolean(string="Subcontractor")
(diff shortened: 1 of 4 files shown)


mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,7 +6,7 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    is_subcontractor_partner = fields.Boolean(string="Subcontractor")
+    is_subcontractor_partner = fields.Boolean(string="Subcontractor partner")
```
pedrobaeza: Isn't better?

```suggestion
    is_subcontractor_partner = fields.Boolean(string="Is subcontractor")
```

mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -222,7 +222,11 @@ def _create_route_rule_for_subcontracting_resupply(self, vals):
         prop = self.env["ir.property"]._get(
             "property_stock_production", "product.template"
         )
-        picking_type = self.env.ref("stock.picking_type_out", raise_if_not_found=False)
+        company = self.company_id or self.env.company
+        warehouse = self.env["stock.warehouse"].search(
```
pedrobaeza, line 226: Why taking always the first warehouse? That might not be correct in some cases.

carlos-lopez-tecnativa (approved):
The first commit does not explain the reason, but after reading the internal task and the code, it looks good to take the picking from the warehouse instead of the default record. Thanks!

pedrobaeza:
Then please document in the commit that reasons.

pedrobaeza:
And here in the main message of the PR.

mrp_subcontracting_partner_management/models/res_partner.py
victoralmau: Exactly, t won't always be correct, but there is also no way of knowing the warehouse that will be used later. This approach is something that is already being used at https://github.com/OCA/manufacture/blob/14.0/mrp_subcontracting_partner_management/models/res_partner.py#L156.
It is important to clarify that the rule that is created can be edited afterwards.
What other option do you propose?

pedrobaeza (approved):
OK, this is the less bad one:

/ocabot merge patch

PR #1345, opened 2024-08-30
[16.0][IMP-FIX] mrp_subcontracting_partner_management: Define the correct picking_type (out) in the subcontracting_resupply rule

victoralmau:
FWP from 14.0: https://github.com/OCA/manufacture/pull/1342

Changes done:
- [x] Change the string of the `is_subcontractor_partner` field
- [x] Define the correct picking_type (out) in the subcontracting_resupply rule

Before this change the picking_type_id used to create the subcontracting_resupply rule was "fixed”, which is incorrect if you do not use the main company or multi-company.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT49962

### DIFF
--- a/mrp_subcontracting_partner_management/models/res_partner.py
+++ b/mrp_subcontracting_partner_management/models/res_partner.py
@@ -207,7 +207,11 @@ class ResPartner(models.Model):
         prop = self.env["ir.property"]._get(
             "property_stock_production", "product.template"
         )
-        picking_type = self.env.ref("stock.picking_type_out", raise_if_not_found=False)
+        company = self.company_id or self.env.company
+        warehouse = self.env["stock.warehouse"].search(
+            [("company_id", "=", company.id)], limit=1
+        )
+        picking_type = warehouse.out_type_id
         route = self.env.ref(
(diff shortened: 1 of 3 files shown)


mrp_subcontracting_partner_management/models/res_partner.py
```diff
@@ -6,7 +6,7 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    is_subcontractor_partner = fields.Boolean(string="Subcontractor")
+    is_subcontractor_partner = fields.Boolean(string="Is subcontractor")
```
pedrobaeza: Thinking twice, and following how is the rest of the module, this seems to be `Subcontractor partner`.

victoralmau: Ok, do I change this also in v14?

pedrobaeza: As you prefer

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #1346, opened 2024-09-01
[17.0][ADD] mrp_repair_order: New module mrp_repair_order

peluko00:
**New module**
The mrp_repair_order module allow to create a repair order from the manufacturing order when the manufacturing order is done.
You can access from the boths models to other order.

cc https://github.com/APSL 159985
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko please review


BernatObrador (approved):
LGTM, tested in runboat and works as intended.

ppyczko (approved):
LGTM, tested in runboat

mpascuall (approved):
LGTM, tested in runboat

mrp_repair_order/models/repair.py
```diff
@@ -0,0 +1,35 @@
...
+    mrp_count = fields.Integer(
+        "Mrp production order count", compute="_compute_mrp_production_count"
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        self.env["mrp.production"].browse(
+            self.env.context["default_mrp_id"]
+        ).repair_id = res.id
```
gbrito: This will fail if more then one repair order is created, res will contain multiple records

mrp_repair_order/models/repair.py
```diff
@@ -0,0 +1,35 @@
...
+        return res
+
+    def _compute_mrp_production_count(self):
+        self.mrp_count = len(
+            self.env["mrp.production"].search([("repair_id", "=", self.id)])
```
gbrito: Compute methods expect a record set. You need to loop through self.
I notice you are using a lot of search for the is of the repair order on the manufacturer order. Maybe would be a good idea to create a one2many for production orders and match on the field repair_id . Let Odoo have the records available to you

gbrito (changes requested):

mrp_repair_order/models/repair.py
peluko00: Thanks for the anotations, i make a one2many for production orders.
Can you review again please

peluko00:
Hi @gbrito, can you review again please. Thanks!

gbrito (approved):

peluko00:
Hi @pedrobaeza, could you review please when you have a moment? Thanks!

PR #1353, opened 2024-09-17
mrp_sale_info: Materialize link between sale order lines and created MOs

SirAionTech:
Forward port of https://github.com/OCA/manufacture/pull/1298 and https://github.com/OCA/manufacture/pull/1337.

### DIFF
--- a/mrp_sale_info/models/__init__.py
+++ b/mrp_sale_info/models/__init__.py
@@ -2,4 +2,6 @@
 
 from . import mrp_production
 from . import mrp_workorder
+from . import sale_order_line
+from . import stock_move
 from . import stock_rule
--- a/mrp_sale_info/models/sale_order_line.py
+++ b/mrp_sale_info/models/sale_order_line.py
@@ -1,0 +1,57 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    created_production_ids = fields.Many2many(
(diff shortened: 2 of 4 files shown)


SirAionTech:
@grindtildeath @sebalix @ajaniszewska-dev @JuMiSanAr @pedrobaeza maybe you want to have a look since you worked in the linked PRs?

I just edited a bit the commit messages in order to follow https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message.

pedrobaeza (changes requested):
This has provoked is done headaches, as the compute is launched on update time

SirAionTech:
Thanks for the review!
> This has provoked is done headaches, as the compute is launched on update time

I guess that's the usual behavior when a new stored computed field is added, why is this case different?
Sorry I didn't know about the headaches, I can't find any reference to this in the linked PRs.

pedrobaeza:
I mean that you should provide a init hook or auto_init override for pre-creating the DB schema and fast-populate data for existing records.

SirAionTech:
> I mean that you should provide a init hook or auto_init override for pre-creating the DB schema and fast-populate data for existing records.

Thanks for the hint, so at upgrade time the computed fields should be populated with an optimized calculation (you mean like SQL queries?) instead of letting the field compute itself, did I understand that correctly?

pedrobaeza:
Correct.

SirAionTech:
I tried to do a SQL query for computing all the values in `_auto_init` but it's a hell of relations of *2many fields (`created_production_ids`, `move_orig_ids`, `mrp_production_ids`); I tried to optimize the compute method itself, would you have another look?

PR #1355, opened 2024-09-24
[ADD] mrp_bom_line_skip_empty

SirAionTech:


anajuaristi:
Hi @SirAionTech 
Thank you for the contribution but... I see functional issue here, since having quantity=0 bom lines, must be an user error. If you avoid this lines, maybe production is not going to be right. 
I think that simply avoiding Bom lines having quantity = 0, would be better aproach. 
I would set quantity by default = 1 on bom line creation and I would not allow to write it if quantity is <=0, so you are forcing all the lines to have quantity > 0. 

My 2 cents.

IMHO you would obtain better result with simplier code but I would like to ask @JordiBForgeFlow and @pedrobaeza  if they agree on my aproach.

Thank you very much!
Ana

SirAionTech:
> I see functional issue here, since having quantity=0 bom lines, must be an user error. If you avoid this lines, maybe production is not going to be right.
> I think that simply avoiding Bom lines having quantity = 0, would be better aproach.
> I would set quantity by default = 1 on bom line creation and I would not allow to write it if quantity is <=0, so you are forcing all the lines to have quantity > 0.

Thanks for having a look!
That would be actually simpler, but _thanks_ to https://github.com/OCA/manufacture/tree/1f34d36b10fb9079add5e43cf7aaaf6811e2f60c/mrp_bom_line_formula_quantity the quantity in the BoM lines is irrelevant and it will be computed dynamically when the production is created.
At first I thought of modifying the linked module but then I noticed that even without it, the user could add BoM lines with 0 quantity and they would create production lines with 0 quantity.

Then I assumed that it is how the `mrp` module is supposed to work: if the user wants to add BoM lines with 0 quantity (not because of a user error), they will create production lines with 0 quantity.
This is why I created this module to modify such behavior.

pedrobaeza:
There's a reason for adding 0 quantity lines, as you may have some materials that are "optional", so you introduce them with 0, and thus, they create the corresponding consuming line in the production. You keep it 0 if that order doesn't consume them, or fill it with the proper quantity if not, so I think you should include the option of not creating the lines in `mrp_bom_line_formula_quantity`, not as an independent module, as both cases may coexist.

anajuaristi:
Hi both
@pedrobaeza  if you want to mantain quantity = 0 lines in production... this module has nosense because it will avoid to create lines to consume with quantity =0

@SirAionTech  if the user want to certenly include quantity = 0 lines in bom, then... why do you want to avoid creating them on production?

SirAionTech:
> There's a reason for adding 0 quantity lines, as you may have some materials that are "optional", so you introduce them with 0, and thus, they create the corresponding consuming line in the production. You keep it 0 if that order doesn't consume them, or fill it with the proper quantity if not, so I think you should include the option of not creating the lines in `mrp_bom_line_formula_quantity`, not as an independent module, as both cases may coexist.

Thanks, I didn't know how those lines could be useful.
I see two different features:
- dynamic quantity using formulas `mrp_bom_line_formula_quantity`
- not creating production lines with 0 quantity from BoM lines with 0 quantity

I think it's better if these features stay independent so that the user can have one or the other.
If the user does not want to use formulas but they do not want BoM lines with 0 quantity to become production lines with 0 quantity, then they only install this module.

> if the user want to certenly include quantity = 0 lines in bom, then... why do you want to avoid creating them on production?

In the use-case I've seen, the user didn't expect the BoM lines with 0 quantity to become Production lines.

pedrobaeza:
I don't think so, as you may have both formulas and some BoMs with manual lines with 0, so you want to skip the 0 lines of the formulas, but not the manual ones.

PR #1369, opened 2024-10-21
[16.0][FIX] mrp_multi_level: check for variant bom

### LINKED ISSUE #1366: [16.0][BUG] mrp_multi_level: using wrong variant BOM
Version 16.0 module mrp_multi_level is using the wrong variant BOM, when multiple variant BOMs exist.

This code should look for a variant specific BOM:

https://github.com/OCA/manufacture/blob/1f34d36b10fb9079add5e43cf7aaaf6811e2f60c/mrp_multi_level/wizards/mrp_multi_level.py#L183-L185

## To Reproduce

**Add attribute "Legs" and values "Steel" and "Aluminum" on product FP-2:**

Screenshot at 2024-10-16 18-37-06

**Create BOMs for the two variants as follows:**

Screenshot at 2024-10-16 18-32-09

Screenshot at 2024-10-16 18-32-38

**Create outbound move for product FP-2 (Aluminum):**

Screenshot at 2024-10-16 18-42-28

**We expect a planned order for SF-2, but we get a planned order for SF-1:**

Screenshot at 2024-10-16 18-35-28

matt454357:
fixes #1366

Using the _bom_find() method, we get the BOM with lowest sequence, whether it's a variant BOM or a template-only BOM (no product_id).

### DIFF
--- a/mrp_multi_level/models/product_mrp_area.py
+++ b/mrp_multi_level/models/product_mrp_area.py
@@ -85,6 +85,11 @@ class ProductMRPArea(models.Model):
         ],
         compute="_compute_supply_method",
     )
+    supply_bom_id = fields.Many2one(
+        comodel_name="mrp.bom",
+        string="Supply BoM",
+        compute="_compute_supply_method",
+    )
     qty_available = fields.Float(
         string="Quantity Available", compute="_compute_qty_available"
     )
@@ -205,20 +210,21 @@ class ProductMRPArea(models.Model):
         return rule
 
     def _compute_supply_method(self):
+        boms_by_product = self.env["mrp.bom"]._bom_find(self.mapped("product_id"))
         for rec in self:
             rule = rec._get_rule()
             if not rule:
                 rec.supply_method = "none"
+                rec.supply_bom_id = False
                 continue
             # Determine the supply method based on the final rule.
-            boms = rec.product_id.product_tmpl_id.bom_ids.filtered(
-                lambda x: x.type in ["normal", "phantom"]
-            )
+            bom = boms_by_product.get(rec.product_id, self.env["mrp.bom"])
             rec.supply_method = (
                 "phantom"
-                if rule.action == "manufacture" and boms and boms[0].type == "phantom"
+                if rule.action == "manufacture" and bom.type == "phantom"
                 else rule.action
             )
+            rec.supply_bom_id = bom
 
     @api.depends(
         "mrp_area_id", "supply_method", "product_id.route_ids", "product_id.seller_ids"
--- a/mrp_multi_level/wizards/mrp_multi_level.py
+++ b/mrp_multi_level/wizards/mrp_multi_level.py
@@ -178,14 +178,7 @@ class MultiLevelMrp(models.TransientModel):
 
     @api.model
     def _get_bom_to_explode(self, product_mrp_area_id):
-        boms = self.env["mrp.bom"]
-        if product_mrp_area_id.supply_method in ["manufacture", "phantom"]:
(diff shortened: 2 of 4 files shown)


JordiBForgeFlow:
Why not using _bom_find, which is the actual method that is used everywhere?

matt454357:
@JordiBForgeFlow
I chose not to use _bom_find for performance and clarity.  If you think _bom_find is better, I'm happy to change it.

JordiBForgeFlow:
@matt454357 My concern is that then the actual business logic is duplicated in _bom_find and in this code, introducing the possibility to diverge.

matt454357:
@JordiBForgeFlow I've changed it to use _bom_find.

mrp_multi_level/models/product_mrp_area.py
```diff
@@ -205,20 +210,19 @@ def _get_rule(self):
...
-            boms = rec.product_id.product_tmpl_id.bom_ids.filtered(
-                lambda x: x.type in ["normal", "phantom"]
-            )
-            rec.supply_method = (
-                "phantom"
-                if rule.action == "manufacture" and boms and boms[0].type == "phantom"
-                else rule.action
-            )
+                rec.supply_bom_id = False
+            elif rule.action == "manufacture":
```
LoisRForgeFlow: You are changing aditional the logic, BoM is applicable also in phanton and subcontracting for example, please keep the logic as it was without this new elif and else sentences and keep only the change regarding supply BoM and using the variant

mrp_multi_level/wizards/mrp_multi_level.py
```diff
@@ -195,7 +184,7 @@ def explode_action(
         mrp_date_demand = mrp_action_date
         if mrp_date_demand < date.today():
             mrp_date_demand = date.today()
-        bom = self._get_bom_to_explode(product_mrp_area_id)
+        bom = product_mrp_area_id.supply_bom_id
```
LoisRForgeFlow: For keeping extensibility, can you keep the method `_get_bom_to_explode` and return `product_mrp_area_id.supply_bom_id` in it?

mrp_multi_level/models/product_mrp_area.py
```diff
@@ -85,6 +85,11 @@ class ProductMRPArea(models.Model):
         ],
         compute="_compute_supply_method",
     )
+    supply_bom_id = fields.Many2one(
+        comodel_name="mrp.bom",
+        string="Supply BOM",
```
LoisRForgeFlow: BOM -> BoM

LoisRForgeFlow (changes requested):
Thanks for your contribution. See a couple of comments below, basically you are changing more things that what is strictly stated in the description of the PR

mrp_multi_level/models/product_mrp_area.py
matt454357: @LoisRForgeFlow  I pushed new code for this.  There is still a very small difference in the logic, but I can't find any reason it will be a problem.  It would be ugly to use the _bom_find() method, and match the original logic exactly.

Matching the original logic exactly, and using the _bom_find() method, would require one of the following:

- Call _bom_find() again (and potentially again), if BoM type not in ("normal", "phantom")
- Modify _bom_find_domain() and pass multiple BoM types in context
- Do a PR with core Odoo to make _bom_find_domain() accept a list of types

LoisRForgeFlow: Yes, I think now it is perfect, I don't see any big difference apart from the way the BoM is searched (which is now better)

LoisRForgeFlow (approved):
All good now, thanks :+1:

PR #1375, opened 2024-10-28
[14.0][IMP] repair_stock_move: relate lot_id and common methods

DavidJForgeFlow:
The value of lot id in repair lines is not related to the one of the move related. The idea is to make them correlated to improve visibility. Also added stock_rule._get_custom_move_fields method from repair_picking as it should be added in the module that adds the field.

@ForgeFlow

### DIFF
--- a/repair_stock_move/models/repair_line.py
+++ b/repair_stock_move/models/repair_line.py
@@ -11,6 +11,8 @@ class RepairLine(models.Model):
         comodel_name="stock.move",
         inverse_name="repair_line_id",
     )
+    # Actually the lot_id does not matches the value of the related move.line
+    lot_id = fields.Many2one(compute="_compute_lot_id", readonly=False, store=True)
 
     def create_stock_move(self):
         self.ensure_one()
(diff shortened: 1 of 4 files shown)


gbrito (changes requested):
Have a look if my comment makes sense, maybe it isn't the desired functionality

DavidJForgeFlow:
> Have a look if my comment makes sense, maybe it isn't the desired functionality

Hi, I can't see the comment!

repair_stock_move/models/repair_line.py
```diff
@@ -50,5 +52,18 @@ def _onchange_location(self):
...
+        for rec in self:
+            # The repair line only can contain 1 lot and the stock.move can have multiple, so
+            # if the lot id in the repair line is not present in the stock.move,
+            # it will be substituted by the first lot in the stock.move.
+            if rec.move_id.move_line_ids:
+                rec.lot_id = (
+                    rec.move_id.move_line_ids[0].lot_id
+                    if rec.move_id.move_line_ids.mapped("lot_id")
+                    else False
+                )
```
gbrito: There could be an edge case if the lot_id is not assigned on the first move_line_ids.
Don't know if makes sense like this. give it a try
```suggestion
            if rec.move_id.move_line_ids.lot_id:
                rec.lot_id = rec.move_id.move_line_ids.lot_id[0]
            else:
                rec.lot_id = False
```

gbrito (changes requested):
sorry, now I saved the comment. try again :)

repair_stock_move/models/repair_line.py
DavidJForgeFlow: I fixed the corner case thx! Your code give me some errors so I changed it a little bit!

Issue #1307, opened 2024-07-15
[15.0] CI tests broken by mrp_multi_level


grindtildeath:
Builds on branch 15.0 (I didn't check others) are broken by module mrp_multi_level:

```
ERROR odoo odoo.addons.mrp_multi_level.tests.test_mrp_multi_level: FAIL: TestMrpMultiLevel.test_04_mrp_multi_level
Traceback (most recent call last):
  File "/__w/manufacture/manufacture/mrp_multi_level/tests/test_mrp_multi_level.py", line 120, in test_04_mrp_multi_level
    self.assertEqual(len(sf_1_line_2), 1)
AssertionError: 0 != 1
 
```

Example on following PRs as well as master:
 - https://github.com/OCA/manufacture/pull/1300
 - https://github.com/OCA/manufacture/pull/1298

ping @JordiBForgeFlow @LoisRForgeFlow @ChrisOForgeFlow can you guys please have a look? :pray:

LoisRForgeFlow:
Hi @grindtildeath 

After restarting the jobs, the issue seems to be gone. Maybe it was temporary or maybe there is a window of time where they fail due to being out of working warehouse hours (some test uses the warehouse calendar).

rvalyi:
this was fixed by #1398

PR #1388, opened 2024-11-11
[TMP][DO NOT MERGE] Test for seeing CI state

pedrobaeza:


pedrobaeza:
@LoisRForgeFlow there's a failing test in `mrp_multi_level`. Can you please check?

LoisRForgeFlow:
I think it might be the same thing reported here https://github.com/OCA/manufacture/issues/1307. It is probably a timing issue, I suspect there is a range of time in which that tests fails. In my local currently is fine, I restarted the CI here to see if it gets green now.

pedrobaeza:
I don't think so. I have just opened the PR (so not around midnight), and it's consistent failing in all the attempts.

LoisRForgeFlow:
I'm checking in local with freeze_time and I am not succeeding making that test fail. It can also be an interference with another module. Checking.

LoisRForgeFlow:
An update, the conflict is with mrp_production_date_planned_finished, I will check in more details when I have time later.

LoisRForgeFlow:
@pedrobaeza Today tests passed, which I suspected. So the issue is only happening on Mondays. I will try to get care of it during this week and propose a fix for next Monday. In the meanwhile your other PR should be no longer blocked.

pedrobaeza:
Wow, it's the first time I see that tests are blocked depending on the week day. It would be good to have that fixed. Launching the merge on the other PRs.

PR #1382, opened 2024-11-07
[14.0][MIG] mrp_lot_number_propagation

TDu:
Backport from 15.0

PR #1390, opened 2024-11-11
[14.0][FIX] mrp_bom_attribute_match: Do not empty bom line uom when creating

grindtildeath:
When creating a new line, an onchange is played before the user selects the product. With the previous code, since no product was selected, this was emptying the default value. This was in turn breaking any other function that relied on this value being set.


grindtildeath:
cf https://github.com/OCA/manufacture/pull/1382#pullrequestreview-2424103146

TDu (approved):

ivantodorovich (approved):
/ocabot merge patch

PR #1407, opened 2024-11-28
[16.0][ADD] mrp_bom_priority

quentinDupont:
Replace sequence by priority in order to prioritize easily BoMs.
Priority is used to bring up the most important BoMs, the others being classified as by default.
This priority will be used to select the right BoM during manufacturing (thus replacing the sequence).

mrp_bom_view_tree

mrp_bom_view_form


mrp_bom_priority/views/view_mrp_bom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <record id="view_mrp_bom_tree" model="ir.ui.view">
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_tree_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="after">
+                <field name="priority" widget="priority" />
+            </xpath>
+            <xpath expr="//tree" position="attributes">
+              <attribute name="default_order">priority desc</attribute>
```
legalsylvain: default_order (or _order), make the item order by the field(s). if there are many items with the same value, the next order is "by id". that is not very user intuitive. See : 

image

In that case, I expect both "extra coulis" boms to be next to it.
something like that ? (not sure, just a proposal).

```suggestion
              <attribute name="default_order">priority desc, product_tmpl_id,code</attribute>
```

mrp_bom_priority/views/view_mrp_bom.xml
```diff
@@ -0,0 +1,40 @@
...
+    <!-- ************************************************************ -->
+    <!-- MRP Bill of Materials - Tree -->
+    <!-- ************************************************************ -->
+
+    <record id="view_mrp_bom_tree" model="ir.ui.view">
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_tree_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_id']" position="after">
+                <field name="priority" widget="priority" />
```
legalsylvain: Allow user to hide the field to simplify the view, if they don't use it.

```suggestion
                <field name="priority" widget="priority" optional="show"/>
```

legalsylvain (approved):
General review. LGTM.

some remarks inline. I wonder if we should better change the order in the code, to make it working everywhere. 

for the time being, here is the default order : sequence, id

https://github.com/odoo/odoo/blob/16.0/addons/mrp/models/mrp_bom.py#L19

mrp_bom_priority/views/view_mrp_bom.xml
quentinDupont: Yep I think it makes sense !

LoisRForgeFlow:
@quentinDupont Could you squash all commits into one?

quentinDupont:
Done @LoisRForgeFlow :-)

mrp_bom_priority/models/mrp_bom.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    # Column Section
+    priority = fields.Selection(
```
LoisRForgeFlow, line 12: @quentinDupont Looks like you are adding priority without any feature/behavior related to it. I would have expected to then modify the standard algorithm to find the bom to apply by default.

What is your use case?

quentinDupont: The use case is to quickly see your favorite BoMs, the one you check or change often
No algorithm changes but priority change in tree view here : https://github.com/OCA/manufacture/pull/1407/files#diff-a9763f49357eeb3d0f9b21812afde85a545efb1d3ccbf07c5593c97d000e8765R37
Thanks for your review and time :)

LoisRForgeFlow: I can imagine this being a bit confusing, because in the same view you have the sequence widget that it's what actually will be used to select a BoM first when manufacturing(https://github.com/odoo/odoo/blob/938b772ab60501ac7028da0edf2f66a758004363/addons/mrp/models/mrp_bom.py#L374) but that widget will not make any effect and you will see first the ones with more stars.

quentinDupont: Indeed, I understand your point of view.
In our Odoo, we've changed the order of BoM by priority then alphabetically: a behavior that we feel is much more expected and understandable for users. But I agree that it would be nice to be taken into account when choosing a BoM at the time of manufacturing → in the ROADMAP for this module?

LoisRForgeFlow: If you just add it to the ROADMAP and do not implement it, I won't block the PR but I won't approve it either. But I'm sure you can convince someone else.

quentinDupont: Ok so I think your criticism are relevant, I make the following changes : 
- replacing sequence in tree view → it was no longer working anyway with priority
- overriding native function to select best prioritized BoM for manufacturing (insted of their sequence) as you suggest
- add a test to cover this change and update readme

mrp_bom_priority/models/mrp_bom.py
```diff
@@ -14,3 +16,37 @@ class MrpBom(models.Model):
         default="0",
         help="Helps prioritize BoM.",
     )
+
+    @api.model
+    def _bom_find(self, products, picking_type=None, company_id=False, bom_type=False):
+        """Override native function to change order"""
```
LoisRForgeFlow: Maybe you can try to propose the introduction of a hook to odoo?

Otherwise the correct way to do it is by adding a hook via OCA module, see for instance this one https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_account_product_run_fifo_hook/hooks.py. We need this to keep proper inheritance, we cannot merge a module that simply overrides an standard one like this.

LoisRForgeFlow (commented):
Thanks for the improvements! a comment below

marcelsavegnago (approved):

LoisRForgeFlow (changes requested):
As per https://github.com/OCA/manufacture/pull/1407#discussion_r2180176382. We cannot merge with such an override.

legalsylvain:
Hi @LoisRForgeFlow. thanks for your review. could you take a look on my last commit. I implemented a lighter code that respect inheritance.
thanks !

LoisRForgeFlow (approved):
@legalsylvain Thanks, LGTM now :+1:

legalsylvain:
Thanks for tour review !

/ocabot merge nobump

PR #1408, opened 2024-12-02
[FIX] mrp_bom_image: use new syntax to filter elements by their classes

legalsylvain:
So, remove warning during the installation or update of this module

For the time being, here are the log during an update : 

```
WARNING odoo.addons.base.models.ir_ui_view: Error-prone use of @class in view mrp.bom kanban (mrp_bom_image.view_mrp_bom_kanban): use the hasclass(*classes) function to filter elements by their classes 
WARNING odoo.addons.base.models.ir_ui_view: Error-prone use of @class in view mrp.bom kanban (mrp_bom_image.view_mrp_bom_kanban): use the hasclass(*classes) function to filter elements by their classes 


```

CC : @quentinDupont 

thanks.


quentinDupont (approved):
Code review and tested, works fine :heavy_check_mark:

PR #1413, opened 2024-12-09
[17.0][FIX] mrp_multi_level: properly handle UserError in procurement.

RicardCForgeFlow:
Small fix
AttributeError: 'UserError' object has no attribute 'name'.

Example:
error_example

After the fix:
image

### DIFF
--- a/mrp_multi_level/wizards/mrp_inventory_procure.py
+++ b/mrp_multi_level/wizards/mrp_inventory_procure.py
@@ -97,7 +97,7 @@ class MrpInventoryProcure(models.TransientModel):
             for item in self.item_ids:
                 item.planned_order_id.qty_released += item.qty
         except UserError as error:
-            errors.append(error.name)
+            errors.append(str(error))
         if errors:
             raise UserError("\n".join(errors))
         return {"type": "ir.actions.act_window_close"}


MiquelRForgeFlow (approved):
The video captures are really nice.

LoisRForgeFlow (changes requested):
Hi @RicardCForgeFlow 

I'am not able to reproduce the issue, I cannot set a different UoM in the first place because the error raises:

https://github.com/user-attachments/assets/0eca9cf9-83ce-4e40-a84b-8c04a59c50df

Can you indicate the steps to reproduce?

RicardCForgeFlow:
Hi @LoisRForgeFlow I found the error doing this:

https://github.com/user-attachments/assets/762471ce-32cb-4e47-bf7e-2d1a77e8c4fc



> Hi @RicardCForgeFlow
> 
> I'am not able to reproduce the issue, I cannot set a different UoM in the first place because the error raises:
> 
>  2024-12-27.08-58-50.mp4 
> Can you indicate the steps to reproduce?

mrp_multi_level/wizards/mrp_inventory_procure.py
```diff
@@ -72,7 +72,6 @@ def default_get(self, fields):
 
     def make_procurement(self):
         self.ensure_one()
-        errors = []
```
LoisRForgeFlow: @RicardCForgeFlow  I see, thanks for the video, could you still keep the multiple error handling? I think you can fix the issue without raising in the very first error.

LoisRForgeFlow (changes requested):

mrp_multi_level/wizards/mrp_inventory_procure.py
RicardCForgeFlow: image
You are right, just changing this works

LoisRForgeFlow (approved):
right on! thanks

PR #1424, opened 2024-12-18
[17.0][FIX] mrp_multi_level: avoid create warehouse and location in mrp area view

RicardCForgeFlow:
Avoid bug that occurs when clicking `New` on the `warehouse_id` field if the Storage Locations setting is not active.
Before PR: once you click on `New` you cannot close
PR-mrp_mulit_level

After PR: `New` button removed
PR-mrp_mulit_level2

### DIFF
--- a/mrp_multi_level/views/mrp_area_views.xml
+++ b/mrp_multi_level/views/mrp_area_views.xml
@@ -35,8 +35,8 @@
                     <group colspan="4" col="2">
                         <group>
                             <field name="active" invisible="1" />
-                            <field name="warehouse_id" />
-                            <field name="location_id" />
+                            <field name="warehouse_id" options="{'no_create': True}" />
+                            <field name="location_id" options="{'no_create': True}" />
                             <field
                                 name="company_id"
                                 groups="base.group_multi_company"


mrp_multi_level/views/mrp_area_views.xml
```diff
@@ -35,7 +35,11 @@
                     <group colspan="4" col="2">
                         <group>
                             <field name="active" invisible="1" />
-                            <field name="warehouse_id" />
+                            <field
+                                name="warehouse_id"
+                                options="{'no_create': True, 'no_create_edit': True}"
+                                force_save="1"
```
LoisRForgeFlow: why the force_save?

RicardCForgeFlow: You are right, the only change needed is `'no_create': True`

mrp_multi_level/views/mrp_area_views.xml
```diff
@@ -35,7 +35,7 @@
                     <group colspan="4" col="2">
                         <group>
                             <field name="active" invisible="1" />
-                            <field name="warehouse_id" />
+                            <field name="warehouse_id" options="{'no_create': True}" />
                             <field name="location_id" />
```
LoisRForgeFlow: Can you also add the same for location_id?

RicardCForgeFlow: Is it necessary? The “Storage Locations” feature does not affect the creation of new locations, only new warehouses. So it doesn't fail in this field

LoisRForgeFlow: not strictly but to avoid errors, not a good place to create locations on the fly.

LoisRForgeFlow (approved):
LGTM :+1:

JasminSForgeFlow (approved):

LoisRForgeFlow:
@RicardCForgeFlow could you fwport to 18.0, and backport to 16.0 and 15.0?

PR #1427, opened 2024-12-18
[17.0][ADD] mrp_bom_compute_sales_price

SMaciasOSI:


quentinDupont:
Hello @SMaciasOSI and thanks for this work
If i understand well the code, your module permits to quickly apply sale price to a product thanks to the sum of the sale price of its BoM Components.
It means that you have BoMs composed of Products that you sale.

In my area (handmade food industry), it's the "opposite", you have BoM composed of Raw materials that you have to buy. So the components don't have sale price but purchase (cost) price.

I don't have issues with your approach but maybe you can describe this precisely in your readme ? :slightly_smiling_face: 

Here is my pending module that handle cost, sale and margin prices : https://github.com/OCA/manufacture/pull/1227
What do you think of that ?

SMaciasOSI:
> Hello @SMaciasOSI and thanks for this work If i understand well the code, your module permits to quickly apply sale price to a product thanks to the sum of the sale price of its BoM Components. It means that you have BoMs composed of Products that you sale.
> 
> In my area (handmade food industry), it's the "opposite", you have BoM composed of Raw materials that you have to buy. So the components don't have sale price but purchase (cost) price.
> 
> I don't have issues with your approach but maybe you can describe this precisely in your readme ? 🙂
> 
> Here is my pending module that handle cost, sale and margin prices : #1227 What do you think of that ?

Hello!

Yes, this module intends to allow the users to set the sales price for the manufactured product based on the components' sales price, which may be sold separately as well.

I've seen your module, it looks very useful, however the use case for this module is different. In this scenario, each specific component, since it can also be sold, has a different sales margin, so the sales margin for the manufactured product can't be extrapolated from the components as easily.

quentinDupont:
Yep ok ! Two modules for two very differents usages ! (don't hesitate to review mine ;-))
It's seems to be quite a draft as you are fixing stuffs
You can ping me when it's ready to review and test :)

quentinDupont:
> each specific component, since it can also be sold, has a different sales margin, so the sales margin for the manufactured product can't be extrapolated from the components as easily.

However, Manufactured product **cost** should be the sum of its component as well, with a margin = 0 so its sale price is equal to its cost price.

Do you have example for this use cases so it can be more clear in my mind ?

SMaciasOSI:
> Yep ok ! Two modules for two very differents usages ! (don't hesitate to review mine ;-)) It's seems to be quite a draft as you are fixing stuffs You can ping me when it's ready to review and test :)

This is ready to be reviewed now

SMaciasOSI:
> > each specific component, since it can also be sold, has a different sales margin, so the sales margin for the manufactured product can't be extrapolated from the components as easily.
> 
> However, Manufactured product **cost** should be the sum of its component as well, with a margin = 0 so its sale price is equal to its cost price.
> 
> Do you have example for this use cases so it can be more clear in my mind ?

The scenario is as follows

A manufacturing company sells manufactured products, but can also sell the components they use for manufacturing

Each component has a different sales margin

When manufacturing, the manufactured product can't be sold for less than the combined sales price of its components, but it can be sold for more

When a client sends an RFQ, a quotation engineering department has to evaluate what the sales price of the manufactured product should be. This is where this module comes into play. A BoM is generated by the design department, which is then sent to the quotation engineering folks, whose job is to set the sales price for the designed product.

PR #1430, opened 2024-12-20
[16.0][ADD] mrp_bom_simple_packaging_description

quentinDupont:
Simple module : adds a text field in BoM for packaging description.

image


mrp_bom_simple_packaging_description/views/view_mrp_bom.xml
```diff
@@ -0,0 +1,23 @@
...
+    <record id="view_mrp_bom_form" model="ir.ui.view">
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='product_uom_id']/.." position="after">
+              <field name="description_packaging" />
+            </xpath>
+        </field>
+    </record>
+
```
legalsylvain, line 22: you can maybe add this field in the tree view also, with optional="hide" ?

legalsylvain (approved):
Thanks ! 

A non blocking remark inline. otherwise, LGTM. Code review / no test.

mrp_bom_simple_packaging_description/views/view_mrp_bom.xml
quentinDupont: add the field on tree view on last commit, thanks !

legalsylvain (approved):

legalsylvain (approved):

LoisRForgeFlow (commented):
The field does not have any particular use? Just a plain text in the BoM form? Could you explain a bit the use case to understand better?

quentinDupont:
> The field does not have any particular use? Just a plain text in the BoM form? Could you explain a bit the use case to understand better?

In food area, you sometimes have a lot of quite similar BoM and product name. This field is a helper to distinguish quite similar BoM.
For example : a coffee maker will have theses products and according BoMs : 
- ground coffee Guatemala
- bean coffee Guatemala
- ground coffee Ethiopia
- bean coffee Ethiopia
Each products sold in 250gr, 500gr, 1kg, each BoM has a different component for the bag

This field permits to quickly see the weight of the product, search for it etc.

mihien:
@OCA/manufacturing-maintainers could this please be merged?

ivantodorovich:
> The field does not have any particular use? Just a plain text in the BoM form? Could you explain a bit the use case to understand better?

@LoisRForgeFlow can you update your review? is your question answered?

quentinDupont:
@mihien you can review and approve (or not) this PR :)

mihien (approved):
LGTM !

grindtildeath (commented):
Code looks ok, but if this field has nothing to do with the product.packaging model, could we avoid using "packaging" in field and module name?

I'd suggest to call the field  "description" and module name "mrp_bom_description" to make it easier for everyone to understand what it's all about

quentinDupont:
> Code looks ok, but if this field has nothing to do with the product.packaging model, could we avoid using "packaging" in field and module name?
> 
> I'd suggest to call the field "description" and module name "mrp_bom_description" to make it easier for everyone to understand what it's all about

I get your point. But i'm afraid that if this field become even more simpler and less accurate, it will become a field where users sets all kind of stuffs, and non structured informations become always useless or hard to work with.
Maybe we can juste rename the field, keeping the packaging meaning but with a other word to not be confusing with product.packaging, maybe `conditioning` ?

PR #1431, opened 2024-12-20
[16.0][MIG] mrp_production_back_to_draft: Migration to 16.0

giarve:


quentinDupont (approved):
Tested locally, works fine :heavy_check_mark: 
Code review

LoisRForgeFlow (approved):
LGTM

There are a couple of commits that could be squashed, but since this module is actually being backported, the noise of commits will never be moved to next version and will stay here in 16.0. Therefore I think we can merge.

/ocabot merge nobump

PR #1239, opened 2024-03-20
[16.0][ADD] mrp_bom_select_product_variant

quentinDupont:
New module for v16 that favors Product variant selection for BOM creation
When selecting a product variant, the linked product template is automatically selected.

Tree and filter view are also changed : you see and search with product variant.

mrp_bom_select_product_variant

PR #1435, opened 2025-01-01
[18.0][MIG] mrp_bom_select_product_variant: Migration to 18.0

BhaveshHeliconia:


quentinDupont (approved):
tested on runboat, migration code trivial :heavy_check_mark:

LoisRForgeFlow (changes requested):
I think this module is not needed. In v18, you cannot select a product variant in a BoM without selecting first the template, after which, only related variants are available for selection.

BhaveshHeliconia:
@LoisRForgeFlow, yes you are right, that's why i am closing this one

BhaveshHeliconia:
@LoisRForgeFlow, yes you are right, that's why i am closing this one

quentinDupont:
> I think this module is not needed. In v18, you cannot select a product variant in a BoM without selecting first the template, after which, only related variants are available for selection.

As discussed in the original PR here https://github.com/OCA/manufacture/pull/1239#issue-2197315403 it is exactly the behaviour this module tries to aim and change :smile:

LoisRForgeFlow:
@quentinDupont Thanks for clarifying

quentinDupont:
The aim of this module was to respond to the Odoo where the products are used as a priority over the product templates. In our case, users don't see the product templates and so we needed that in the BoM they could directly choose a product product.

LoisRForgeFlow:
Yes, it makes sense. The PR can be reopened and we can continue the migration

BhaveshHeliconia:
@LoisRForgeFlow Please review this PR when you have a moment. Thanks!

LoisRForgeFlow (approved):
LGTM

PR #1439, opened 2025-01-03
[17.0] [MIG] mrp_production_picking_type_from_route

MeritxellAForgeFlow:
Standard migration to 17.0

@ForgeFlow


mrp_production_picking_type_from_route/models/mrp_production.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    # Skip picking_type_id as a onchange trigger, even though the
+    # extended method is an onchange for picking_type_id too.
+    # If triggered also from picking_type_id changes, it would be impossible
+    # to adjust picking_type_id manually.
+    @api.onchange("product_id", "company_id")
```
JordiMForgeFlow, line 14:
```suggestion
    @api.onchange("product_id", "move_raw_ids", "company_id")
```
keep the move raw ids in the onchange decorator from the original method https://github.com/odoo/odoo/blob/e59261506f9258112475c7a7798cb2fd15e39b5b/addons/mrp/models/mrp_production.py#L816

ThiagoMForgeFlow (approved):
LGTM :+1:

JordiMForgeFlow (commented):
the changes in the module are no longer valid as the field picking_type_id is now computed: https://github.com/odoo/odoo/blob/e59261506f9258112475c7a7798cb2fd15e39b5b/addons/mrp/models/mrp_production.py#L277

We should apply the logic of the module in the compute method

AaronHForgeFlow (approved):
Code LGTM

AaronHForgeFlow (changes requested):
As per @JordiMForgeFlow comment

mrp_production_picking_type_from_route/models/mrp_production.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class MrpProduction(models.Model):
+    _inherit = "mrp.production"
+
+    # Skip picking_type_id as a onchange trigger, even though the
```
AaronHForgeFlow: I think this comment can be removed

AaronHForgeFlow (commented):
Code review LGTM.

BTW, It is strange the tests passed with the previous code. I am wondering if they are passed without the _compute_picking_type_id method. It you remove the method _compute_picking_type_id and the tests are passed then the tests need some fixing.

mrp_production_picking_type_from_route/models/mrp_production.py
MeritxellAForgeFlow: image

The test works correctly, when _compute_picking_type_id is removed, the test fails as seen in the image.

DavidJForgeFlow (approved):

LoisRForgeFlow (approved):
Functional and code review :+1:

PR #1448, opened 2025-01-10
[17.0][FIX] mrp_multi_level: fix main supplier computation

LoisRForgeFlow:
* api.depends

supply_method is not stored and therefore not reliable in the api.depends, including procure location in the api.depends cover changes in the supply method due to procure location change.

* remove usages of filtered

it was iterating the recordset more than 2 times when it could be done in a single interation.


JordiMForgeFlow (approved):
LGTM 👍🏼

PR #1457, opened 2025-01-21
[16.0][MIG] mrp_bom_attribute_match

Kosaaaaa:


LoisRForgeFlow (commented):
@Kosaaaaa could you check CI

mrp_bom_attribute_match/models/mrp_bom.py
```diff
@@ -0,0 +1,400 @@
...
+
+        if not component_template_id:
+            return line_product_id
+
+        # check attributes
+        if not {
+            component_template_id.valid_product_template_attribute_line_ids.attribute_id
+        }.issubset(
+            bom_product_id.valid_product_template_attribute_line_ids.attribute_id
+        ):
```
espo-tony, line 167: @Kosaaaaa This condition seems incorrect.
You’re not turning the contents into a Python set; you’re making a set with one element: the recordset object itself.
So the check becomes: “Is this single element (the recordset object) contained in the iterable on the right?" Which is always no, and not what you want here.

Please, fix it by processing `ids` lists (like you do at line 116) or change into something more Odoo like such as:

`        if not component_template_id.valid_product_template_attribute_line_ids.attribute_id <= bom_product_id.valid_product_template_attribute_line_ids.attribute_id:`

espo-tony (changes requested):
@Kosaaaaa 
Please, check comment

PR #1371, opened 2024-10-22
[17.0][MIG] mrp_bom_attribute_match

ferran-S73:
Sandard migration from v15

PR #1465, opened 2025-01-28
[17.0][MIG] mrp_account_bom_attribute_match: Migration to 17.0

arantxa-s73:
Module migrated to version 17.0

Depends on:

- [ mrp_bom_attribute_match ] https://github.com/OCA/manufacture/pull/1371


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-mrp_bom_attribute_match @ git+https://github.com/OCA/manufacture.git@refs/pull/1371/head#subdirectory=mrp_bom_attribute_match
```
LoisRForgeFlow: @arantxa-s73 Dependency has been merged, you can remove this commit now.

eugenios73:
@LoisRForgeFlow @ivantodorovich @mymage Can you review the PR after making these latest changes? Thanks in advance

mrp_account_bom_attribute_match/tests/test_mrp_account_bom_attribute_match.py
```diff
@@ -21,3 +23,58 @@ def test_bom_cost(self):
...
+                    ),
+                ],
+            }
+        )
+
+        skip_bom_line_iter = iter([True, False])
+        get_component_template_product_iter = iter([None])
+
+        def skip_bom_line_side_effect(self_):
+            # Devuelve el siguiente valor o False si se acaban
```
LoisRForgeFlow: Please, use english for comments

mrp_account_bom_attribute_match/tests/test_mrp_account_bom_attribute_match.py
```diff
@@ -21,3 +23,58 @@ def test_bom_cost(self):
...
+                        },
+                    ),
+                ],
+            }
+        )
+
+        skip_bom_line_iter = iter([True, False])
+        get_component_template_product_iter = iter([None])
+
+        def skip_bom_line_side_effect(self_):
```
LoisRForgeFlow: `self_`?

mrp_account_bom_attribute_match/tests/test_mrp_account_bom_attribute_match.py
```diff
@@ -21,3 +23,58 @@ def test_bom_cost(self):
         sword_magenta.button_bom_cost()
         self.assertEqual(sword_cyan.standard_price, 1.00)
         self.assertEqual(sword_magenta.standard_price, 2.00)
+
+    def test_bom_cost_template_skip_and_ignore(self):
```
LoisRForgeFlow, line 27: I don't understand this test, there is not a single assert so nothing is being actually tested.

eugenios73:
Hi @LoisRForgeFlow thanks for your comments. I made modifications and add assert in test.

The test verifies the behavior of a BOM cost calculation, especially when certain lines must be ignored or skipped based on specific conditions.

It uses mocks (`patch.object`) to force the behavior of the `_skip_bom_line` and `_get_component_template_product` methods:
* The first BOM line will be skipped (`_skip_bom_line` returns `True`).
* The second line is not skipped (`_skip_bom_line` returns `False`), and its component product is forced to `None` so that it is ignored.

It calls `_compute_bom_price` to calculate the cost, ensuring the code flow passes through the desired branches (skipping and ignoring lines as simulated).

This allows for covering and testing specific code branches related to the special handling of BOM lines.

The underscores (`_`, `__`) in the functions `definition skip_bom_line_side_effect(_)` and `skip_bom_line_side_effect(_, __, product_id)` are used as argument names to indicate that these parameters will not be used within the function. In Python, it is a convention to use `_` for variables or arguments whose value does not matter or will not be used. This allows the function to receive the same arguments as the original method (so that the mock works), while only using the one that is actually needed (`product_id` in this case).

mymage (commented):
I started testing and found an error when adding a product template in the bom, asking for the UoM.
It resolves adding a line with a product and then changing it adding a product template.
I tested it on 15.0 and goes without problem.

mymage:
I'm probably doing something wrong, but if I create a purchase order for a variant with an attribute value unavailable for the component, the component isn't added to the wo-compnonent-list. 
If this is confirmed, I think this isn't good behavior: you could "create" a product, losing one of its components.
Let me know if I'm wrong.
(is the same on v15.0 so isn't a porting problem)

eugenios73:
Hi @mymage, thanks for your review. Your comments are very interesting. I'll test these errors.

eugenios73:
> I started testing and found an error when adding a product template in the bom, asking for the UoM. It resolves adding a line with a product and then changing it adding a product template. I tested it on 15.0 and goes without problem. 

@mymage, following your comment, I've reviewed the code and confirmed exactly why this is happening in v17.0. You are indeed right, but it's not because of this PR. I've created a fix that solves this issue. You can review it and check that it works correctly with this fix. Let me know if you have any questions, and thank you very much for your review.

https://github.com/OCA/manufacture/pull/1579

mymage:
> Hi @mymage, thanks for your review. Your comments are very interesting. I'll test these errors.

You're welcome, let me know if you start working to solve the "lost component" problem.

eugenios73:
@mymage Hello again!

My team and I have been trying to reproduce what you reported, but we don't see this as a bug. From what we can tell, it seems to be the intended behavior of the module.

We think the key is to ensure the product, the Bill of Materials, and their behavior are all configured correctly before creating a manufacturing order. This applies whether the order is generated automatically from a purchase order or created manually.

If any other contributor would like to add their thoughts, @LoisRForgeFlow @ivantodorovich @legalsylvain or @rousseldenis we'd be happy to hear them to clear up any confusion. Otherwise, we don't believe this is the module where this functionality should be changed.

mymage:
> @mymage Hello again!
> 
> My team and I ...

Hmm, I'm not sure this is good behavior.
I know that in a perfect world all data are accurate and up-to-date, but in a real production management can happen that something is not set up in the right way.
So you configure product A with porperty P1 and P2; this product has the component CA with property P1 and P2 and the component CB with property P1 and P3.
You create a WO for A-P2 and the WO components list will be missing the CB because it's not configured (and related inventory reservation).
You can't see it because the BOM has dozen of components and you create dozen of WO (your business is doing well ;-)).
No one notices the unavailability of the CB-P2 (lead time 1 month), and when it is requested in the shop floor, production stops, the 1.000.000€ order is lost due to mandatory delivery date, and you are fired!
I would have liked to see at least a warning that a component was not compatible.
But let's wait and see what other more expert people think.

PS: luckily you are a good Odoo developer so don't worry if they fired you :-)

PS2: I don't know if it's better to move this discussion on a new issue.

mymage:
I retried but the error persist, we have to wait a sort of runboat rebuilt before the fix is included?

ferran-S73:
> > @mymage Hello again!
> > My team and I ...
> 
> Hmm, I'm not sure this is good behavior. I know that in a perfect world all data are accurate and up-to-date, but in a real production management can happen that something is not set up in the right way. So you configure product A with porperty P1 and P2; this product has the component CA with property P1 and P2 and the component CB with property P1 and P3. You create a WO for A-P2 and the WO components list will be missing the CB because it's not configured (and related inventory reservation). You can't see it because the BOM has dozen of components and you create dozen of WO (your business is doing well ;-)). No one notices the unavailability of the CB-P2 (lead time 1 month), and when it is requested in the shop floor, production stops, the 1.000.000€ order is lost due to mandatory delivery date, and you are fired! I would have liked to see at least a warning that a component was not compatible. But let's wait and see what other more expert people think.
> 
> PS: luckily you are a good Odoo developer so don't worry if they fired you :-)
> 
> PS2: I don't know if it's better to move this discussion on a new issue.

Hi @mymage , your issue isn't related at all with this module. It isn't even related with it's "base" module mep_bom_attribute_match. 
If I'm understanding correctly you're trying to say that the system should create a variant for the component with an attribute value that's not set in it's template. That is no standard Odoo behaviour so I don't think it should be implemented in neither of these two modules.

Feel free to open an issue/develop the fix you think is needed in a new PR. Thanks for your comments

mymage:
> Hi @mymage , your issue ...

Not create the variant but warning that it is not available, or better that a component can't be added because it hasn't that variant.
Opened https://github.com/OCA/manufacture/issues/1582

PR #1467, opened 2025-01-31
[16.0][FIX] mrp_subcontracting_skip_no_negative: fix creating backorder

AungKoKoLin1997:
This PR fixes the issue of creating a backorder when a picking contains both subcontracted and normal products and is validated.
Depends on https://github.com/odoo/odoo/pull/195958

@qrtl QT5016

### DIFF
--- a/mrp_subcontracting_skip_no_negative/models/stock_move.py
+++ b/mrp_subcontracting_skip_no_negative/models/stock_move.py
@@ -69,10 +69,10 @@ class StockMove(models.Model):
                             location=location.complete_name,
                         )
                     )
-        res = super(StockMove, self - moves_with_no_check)._action_done(
+        res = super(StockMove, moves_with_no_check)._action_done(
             cancel_backorder=cancel_backorder
         )
-        res += super(StockMove, moves_with_no_check)._action_done(
+        res += super(StockMove, self - moves_with_no_check)._action_done(
             cancel_backorder=cancel_backorder
         )
         return res
--- a/mrp_subcontracting_skip_no_negative/models/stock_picking.py
+++ b/mrp_subcontracting_skip_no_negative/models/stock_picking.py
@@ -20,3 +20,25 @@ class StockPicking(models.Model):
             ]
         )
         quants.check_negative_qty()
+
+    def _get_moves_to_backorder(self):
+        self.ensure_one()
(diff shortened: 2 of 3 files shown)


mrp_subcontracting_skip_no_negative/models/stock_move.py
```diff
@@ -69,10 +69,10 @@ def _action_done(self, cancel_backorder=False):
...
                         )
                     )
-        res = super(StockMove, self - moves_with_no_check)._action_done(
+        res = super(StockMove, moves_with_no_check)._action_done(
             cancel_backorder=cancel_backorder
         )
-        res += super(StockMove, moves_with_no_check)._action_done(
+        res += super(StockMove, self - moves_with_no_check)._action_done(
             cancel_backorder=cancel_backorder
         )
```
yostashiro, line 77: Correct me if I'm wrong, but I doubt that the proposed implementation works well when the receipt picking contains subcontracted and normal products, with a partial quantities processed for the normal product (i.e., there should be a backorder which contain both products for the remaining quantities). I think the current logic will generate two backorder pickings instead of one.

Instead, would it perhaps work if we simply process `_action_done()` once for all moves?

```py
        moves_with_check = self - moves_with_no_check
        res = super(
            StockMove, moves_with_check + moves_with_no_check
        )._action_done(cancel_backorder=cancel_backorder)
```

AungKoKoLin1997: > Instead, would it perhaps work if we simply process _action_done() once for all moves?

@yostashiro We just need to pass context for subcontract moves to skip the check. 
So, processing all moves at once will not be worked well.

AungKoKoLin1997: > We just need to pass context for subcontract moves to skip the check.
> So, processing all moves at once will not be worked well.

As we talked, context can only be passed for the recordset.
Can't pass different context for the record of recordset.

AungKoKoLin1997: > I think the current logic will generate two backorder pickings instead of one.

No, It will just create only one backorder.

LoisRForgeFlow (approved):
LGTM

PR #1469, opened 2025-02-03
[18.0][MIG] mrp_production_auto_validate: Migration to 18.0

chaule97:
- I think we don't need field `mo_auto_validation_warning`, we only show the warning `<div/>` tag when `mo_auto_validation = True`
- `mrp.immediate.production.line` was [removed](https://github.com/odoo/odoo/pull/108446), we have to handle `mrp.production.backorder` instead


mrp_production_auto_validate/models/mrp_production.py
```diff
@@ -12,36 +12,38 @@ class MrpProduction(models.Model):
...
-                return
-            if mo.bom_id and mo.auto_validate and not qty_ok:
-                raise exceptions.ValidationError(
-                    _(
-                        "The quantity to produce is restricted to {qty} "
-                        "as the BoM is configured with the "
-                        "'Order Auto Validation' option."
-                    ).format(qty=mo.bom_id.product_qty)
+
+                qty_ok = tools.float_is_zero(
```
ivantodorovich: Why not sticking to float_compare here, like it was before?

The result won't be exactly the same.. with `float_compare`, both values are rounded and then compared.
With your new suggestion, only their diff is rounded and checked against 0.

e.g.: for the sake of a dumb example, let's assume we are rounding to 0 decimals:

value1: 2.4
value2: 1.6

before: 
- `round(value1) == round(value2)`
- `round(2.4) == round(1.6)`
- `2 == 2`

now:
- `round(value1 - value2) == 0`
- `round(2.4 - 1.6) == 0`
- `round(0.8) == 0`
- `1 == 0`

mrp_production_auto_validate/models/mrp_bom.py
```diff
@@ -15,24 +15,9 @@ class MrpBom(models.Model):
...
-    @api.depends("mo_auto_validation")
-    def _compute_mo_auto_validation_warning(self):
-        for bom in self:
-            bom.mo_auto_validation_warning = False
-            if bom.mo_auto_validation:
-                bom.mo_auto_validation_warning = _(
-                    "The Quantity To Produce of an order is now "
-                    "restricted to the BoM Quantity."
-                )
+            self.mo_auto_validation = False
```
ivantodorovich, line 24: We could take the opportunity to convert this into a computed field with store=True, readonly=False and precompute=True

mrp_production_auto_validate/views/mrp_bom.xml
```diff
@@ -8,18 +8,16 @@
...
-                    nolabel="1"
-                    colspan="2"
+                <field name="mo_auto_validation" invisible="type != 'normal'" />
+                <div
                     class="alert alert-warning"
+                    colspan="2"
                     role="alert"
-                    attrs="{'invisible': ['|', ('type', '!=', 'normal'), ('mo_auto_validation', '=', False)]}"
-                />
+                    invisible="type != 'normal' or mo_auto_validation == False"
```
ivantodorovich: `or not mo_auto_validation`

ivantodorovich (approved):
Thanks

twalter-c2c (approved):

chaule97:
Ciao @simahawk , I fixed the issue, can you merge again?

PR #1470, opened 2025-02-04
[18.0][MIG] mrp_bom_location: Migration to 18.0

Kimkhoi3010:


mrp_bom_location/models/mrp_bom.py
```diff
@@ -0,0 +1,21 @@
...
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    location_id = fields.Many2one(string="Location", comodel_name="stock.location")
+
+    @api.onchange("picking_type_id")
+    def _onchange_picking_type_id(self):
+        if self.picking_type_id and self.picking_type_id.default_location_src_id:
+            self.location_id = self.picking_type_id.default_location_src_id
```
ivantodorovich: We should make this a computed field with `store=True, readonly=False`, IMO

mrp_bom_location/report/bom_structure.py
```diff
@@ -0,0 +1,89 @@
...
+            product=product,
+            line_qty=line_qty,
+            bom_line=bom_line,
+            level=level,
+            parent_bom=parent_bom,
+            index=index,
+            product_info=product_info,
+            ignore_stock=ignore_stock,
+        )
+        line_ids = self.env["mrp.bom.line"].search([("bom_id", "=", bom.id)])
```
ivantodorovich:
```suggestion
        line_ids = bom.bom_line_ids
```

I don't undertand this search. Why don't we just read the field?

ivantodorovich (commented):
Can you please take the opportunity to add some unit tests? 🙏🏻

mrp_bom_location/models/mrp_bom.py
ivantodorovich: The readonly=False is still missing, it seems.
Without it, it's only computed but users can't change it

mrp_bom_location/models/mrp_bom.py
```diff
@@ -0,0 +1,30 @@
...
+        store=True,
+    )
+
+    @api.depends("picking_type_id")
+    def _compute_location_id(self):
+        for record in self:
+            if (
+                record.picking_type_id
+                and record.picking_type_id.default_location_src_id
+            ):
```
ivantodorovich:
```suggestion
            if record.picking_type_id.default_location_src_id:
```

it's the same, but shorter

mrp_bom_location/models/mrp_bom.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class MrpBom(models.Model):
+    _inherit = "mrp.bom"
+
+    location_id = fields.Many2one(
+        string="Location",
+        comodel_name="stock.location",
+        compute="_compute_location_id",
+        store=True,
```
ivantodorovich, line 14: readonly=False missing

sebalix (approved):

ivantodorovich (approved):

mrp_bom_location/tests/test_mrp_bom.py
```diff
@@ -0,0 +1,48 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestMrpBomLocation(TransactionCase):
+    def setUp(self):
+        super().setUp()
```
twalter-c2c: I think you should use the `setUpClass` here instead.

mrp_bom_location/tests/test_mrp_bom.py
```diff
@@ -0,0 +1,48 @@
+from odoo.tests.common import TransactionCase
```
twalter-c2c: Is there a reason for not using the `BaseCommon` here?

twalter-c2c (approved):
Thank you!

marcelsavegnago (approved):

PR #1474, opened 2025-02-06
[16.0][ADD] quality_control_images_oca

edescalona:
@BinhexTeam

Allows you to add images to questions in quality control inspections.


pedrobaeza:
I don't think a new tab and module is needed. Just put the chatter in the standard module, and use the attachments features (including preview) for it.

edescalona:
> I don't think a new tab and module is needed. Just put the chatter in the standard module, and use the attachments features (including preview) for it.

Hi @pedrobaeza , it is no longer a new tab, it is the possibility of assigning an image for each response in the inspection, as long as said response is not satisfactory.

Maybe the qc.test and qc.inspection models already have the chatter assigned.

I see that the code was not modified correctly, I will correct it so that you can see the change, thanks

pedrobaeza:
Yes, and the module name shouldn't be in plural. Anyway, I think this can be interesting to be part of the main module.

edescalona:
> Yes, and the module name shouldn't be in plural. Anyway, I think this can be interesting to be part of the main module.

So what you are telling me is to make a PR to improve the original module?

pedrobaeza:
Yeah, I think so. Anyway, I think the mechanism for attaching images can be the many2many_binary (similar to the extra images in product for website_sale).

edescalona:
> Yeah, I think so. Anyway, I think the mechanism for attaching images can be the many2many_binary (similar to the extra images in product for website_sale).

Ok, fine, the idea is also to be able to use the image widget to make it more dynamic on mobile, and well, attaching documents with the chatter can be solved, that is, making an integration with it. I'm already looking into the possibility of making a PR to improve the original, thanks.

PR #1475, opened 2025-02-07
[16.0][IMP] quality_control_oca

edescalona:
@BinhexTeam

Added features:

1. Adding a new **Image** test type.


diggy128:
Very useful. It would indeed make the workflow cleaner.
Since this could be used as a metric to evaluate the cost of quality but cost of quality includes both cost of Good quality and cost of Bad quality, would it be possible to require an image regardless of whether the inspection failed?

edescalona:
Hi @diggy128  Sorry for the delay, improvements are being made to the image loading, when it is finished I will let you know, thanks

quality_control_oca/models/qc_inspection.py
```diff
@@ -345,7 +376,11 @@ def _compute_valid_values(self):
         help="UoM of the inspection value for a quantitative question.",
     )
     question_type = fields.Selection(
-        [("qualitative", "Qualitative"), ("quantitative", "Quantitative")],
+        [
+            ("qualitative", "Qualitative"),
+            ("quantitative", "Quantitative"),
+            ("image", "Image"),
+        ],
```
diggy128, line 386: I.m.o. image should be a separate field.
Universally, variable types fall in one of two categories: qualitative and category (aka qualitative). 
Adding image as a variable type here gives no context of whether the test passed or failed,  and furthermore, the lack of denumeration of the result (whether quantity or as reference to a qualitative value) makes statistical analysis for these tests imposible.

diggy128 (changes requested):

quality_control_oca/models/qc_inspection.py
edescalona: Hi @diggy128 , the image is in a separate field in the inspection lines, which is required if defined as Image type.

edescalona:
Ready for review

quality_control_oca/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 
 {
     "name": "Quality Control OCA",
-    "version": "16.0.1.3.0",
+    "version": "16.0.1.2.1",
```
christian-ramos-tecnativa: You don't have to change the version manually, the bot will do it after it get merged

christian-ramos-tecnativa (changes requested):

quality_control_oca/__manifest__.py
edescalona: Hi @Christian-RB ready for review

quality_control_oca/models/qc_inspection.py
```diff
@@ -354,3 +380,5 @@ def _compute_valid_values(self):
     success = fields.Boolean(
         compute="_compute_quality_test_check", string="Success?", store=True
     )
+    object_id = fields.Reference(related="inspection_id.object_id")
+    object_model = fields.Char(compute="_compute_object_model", store=True)
```
christian-ramos-tecnativa: What are these fields for? Are they really needed?

quality_control_oca/models/qc_inspection.py
```diff
@@ -16,8 +16,10 @@ class QcInspection(models.Model):
 
     @api.depends("inspection_lines", "inspection_lines.success")
     def _compute_success(self):
-        for i in self:
-            i.success = all([x.success for x in i.inspection_lines])
+        for inspection in self:
+            lines = inspection.inspection_lines
+            inspection.success = all([x.success for x in lines])
+            inspection.count_success = len(lines.filtered(lambda x: x.success is True))
```
christian-ramos-tecnativa: I think this may be a better approach
```suggestion
            inspection.success = all([x.success for x in lines])
            inspection.count_success = len(inspection.success)
```

christian-ramos-tecnativa (changes requested):

quality_control_oca/models/qc_inspection.py
edescalona: Hi @Christian-RB , in this case, the lines that are successful are counted to display in the kanban view of the inspections.

quality_control_oca/models/qc_inspection.py
edescalona: Hi @Christian-RB  the object_id field is to show in the form view of the lines which object is referenced, in case of selecting it, and the other if it could be deleted.

LoisRForgeFlow:
@edescalona could you rebase this? there are some conflicts now.

edescalona:
Hi @LoisRForgeFlow, I'll solve it and comment.

edescalona:
Hi @LoisRForgeFlow, the conflict resolved.

LoisRForgeFlow:
> Hi @LoisRForgeFlow, the conflict resolved.

@edescalona they are still not resolved.

image

Make sure that you are rebasing against the latest 16.0 commit.

edescalona:
Hi @LoisRForgeFlow, the conflict resolved.

image

LoisRForgeFlow:
@edescalona Thank you! now it is ready for review :+1:

edescalona:
Ready for review @LoisRForgeFlow, the conflict resolved..

LoisRForgeFlow:
@Christian-RB @diggy128 Could you update your reviews?

christian-ramos-tecnativa (approved):

diggy128:
I still think this is the wrong way to handle this, as I said so in my first review.

Check out PR#1646 for a very similar reasoning.

Image cannot be a variable type, but let's go with it for a minute.

1) You set a variable type as image.
2) You create an inspection based on the previous test
3) The QA employee looks at the inspection line: Type: Image

What would make this test pass / fail?
1) Attaching an image or not?
2) Attaching an image of the product or a kitten?
3) Attaching an image using the product or the QA eating a banana?

The list can go on...

Quality variables should be set so that getting the results is explicit and not taken on interpretation.

I.M.O. If you want to pass / fail a test based on an image it should work like this:

1) You add a boolean field in test.line requires_image (changes in this module)

2) You add a method to check if image is set when the test.line requires it

3) You create a test with a qualitative variable and add criteria (aka valid values) for passing or failing the test (E.g. Correct photo of the product -> pass
Wrong photo of the product -> fail) and check requires_photo.

4) You run the inspection. If you try to validate without uploading an image, you get a warning. If you upload an image you can select if the test passes or fails, using the EXPLICIT qualitative value.

This way the logic is true to SPC and analysis and your feature works for ANY check you may do in the image, without changing the code.

As a bonus, as I said in the PR mentioned above, you can get true statistical results and create KPIs that, with the current implementation is impossible.

Issue #1406, opened 2024-11-27
quality_control_stock_oca   trigger manufacturing


aviejo:
## Module
quality_control_stock_oca

## Describe the bug
If in a product I add a quality test for a stock.picking.type of manufacturing. And the stock.picking.type manufacturing types create mrp.production.  When will a test be created for a production trigger?

## To Reproduce
odoo 17

For a manufacturing product, I add a quality control for a manufacturing trigger. In your product sheet. I carried out a production for that product, and the control test is not associated with the new batch created.  In case of reception and delivery it works correctly.

PR #1478, opened 2025-02-13
[16.0][FIX] quality_control_oca

augustodinizl:
that change fix the error with the current set() as seen in #1406


augustodinizl:
closed as i knew its not right

PR #1481, opened 2025-02-18
[16.0][IMP] mrp_produt_characterisation Add groups mrp_user

quentinDupont:
Two improvements : 
- Theses fields are used for filtering so we hide field on product form and display them only for admin
- And for filter, this PR add groups to show them only for MRP user


legalsylvain (approved):

LoisRForgeFlow (approved):
/ocabot merge minor

PR #1493, opened 2025-03-06
[17.0][IMP]quality_control_stock_oca: inspection per lot/serial

manuelregidor:
This improvements adds an option to generate a quality inspection for each serial/lot number used in a stock.picking. This option has to be activated in the triggers related to stock picking operation types.

@sara-castello @HaraldPanten @ValentinVinagre 

T-7403

### DIFF
--- a/quality_control_stock_oca/models/qc_inspection.py
+++ b/quality_control_stock_oca/models/qc_inspection.py
@@ -47,8 +47,7 @@ class QcInspection(models.Model):
             if inspection.object_id._name == "stock.move":
                 inspection.lot_id = first(
                     move_lines.filtered(
-                        lambda x, inspection=inspection: x.move_id
-                        == inspection.object_id
+                        lambda x, ins=inspection: x.move_id == ins.object_id
                     )
                 ).lot_id
             elif inspection.object_id._name == "stock.lot":
--- a/quality_control_stock_oca/models/stock_picking.py
+++ b/quality_control_stock_oca/models/stock_picking.py
@@ -80,5 +80,20 @@ class StockPicking(models.Model):
                     )
                 )
             for trigger_line in _filter_trigger_lines(trigger_lines):
-                inspection_model._make_inspection(operation, trigger_line)
+                inspection = inspection_model._make_inspection(operation, trigger_line)
+                if operation.product_id.tracking != "none":
+                    move_line = operation.move_line_ids[0]
+                    if trigger_line.trigger.inspection_per_lot:
+                        move_lines = operation.move_line_ids
+                        for move_line in move_lines:
+                            inspection.write(
+                                {
+                                    "lot_id": move_line.lot_id and move_line.lot_id.id,
+                                    "qty": move_line.quantity,
+                                }
+                            )
+                            if move_line != move_lines[-1]:
+                                inspection = inspection.copy()
+                                inspection.set_test(trigger_line)
+                                inspection.action_todo()
         return res
--- a/quality_control_stock_oca/README.rst
+++ b/quality_control_stock_oca/README.rst
@@ -38,10 +38,21 @@ It also adds some shortcuts on picking and lots to these inspections.
 .. contents::
    :local:
 
+Configuration
+=============
+
+To create a quality inspection for each lot/serial number used in a
+stock move you need to:
+
+- Go to Quality Control> Configuration > Triggers.
+- Activate the "Inspection Per Lot" option in those triggers which need
+  to force a different inspection for each lot/serial number used in a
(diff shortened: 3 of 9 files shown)


sara-castello (approved):
Functional Review LGTM

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -80,5 +80,25 @@ def _action_done(self):
                     )
                 )
             for trigger_line in _filter_trigger_lines(trigger_lines):
-                inspection_model._make_inspection(operation, trigger_line)
+                inspection = inspection_model._make_inspection(operation, trigger_line)
```
Tisho99, line 83: Have you tought inheriting the _make_inspection() function with this code?

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -80,5 +80,25 @@ def _action_done(self):
...
+                if operation.product_id.tracking != "none":
+                    move_line = operation.move_line_ids[0]
+                    if not trigger_line.trigger.inspection_per_lot:
+                        inspection.write(
+                            {
+                                "lot_id": move_line.lot_id and move_line.lot_id.id,
+                            }
+                        )
```
Tisho99: Can  a operation have a first move line without lot and a second move line with lot?

Maybe it is better to write: {"lot_id": operation.mapped("move_line_ids.lot_id.id")[:1]}

Tisho99 (commented):
Some comments

quality_control_stock_oca/models/stock_picking.py
manuelregidor: Sorry, but I don't know what you mean. What's your suggestion for improving the _make_inspection method?

quality_control_stock_oca/models/stock_picking.py
manuelregidor: I think that all stock move lines related to products with tracking need a serial/lot number when validating the picking. This behaviour has not actually being modified from the previous version of the module.

quality_control_stock_oca/models/stock_picking.py
Tisho99: The idea is putting the new lot of the inspection inside the _make_inspection() function instead of the _action_done() function. 

I think it makes more sense and reduces the complexity of the _action_done inheritance

However it is a small detail, I approve the PR

Tisho99 (approved):

HaraldPanten:
@pedrobaeza @etobella Do you mind review merge this PR?

THX!

quality_control_stock_oca/models/qc_inspection.py
```diff
@@ -35,25 +32,6 @@ def _compute_picking(self):
             elif inspection.object_id._name == "stock.picking":
                 inspection.picking_id = inspection.object_id
 
-    @api.depends("object_id")
```
pedrobaeza: Why removing this?

manuelregidor: There's no point in computing the lot field anymore if an inspection for each lot/serial number used in a stock picking is created, as this value will be assigned when creating the inspection.

HaraldPanten: @pedrobaeza did you hace the opportunity to check Manuel's commets?

THX!

pedrobaeza: But that shouldn't be the default, and we need to keep backwards compatibility.

manuelregidor:
@pedrobaeza I've made some changes. Could you have a look? Thank you.

quality_control_stock_oca/models/qc_inspection.py
```diff
@@ -47,13 +39,20 @@ def _compute_lot(self):
...
                     move_lines.filtered(
-                        lambda x, inspection=inspection: x.move_id
-                        == inspection.object_id
+                        lambda x, ins=inspection: x.move_id == ins.object_id
                     )
                 ).lot_id
             elif inspection.object_id._name == "stock.lot":
                 inspection.lot_id = inspection.object_id
 
+    @api.depends("object_id")
```
etobella: REstore at the same point in order to avoid extra changes please

etobella (commented):
In general it is OK, I left a comment to minimize changes

quality_control_stock_oca/models/qc_inspection.py
manuelregidor: Done

ValentinVinagre (approved):

etobella (approved):

PR #1499, opened 2025-03-13
[17.0][IMP] mrp_mass_production_order: Mass production with one product consumed

peluko00:
This pr introduces the posibility to do the mass production orders with one product consumed for each line.

cc https://github.com/APSL 4231
@miquelalzanillas @javierobcn @mpascuall @BernatObrador @ppyczko @lbarry-apsl please review

### DIFF
--- a/mrp_mass_production_order/README.rst
+++ b/mrp_mass_production_order/README.rst
@@ -47,9 +47,12 @@ To use this module, you need to:
    products location.
 3. Check the produce all checkbox if the productions are already
    produced, otherwise the production will be confirmed
-4. Set the production tags if you want
-5. Select the products that will be produced with the quantity
-6. The manufacturing order will be created
+4. Check the with bom checkbox if the lines use BoMs for the consumed
+   products, otherwise you can select the product for each line to
+   consumed
+5. Set the production tags if you want
+6. Select the products that will be produced with the quantity
+7. The manufacturing order will be created
 
 Bug Tracker
 ===========
@@ -73,9 +76,9 @@ Authors
 Contributors
 ------------
 
-- [APSL-Nagarro](https://apsl.tech):
+-  [APSL-Nagarro](https://apsl.tech):
 
-  - Antoni Marroig <amarroig@apsl.net>
+   -  Antoni Marroig <amarroig@apsl.net>
 
 Maintainers
 -----------
--- a/mrp_mass_production_order/__manifest__.py
+++ b/mrp_mass_production_order/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "MRP Mass Production Order",
     "summary": "Create multiple manufacturing orders in one step",
-    "version": "17.0.1.0.1",
+    "version": "17.0.2.0.0",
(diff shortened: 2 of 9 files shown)


quentinDupont (commented):
Hello !

I'm not sure to understand.
It is already possible to launch productions with products without BoMs right ?

peluko00:
> Hello !
> 
> I'm not sure to understand. It is already possible to launch productions with products without BoMs right ?

Hi! Yes, it's possible

quentinDupont:
> > Hello !
> > I'm not sure to understand. It is already possible to launch productions with products without BoMs right ?
> 
> Hi! Yes, it's possible

So what's the value of your addings sorry i don't understand nor undesrtand it on runboat ? :slightly_smiling_face: can you explain it

peluko00:
> > > Hello !
> > > I'm not sure to understand. It is already possible to launch productions with products without BoMs right ?
> > 
> > 
> > Hi! Yes, it's possible
> 
> So what's the value of your addings sorry i don't understand nor undesrtand it on runboat ? 🙂 can you explain it

It's for mass production orders to select the product to consumed

peluko00:
Can you review again @quentinDupont please?

quentinDupont:
OK so i get it now ! The UX is not very obvious but i think I lack context : do you have example of process where it's used ?

It's weird because you can choose a product_consumed_id which is not on any BoMs of product_id.

peluko00:
> OK so i get it now ! The UX is not very obvious but i think I lack context : do you have example of process where it's used ?
> 
> It's weird because you can choose a product_consumed_id which is not on any BoMs of product_id.

The workflow is if you want to use BoMs you will select with_bom option and the boms will compute for every line that you have otherwise you only need to put in the line the product that you want to produce and the product that consumes that.

quentinDupont:
> > OK so i get it now ! The UX is not very obvious but i think I lack context : do you have example of process where it's used ?
> > It's weird because you can choose a product_consumed_id which is not on any BoMs of product_id.
> 
> The workflow is if you want to use BoMs you will select with_bom option and the boms will compute for every line that you have otherwise you only need to put in the line the product that you want to produce and the product that consumes that.

In which field of activity it is used ? Can you add a description process on readme maybe ?

Also can you be more explicit on field name, for example 
1 : Manufactured Product
2 : Consumed Product

image

peluko00:
> > > OK so i get it now ! The UX is not very obvious but i think I lack context : do you have example of process where it's used ?
> > > It's weird because you can choose a product_consumed_id which is not on any BoMs of product_id.
> > 
> > 
> > The workflow is if you want to use BoMs you will select with_bom option and the boms will compute for every line that you have otherwise you only need to put in the line the product that you want to produce and the product that consumes that.
> 
> In which field of activity it is used ? Can you add a description process on readme maybe ?
> 
> Also can you be more explicit on field name, for example 1 : Manufactured Product 2 : Consumed Product
> 
> image

I changed the fields description and i added more info to the README.

quentinDupont (approved):
Thanks for all the changes !!

PR #1500, opened 2025-03-14
[16.0][IMP] mrp_attachment_mgmt: BoM attachments management

MarinaAForgeFlow:
Currently the module is managing the attachments by making the product attachments accessible in different  mrp views. With this PR the same is done with the BoM attachments.

This can be useful in scenarios where you have a product with multiple bill of materials, and the product has common drawings but also each BoM has specific drawings and documents. Therefore, you may want to have accessible in the different mrp views both, the common product attachments and the BoM specific attachments.

Also the naming of the already existing smart-buttons has been updated, to better differentiate the two buttons in each view.

### DIFF
--- a/mrp_attachment_mgmt/models/mrp_bom.py
+++ b/mrp_attachment_mgmt/models/mrp_bom.py
@@ -2,6 +2,7 @@
 # Copyright 2023 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import api, models
+from odoo.tools.safe_eval import safe_eval
 
 
 class MrpBom(models.Model):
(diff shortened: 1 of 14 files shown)


AaronHForgeFlow (approved):
code review LGTM

JordiMForgeFlow (approved):
code review

mrp_attachment_mgmt/models/mrp_bom.py
```diff
@@ -31,3 +32,22 @@ def action_see_bom_documents(self):
...
+        """Returns the action to show the attachments linked to the bom record."""
+        domain = [
+            ("res_model", "=", "mrp.bom"),
+            ("res_id", "in", self.ids),
+        ]
+        action = self.env["ir.actions.actions"]._for_xml_id("base.action_attachment")
+        context = action.get("context", "{}")
+        context = safe_eval(context)
+        context["create"] = False
+        context["edit"] = False
```
MarinaAForgeFlow, line 51: The context is being updated to disable creation and edition because the purpose of this action is to give visibility to the bom attachments. I believe it would be confusing for the user to try to add a new attachment in this view and see that the attachment is not actually linked to any record and does not appear in the BoM attachment.  Maybe we can leave the edition but disable the creation. What are your thoughts on this?

Also, I am wondering if the same should be done for the product action to keep it coherent, and if so, if it should be in this same PR in a different commit or in a different PR.

JordiMForgeFlow: @MarinaAForgeFlow keep it in a separated commit, but I agree on disable creation and edition.

LoisRForgeFlow:
@victoralmau Is this OK for you?

victoralmau (approved):
Code review OK

LoisRForgeFlow (approved):
/ocabot merge minor

PR #1503, opened 2025-03-20
[18.0][MIG] mrp_subcontracting_inhibit: Migration to 18.0

xaviedoanhduy:


wlin-kencove:
Going through functional testing:

### Purchase order step after subcontracting product setup:
image

Seems correct but after confirming I do not see a production order (Manufacturing order) created 
image


This does not match what is stated in step 10 of the readme:
```
Purchase order flow: #. Go to *Purchase > Orders > Requests for
Quotation* and create new order as follows: #. Vendor: Azure Interior,
Product: Subcontract product #. The unit price of the product will be
10. #. Click on the 'Confirm Order' button. #. A production order will
have been created.
```

Is this a bug or did I misunderstand the instructions?

wlin-kencove:
Found an interesting behavior:

You can see them under Inventory->Operations->Manufacturings
image


You however cannot see them under: Manufacturing -> Operations -> Manufacturing Orders even after removing the filters. 
image


I am guessing an implicit filter is hiding it from the Manufacturing Orders View
This is strange since both seem to be  using mrp.mrp_production_tree_view and pulling from the same model

This could lead to some confusion depending which form the user is looking at.

wlin-kencove:
Replenishment of subcontracting product workflow seems to work fine:
image

The visibility of the manufacturing order different per view was the only issue I found.

xaviedoanhduy:
hello @wlin-kencove,
yes, u r right!
the production order was created and managed from menu` Inventory > Operations > Transfer > Manufacturings`
maybe in the menu `Manufacturing -> Operations -> Manufacturing Orders` they should only show production orders instead of subcontracting orders
but can you agree with me that the module still works as expected?

wlin-kencove:
Looks like the cause of the issue was that the subcontracting type was archived. Would mean there are differences between the  Inventory > Operations > Transfer > Manufacturings form and Manufacturing -> Operations -> Manufacturing Orders since one form will explicitly filter archived manufacturing types while the other does not. 

As far as core functionality goes, it is working.

wlin-kencove (approved):

LoisRForgeFlow (approved):
LGTM :+1:

PR #1504, opened 2025-03-20
[16.0][IMP]quality_control_stock_oca: adding default_object_id on context

adasatorres:
Good morning, I suggest this change as it makes it easier to create a quality control from the delivery order since it will pass the object reference by default.

### DIFF
--- a/quality_control_stock_oca/__manifest__.py
+++ b/quality_control_stock_oca/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "Quality control - Stock (OCA)",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
     "category": "Quality control",
     "license": "AGPL-3",
     "author": "OdooMRP team, AvanzOSC, Serv. Tecnol. Avanzados - Pedro M. Baeza, "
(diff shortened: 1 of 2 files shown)


AlvaroRM11 (approved):
LGTM

rvalyi (approved):

AinohaBH (approved):
LGTM!!

diggy128:
LGTM!, although it should be used with caution as active_id is only propagated correctly to the first level in

quality_control_stock_oca/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Quality control - Stock (OCA)",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
```
LoisRForgeFlow, line 8: Next time do not bump manually the version if you are not adding migration scripts, OCAbot will take care of it!

LoisRForgeFlow (approved):
/ocabot merge nobump (because version was manually bumped)

quality_control_stock_oca/__manifest__.py
adasatorres: Ok.

PR #1506, opened 2025-03-25
[18.0][MIG] mrp_unbuild_subcontracting: Migration to 18.0

xaviedoanhduy:
- `_onchange_picking_id` was removed in https://github.com/odoo/odoo/pull/118568


mrp_unbuild_subcontracting/tests/test_mrp_unbuild_subcontracting.py
```diff
@@ -195,21 +197,14 @@ def test_purchase_partial_receipt_and_refund(self):
...
-                {
-                    "location_id": picking_to_return.location_id.id,
-                    "picking_id": picking_to_return.id,
-                }
+        return_form = Form(
+            self.env["stock.return.picking"].with_context(
+                active_id=receipt.id, active_model="stock.picking"
             )
         )
-        return_wizard._onchange_picking_id()
```
kobros-tech: Was that code necessary to change?

kobros-tech (changes requested):
code review only, it is fine in general but I inquire about some changes in the code of test cases.

mrp_unbuild_subcontracting/tests/test_mrp_unbuild_subcontracting.py
xaviedoanhduy: hi @kobros-tech, yes it is indeed necessary, maybe I was a bit subjective not to mention it in the PR description because this change has been applied since 17.0, see https://github.com/odoo/odoo/pull/118568 - but maybe it would have been better if I explicitly mentioned it in the description (updated)

xaviedoanhduy:
hello @LoisRForgeFlow, I think you'll be interested in this module - could you take a look at it?

PR #1507, opened 2025-03-27
[17.0][FIX] mrp_production_serial_matrix: increase line_ids limit

ArnauCForgeFlow:
When generating a matrix with more than 20 finished_lot_ids, only the first 20 were displayed. 

As we can see we select 30 finished_lot_ids
Screenshot from 2025-03-27 16-05-57

And only the first 20 lots are displayed
Screenshot from 2025-03-27 16-06-30


AlexPForgeFlow (approved):
LGTM! :+1:

LoisRForgeFlow (approved):
Thanks :+1:

PR #1514, opened 2025-04-09
[16.0][ADD] mrp_workorder_priority

LauraCForgeFlow:
This module adds a priority field to the work orders, which is taken from their related manufacturing order.


mrp_workorder_priority/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025 ForgeFlow, S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
JordiMForgeFlow:
```suggestion
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
to be changed in other files also

LauraCForgeFlow: Done! :)

AaronHForgeFlow:
Should we change the sequence when a workorder is higher priority? Perhaps that can be done on a separate module, do you agree?

JordiMForgeFlow (approved):

JordiMForgeFlow:
@AaronHForgeFlow I am not sure if changing the sequence could then mess up with the MO execution, taking a different or wrong order for the WO. Could be something to evaluate though :)

AaronHForgeFlow (approved):
ok! fair enough. LGTM, code + functional

JasminSForgeFlow (approved):
Code Review!

JoanSForgeFlow (approved):
LGTM!

PR #1013, opened 2023-04-04
[16.0][ADD] mrp_stock_owner_restriction

AungKoKoLin1997:
This module adopts the functionality of stock_owner_restriction in the manufacturing order. This module will assign the owner for the produced product.
Related - https://github.com/OCA/stock-logistics-workflow/pull/1231

@qrtl

PR #1518, opened 2025-04-12
[18.0][MIG] mrp_stock_owner_restriction

bosd:
Standard migration using odoo-module-migrator /ocabot migration mrp_stock_owner_restriction .

Migration of:  #1013

@qrtl


bosd:
This module needs fixing: https://github.com/OCA/manufacture/issues/1522

mrp_stock_owner_restriction/models/mrp_production.py
```diff
@@ -11,12 +11,24 @@ class MrpProduction(models.Model):
...
+    def create(self, vals_list):
+        """Override create to ensure owner_id is set correctly."""
+        records = super().create(vals_list)
+        for record in records:
+            if (
+                record.picking_type_id
+                and record.picking_type_id.owner_restriction == "unassigned_owner"
+                and not record.owner_id
+            ):
+                record.owner_id = self.env.company.partner_id
```
AungKoKoLin1997: I don't get the point of assigning company's partner for `unassigned_owner` picking type.
Can you explain me?

bosd:
please keep open

BhaveshHeliconia (approved):
Code review LGTM!

mrp_stock_owner_restriction/models/mrp_production.py
AungKoKoLin1997: @bosd Can you explain the reason of this changes?

bosd: @AungKoKoLin1997 you're right, and thanks for insisting. There is no good reason to assign the company partner there — it was actually wrong.

`unassigned_owner` means the picking type reserves from **unowned** stock (`stock_owner_restriction` forces `force_restricted_owner_id=False` in `_action_assign`). Setting `owner_id = company.partner_id` does the opposite: it restricts reservation to company-partner-owned quants and marks the produced quants as owned. The branch was also untested (the assertion in `test_04` was commented out).

I've removed that branch (folded into the post-migration fix commit). `create()` now only carries over the owner shared by the source MOs during a merge; for `unassigned_owner` the MO owner correctly stays empty.

bosd:
to fix tthe tests please see #1817

flotho (approved):
Hi @bosd , 
A great thanks for this PR !
Full review on runboat,
Works as expected (y)

carlosdauden:
Reviewing this as suggested in OCA/stock-logistics-workflow#2417 — I had opened a
module with overlapping scope there, this one clearly precedes it, so I am
closing mine and bringing what I found here instead.

Three points. The first two are about code I could not resolve against the
current base module, so I may well be missing something.

### 1. `_get_owner_restriction()` does not seem to exist in `stock_owner_restriction`

In `models/stock_move.py`:

```python
def _get_owner_restriction(self):
    self.ensure_one()
    mo_to_unbuild = self._get_mo_to_unbuild()
    if mo_to_unbuild:
        return mo_to_unbuild.owner_restriction
    return super()._get_owner_restriction()
```

I could not find that method in `stock_owner_restriction` on 18.0, 17.0 or 16.0,
nor an open PR adding it. If that is right, the `super()` call raises
`AttributeError` for any move that gets there, which is the unbuild path. The
tests here cover the merge scenarios and the owner assignment but not unbuild, so
CI would stay green either way.

Is this expecting a companion PR on the base module?

### 2. `_compute_restrict_partner_id` looks like dead code

```python
# Just to add a trigger
@api.depends("production_id.owner_id")
def _compute_restrict_partner_id(self):
    return super()._compute_restrict_partner_id()
```

In core, `restrict_partner_id` is a plain `Many2one` with no `compute`
(`stock/models/stock_move.py`), and `restrict_partner_id` does not appear at all
in `stock_owner_restriction` — 0 occurrences on both 18.0 and 16.0. So nothing
should ever call this, and the `super()` would fail if something did.

What the base module does have is a similarly named
`_compute_restricted_owner_id`, computing its `restricted_owner_id` helper field
from `_get_owner_for_assign()`. Was this meant to hook that one, so the helper
picks up `production_id.owner_id` as a dependency?

### 3. The finished product takes the declared owner without looking at what was consumed

`stock_move_line._action_done` assigns `move_id.production_id.owner_id` to the
finished lines. That is the right idea, and worth spelling out because it is not
obvious: Odoo declines to value the consumption of goods belonging to a partner
(`stock.move._is_out()` skips any line carrying an `owner_id`) and then values
the finished product anyway, at whatever the components cost, which by then is
nothing. Naming the owner on the way out is what stops the order from booking
stock out of thin air.

What it does not check is whether the components actually came from that owner.
An MO with `owner_id` set that consumes the company's own stock still hands the
output to the partner, unvalued, so that material leaves the company's valuation
with nothing recording it.

In the module I was proposing I refused to validate in that case, raising when
the consumed lines come from more than one source — a partner plus the company,
or two partners — on the grounds that there is no way to express partial
ownership of the output. Happy to send that as a follow-up PR here if you find it
useful.

I can test the migration on 18.0 once the first point is clarified.

bosd:
Thanks @carlosdauden, this was a very useful review — you were right on both counts, and the second one uncovered a third bug. All addressed in 5678bfe.

### 1. `_get_owner_restriction()`

Confirmed: no such method in `stock_owner_restriction` (18.0/17.0/16.0) nor in core `stock`/`mrp`, and nothing in this module calls it either — `mrp.unbuild.action_validate()` reads `mo_id.owner_restriction` directly. It was dead code that could only ever raise, no companion PR expected. Removed.

### 2. `_compute_restrict_partner_id()`

Also confirmed: `restrict_partner_id` is a plain `Many2one` in `stock/models/stock_move.py` on 18.0, no compute anywhere. Your reading of the intent is the right one — it is now an override of the base module's `_compute_restricted_owner_id()`, declaring the manufacturing dependencies that the field's owner resolution actually goes through:

```python
@api.depends(
    "raw_material_production_id.owner_id",
    "production_id",
    "move_dest_ids.raw_material_production_id.owner_id",
    "consume_unbuild_id.mo_id.owner_id",
    "unbuild_id.mo_id.owner_id",
)
def _compute_restricted_owner_id(self):
    return super()._compute_restricted_owner_id()
```

Hooking that compute surfaced a latent crash: `_get_owner_for_assign()` returned the bool `False` for finished moves, and the base compute does `move._get_owner_for_assign()[:1]` — `False[:1]` is a `TypeError`. It now returns an empty recordset.

And a second one, which the new coverage caught: `mrp.production.write()` released the reservation through `move_line_raw_ids`, a field that no longer exists on `mrp.production` in 18.0. Assigning an owner to an MO raised `AttributeError`. It now uses `move_raw_ids._do_unreserve()`, skipping moves already done or cancelled.

### 3. Finished product vs. what was actually consumed

Agreed, and thank you for spelling out the valuation mechanics — that reasoning was nowhere in the module. I have documented the gap in `readme/ROADMAP.md` for now rather than changing the behaviour in a migration PR:

> The owner declared on the manufacturing order is assigned to the finished product without checking who owned the components that were actually consumed. An order carrying an owner that consumes the company's own stock (or stock belonging to another partner) still hands the output to that owner, so the consumed material leaves the company valuation with nothing recording it. Validating such an order is not blocked yet.

Your follow-up PR refusing to validate when the consumed lines come from more than one source would be very welcome on top of this. Blocking seems right to me — there is indeed no way to express partial ownership of the output.

### Coverage

`codecov/patch` was at 84.58%. The module is now at 100% statement and branch coverage locally (run against `ghcr.io/oca/oca-ci/py3.10-odoo18.0`), with tests added for the unbuild path, the reservation released when an owner is assigned, and the owner resolution of chained component moves. The `action_merge()`/`create()` pair was also simplified — the owner is carried over through one context key, and the debug logging is gone.

DorianMAG:
@bosd 
Can you rebase plz?
Regards

PR #1519, opened 2025-04-15
[16.0][IMP] mrp_unbuild_restore_origin

nobuQuartile:
- Clarified help text for restore_rm_stock_in_origin_loc
- Hide location_dest_id when restore_rm_stock_in_origin_loc is enabled

@qrtl
QT5275

### DIFF
--- a/mrp_unbuild_restore_origin/models/mrp_unbuild.py
+++ b/mrp_unbuild_restore_origin/models/mrp_unbuild.py
@@ -8,9 +8,9 @@ class MrpUnbuild(models.Model):
     _inherit = "mrp.unbuild"
 
     restore_rm_stock_in_origin_loc = fields.Boolean(
-        string="Restore Raw Materials to Consumed Location",
-        help="If enabled, the source location of the component stock move lines will be used"
-        " as the destination location for the unbuild move lines.",
+        string="Return Components to Original Location",
+        help="If selected, components will be returned to the same location"
+        " they were taken from during manufacturing.",
     )
 
     def action_unbuild(self):
--- a/mrp_unbuild_restore_origin/views/mrp_unbuild_views.xml
+++ b/mrp_unbuild_restore_origin/views/mrp_unbuild_views.xml
@@ -5,11 +5,16 @@
         <field name="model">mrp.unbuild</field>
         <field name="inherit_id" ref="mrp.mrp_unbuild_form_view" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='mo_id']" position="after">
+            <xpath expr="//field[@name='location_dest_id']" position="before">
                 <field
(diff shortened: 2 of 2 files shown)


mrp_unbuild_restore_origin/views/mrp_unbuild_views.xml
yostashiro, line 1: Can we move `restore_rm_stock_in_origin_loc` to before `location_dest_id`?

mrp_unbuild_restore_origin/models/mrp_unbuild.py
```diff
@@ -8,9 +8,9 @@ class MrpUnbuild(models.Model):
     _inherit = "mrp.unbuild"
 
     restore_rm_stock_in_origin_loc = fields.Boolean(
-        string="Restore Raw Materials to Consumed Location",
-        help="If enabled, the source location of the component stock move lines will be used"
-        " as the destination location for the unbuild move lines.",
+        string="Return to Consumption Location",
```
yostashiro:
```suggestion
        string="Return Components to Original Location",
```

mrp_unbuild_restore_origin/models/mrp_unbuild.py
```diff
@@ -8,9 +8,9 @@ class MrpUnbuild(models.Model):
...
+        help="If enabled, raw materials will be returned to the same location"
```
yostashiro:
```suggestion
        help="If selected, components will be returned to the same location"
```

mrp_unbuild_restore_origin/models/mrp_unbuild.py
nobuQuartile: Done

mrp_unbuild_restore_origin/models/mrp_unbuild.py
nobuQuartile: Done

mrp_unbuild_restore_origin/views/mrp_unbuild_views.xml
nobuQuartile: That might actually be a better idea.
Done

yostashiro:
The test errors come from mrp_multi_level and I suspect the timezone assignment to the calendar here https://github.com/OCA/manufacture/blob/d5d0990/mrp_multi_level/tests/test_mrp_multi_level.py#L352 may be affecting the subsequent tests, although I haven't dug into the details. In that case, the tests should pass if we rerun the CI now...

yostashiro (approved):
LGTM.

AungKoKoLin1997 (approved):
LG

LoisRForgeFlow (approved):
/ocabot merge minor

PR #1530, opened 2025-05-12
[15.0][ADD] mrp_production_generator_by_date_interval: New module to create multiple production orders with a wizard

CarlosRoca13:
cc @Tecnativa TT56244

Todos:
- [x] Wizard to create multiple production orders for a given period and a product or BoM.
- [x] Add test

please @pedrobaeza @pilarvargas-tecnativa


mrp_production_generator/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+1. Go to *Manufacturing > Operations > Production Generator*
```
pedrobaeza: Shouldn't this be in Manufacturing > Planning instead?

pedrobaeza (approved):
Tested on runboat. Little comment. Put also in ROADMAP that this doesn't take into account MOs spawning more than one day.

CarlosRoca13:
Done @pedrobaeza

LoisRForgeFlow (commented):
Have you seen [`mrp_mass_production_order`](https://github.com/OCA/manufacture/tree/18.0/mrp_mass_production_order)? Do you think this module can be an improvement for mrp_mass_production_order?

pedrobaeza:
Hi @LoisRForgeFlow, I have checked the module you mention, and first of all, it has a dependency over a module `mrp_tag` that is not logic for our use case (and for the other should be a glue module, not in the main one, but the reviewers didn't stop it). Second, this module is not on version 15 and it doesn't worth to backport it and add features. But finally, the goal is totally different, as the other is for generating production orders from the lines you manually enter with several products, quantities, etc. This one is for generating MOs for the same product across time.

pilarvargas-tecnativa (approved):

quentinDupont:
Thanks for this module!
In order to differentiate it more explicitly from `mrp_mass_production_order`, could the name be `mrp_production_generator_per_bom`?

LoisRForgeFlow:
> Hi @LoisRForgeFlow, I have checked the module you mention, and first of all, it has a dependency over a module `mrp_tag` that is not logic for our use case (and for the other should be a glue module, not in the main one, but the reviewers didn't stop it). Second, this module is not on version 15 and it doesn't worth to backport it and add features. But finally, the goal is totally different, as the other is for generating production orders from the lines you manually enter with several products, quantities, etc. This one is for generating MOs for the same product across time.

@pedrobaeza Thanks for evaluating and clarifying, the differences are clearer now. Therefore, I have nothing against continuing this PR as a new module.

pedrobaeza:
@quentinDupont looking for a more specific name, I would go with `mrp_production_generator_by_date_interval`. What do you think?

quentinDupont:
> @quentinDupont looking for a more specific name, I would go with `mrp_production_generator_by_date_interval`. What do you think?

Very nice ! Very specific and therefore easy to understand :ok_hand:

CarlosRoca13:
Module renamed!

pedrobaeza:
Merging thus

/ocabot merge nobump

PR #1534, opened 2025-05-18
[16.0][FIX] mrp_bom_tracking: catch UoM qty mrp_bom_tracking rounding errors

giarve:
Avoid this:

image

and this:

image


LoisRForgeFlow (approved):
Thanks!

PR #1538, opened 2025-05-29
[14.0][IMP][mrp_bom_attribute_match] improve missing attribute message

Kev-Roche:
Improve missing attribute message by given the names of the missing attributes

### DIFF
--- a/mrp_bom_attribute_match/models/mrp_bom.py
+++ b/mrp_bom_attribute_match/models/mrp_bom.py
@@ -123,11 +123,13 @@ class MrpBomLine(models.Model):
                         rec.component_template_id.display_name,
                     )
                 )
-            if not all(attr in prod_attrs for attr in comp_attrs):
+            missing_attrs = comp_attrs - prod_attrs
+            if missing_attrs:
                 raise ValidationError(
                     _(
                         "Some attributes of the dynamic component are not included into "
-                        "production product attributes."
+                        "production product attributes. Missing attributes: %s",
+                        ", ".join(missing_attrs.mapped("name")),
                     )
                 )
 


mrp_bom_attribute_match/models/mrp_bom.py
```diff
@@ -124,11 +124,15 @@ def _check_component_attributes(self):
                     )
                 )
             if not all(attr in prod_attrs for attr in comp_attrs):
+                missing_attrs = rec.env["product.attribute"].browse(
+                    [attr.id for attr in set(prod_attrs) - set(comp_attrs)]
+                )
```
ivantodorovich: nitpicking: simpler (and opposite)

```suggestion
            missing_attrs = comp_attrs - prod_attrs
            if missing_attrs:
```

mrp_bom_attribute_match/models/mrp_bom.py
```diff
@@ -124,11 +124,15 @@ def _check_component_attributes(self):
...
                 raise ValidationError(
                     _(
                         "Some attributes of the dynamic component are not included into "
-                        "production product attributes."
+                        "production product attributes. Missing attributes: %s"
                     )
+                    % ", ".join(missing_attrs.mapped("name"))
```
ivantodorovich: nitpicking : use gettext built-in string interpolation

```suggestion
                        "production product attributes. Missing attributes: %s",
                        ", ".join(missing_attrs.mapped("name"))
                    )
```

mrp_bom_attribute_match/models/mrp_bom.py
```diff
@@ -124,11 +124,15 @@ def _check_component_attributes(self):
                     )
                 )
             if not all(attr in prod_attrs for attr in comp_attrs):
+                missing_attrs = rec.env["product.attribute"].browse(
+                    [attr.id for attr in set(prod_attrs) - set(comp_attrs)]
```
ivantodorovich: Are you sure it is correct?
Seems like `missing_attrs` is actually product attributes that have no component, the opposite of what the above if-condition is checking

ivantodorovich (changes requested):

mrp_bom_attribute_match/models/mrp_bom.py
Kev-Roche: You are right, I fix it and simplify the code with your suggestions.

ivantodorovich (approved):
Thanks!

LoisRForgeFlow (approved):
:+1: 

/ocabot merge patch

LoisRForgeFlow:
@Kev-Roche could you take care of porting this to newer versions?

PR #1545, opened 2025-06-04
[16.0][FIX] quality_control_stock_oca: qc_inspections_ids access from picking

LauraCForgeFlow:


AaronHForgeFlow (approved):
ok!

AndreuOForgeFlow (approved):

LoisRForgeFlow (approved):
Discussed with Laura, the issue behind this is that users with no quality access could not process pickings even with zero inspections related.

LoisRForgeFlow:
@LauraCForgeFlow what about v17? Could you fwport it?

PR #1566, opened 2025-07-17
[18.0][FIX] mrp_bom_location

DavidJForgeFlow:
There where missing some parameters in the mehtod _get_bom_data in the report. This was causing to loose this parameters in the inherited chain and making it to fail if any other module inherits this method, for example ddmrp.


LoisRForgeFlow (approved):
:+1:

Issue #1378, opened 2024-10-29
quality_control_oca: Broken quality control reference related-type tests


giarve:
## Module
quality_control_oca

## To Reproduce
**Affected versions**: 16.0

1. Create a new test
2. Select type: Related
3. Enjoy a blank list

StuartJMackintosh:
+1

cparadis-impressfoods:
I've posted a similar issue #1286 a few months ago. There was also #823 dating back to V14 with a similar bug. I just retried the module on a new dev database and experiencing the same issue. As per my bug report, I was able to reproduce the bug on runboat for versions going back to V14. 
I've received no response on my initial issue.

aviejo:
Conceptually, what is that field for? What should it contain?

cparadis-impressfoods:
> Conceptually, what is that field for? What should it contain?

It should contain the object reference for related fields. Right now
![image](https://github.com/user-attachments/assets/b13ff359-060f-431d-b684-7a00b7fa9267)
Currently, the selection values for this field are ['s','e','t','(',')'], which strangely is the string representation of the return value of the object_selection_values() method of qc.test: 
https://github.com/OCA/manufacture/blob/171badef793a255a8d5763f55f1be19a5a87c2f0/quality_control_oca/models/qc_test.py#L21C2-L21C3
Since it's a related field, I would expect it to be a list of possible models to link the test to: pickings, production_orders, etc.

aviejo:
I see the technical problem and understand it, but not functionally, I don't know what it's for, what philosophy and if it's relevant.


It is possibly related to this field of inspections.  and can be a filter for when an inspection is for a reference.

Could that be its function? filter the bring test window?

![image](https://github.com/user-attachments/assets/c4db2f57-c8ca-4f44-b733-b23fbeca7e33)

miquelrosell99:
I fixed it in my code by changing the original method to:
```
    @api.model
    def object_selection_values(self):
        return []
```

This allows for adding any model via inheritance like:
```
    def object_selection_values(self):
        # Call super to get existing selection values
        res = set(super(QcTest, self).object_selection_values())
        # Add your custom model(s) to the set
        res.add(('product.template', 'Producto'))
        return list(res)
```

set() and then list() is used to avoid duplicates.

I am no dev and new to github, so I don't really know how to submit this change. If anyone can do it, I appreciate it,

PR #1568, opened 2025-07-21
[17.0][IMP] quality_control_oca: remove unused object_id on qt_test

remi-filament:
As reported in issue #1378 qc_test type field allows for display of object_id for which is selection is set(), creating invalid display.
Since this field does not seem to be used anywhere (none of the depending modules are using it neither), I propose with this PR to remove both object_id and type from qc_test.

Note : a field with same name object_id is used on qc_inspection model and make a lot of sense there !

### DIFF
--- a/quality_control_oca/demo/quality_control_demo.xml
+++ b/quality_control_oca/demo/quality_control_demo.xml
@@ -2,7 +2,6 @@
 <odoo noupdate="1">
     <record model="qc.test" id="qc_test_1">
         <field name="name">Generic Test (demo)</field>
-        <field name="type">generic</field>
         <field name="active" eval="True" />
         <field name="category" ref="qc_test_template_category_generic" />
     </record>
--- a/quality_control_oca/models/qc_test.py
+++ b/quality_control_oca/models/qc_test.py
@@ -17,14 +17,6 @@ class QcTest(models.Model):
     _description = "Quality control test"
     _inherit = "mail.thread"
 
-    def object_selection_values(self):
-        return set()
-
-    @api.onchange("type")
-    def onchange_type(self):
-        if self.type == "generic":
-            self.object_id = False
-
     active = fields.Boolean(default=True)
     name = fields.Char(required=True, translate=True)
     test_lines = fields.One2many(
@@ -33,16 +25,7 @@ class QcTest(models.Model):
         string="Questions",
         copy=True,
     )
-    object_id = fields.Reference(
-        string="Reference object",
-        selection="object_selection_values",
-    )
     fill_correct_values = fields.Boolean(string="Pre-fill with correct values")
-    type = fields.Selection(
-        [("generic", "Generic"), ("related", "Related")],
-        required=True,
-        default="generic",
-    )
     category = fields.Many2one(comodel_name="qc.test.category")
     company_id = fields.Many2one(
         comodel_name="res.company",
--- a/quality_control_oca/views/qc_test_view.xml
+++ b/quality_control_oca/views/qc_test_view.xml
@@ -25,10 +25,6 @@
                     </h1>
                     <group>
                         <group>
-                            <field name="type" />
-                            <field name="object_id" invisible="type == 'generic'" />
-                        </group>
-                        <group>
                             <field name="category" />
                             <field name="fill_correct_values" />
                             <field
@@ -66,8 +62,6 @@
             <tree>
                 <field name="name" />
                 <field name="category" />
-                <field name="type" />
-                <field name="object_id" />
                 <field name="company_id" groups="base.group_multi_company" />
             </tree>
         </field>


remi-filament:
@OCA/manufacturing-maintainers would you mind having a look at this PR and let me know if you think these fields are useful somewhere ?

remi-filament:
Hi @LoisRForgeFlow would you mind having a look at this PR ?
Thanks !

Saran440 (approved):
LGTM 👍 

Remove field is not used.

dreispt (approved):

pedrobaeza (changes requested):
Please put the proper label, as `[REM]` is for removing an addon, which is not the case. If you think it, in any PR you may remove some lines of code as part of your patch. You can use `[IMP]`, as it's not either a bugfix.

remi-filament:
Thanks for your reviews, I rebased and renamed both PR and commit to use [IMP] iso [REM]

pedrobaeza (approved):
/ocabot merge minor

miquelrosell99:
`object_id` removal breaks product-specific test definitions

I understand the design goal: making `qc.test` a pure template and delegating object-linking to `qc.inspection.object_id` and `qc.trigger.product_line`. However, this assumes tests are generic and reusable across products.

In manufacturing reality, **each product carries its own unique measurement specifications**. Example:

| Product | Feature | Dimension | Tolerance |
|---------|---------|-----------|-----------|
| Product A | Shaft diameter | 10 mm | ±0.1 mm |
| Product B | Shaft diameter | 15 mm | ±0.2 mm |
| Product A | Thread pitch | 1.5 mm | ±0.05 mm |
| Product B | *doesn't have this feature* | — | — |

With `object_id` gone, there's no place for product-specific `qc.test.line` records (the actual questions, dimensions, and tolerances). The trigger pattern only links a *pre-existing generic test* to a product — it doesn't let Product A have 6 test lines and Product B have 3 different ones.

**The gap:** `qc.test` is treated as a template, but the data that needs templating (`test_lines`) is inherently product-specific. To use the new architecture, we'd need a separate `qc.test` for every single product anyway, which defeats the purpose of removing `object_id`.

To work around this, I created `quality_control_oca_product` which restores product linking by adding `product_tmpl_id` and `product_id` fields to `qc.test`, with a hybrid template/variant model so base tests can be shared while variant-specific tests can be added on top. It solves our immediate need, but I'd much prefer an upstream solution.

Would the maintainers consider restoring `object_id`, or is there an intended replacement for product-specific test line definitions that I'm missing?

remi-filament:
Hi @miquelrosell99 thank you for reporting on the way you use the test from quality control.
From my point of view, the object_id as it was defined was not usable without overriding object_selection_values() which I could not find in any submodule in this repo.

I suppose that for your need you somehow override this function in your specific code ?

Your use case is very relevant, so I do not mind bringing this field back, though we should make it work when using only this quality_control_oca module !

remi-filament:
As I understand your need, you want to get values from product for filling expected values on qc.test.line, making the expected value and tolerance of these test lines dynamic depending on which product is inspected. Is that right ?

miquelrosell99:
Hi @pedrobaeza, I have submitted a new PR that addresses this issue with a different approach:

**→ #1797**

## Why a new PR instead of continuing this one

After our discussion, I realized the core use case here is **drawing inspection plans**, not dynamic expected values per variant. Each product has its own drawing with specific features (e.g. "Feature 1.3 — Hole Ø 10±0.1"), and the test lines (`qc.test.question`) ARE the drawing features. Product 01 might have Test 01 "Dimensional, post machining" and Test 02 "Roughness value test", each with unique lines that are static drawing data.

## What the new PR does differently

1. **Restores `object_id` on `qc.test`** with a dynamic, extendable selection mechanism via `_get_object_id_models()` — defaults to `product.product` and `product.template`.
2. **`object_id` empty = generic test** (can be used anywhere). **`object_id` set = product-specific drawing test**.
3. **Trigger line domain filtering**: `qc.trigger.product_line` and `qc.trigger.product_template_line` now filter the test dropdown to show only generic tests OR tests linked to the same product/template.
4. **No `type` field**: Removed the `generic`/`related` selection field based on your feedback. An empty `object_id` naturally means generic; a filled one means related.

This keeps the base module simple while allowing submodules (like `quality_control_oca_product`) to extend the model list and provide the richer UX (direct `product_tmpl_id`/`product_id` fields, One2manies, smart buttons) on top.

Would you mind reviewing #1797 when you have a moment? Thanks!

miquelrosell99:
Following the maintainer's feedback on #1797, I am closing that PR and taking a different approach.

**Decision**: Instead of restoring a generic `object_id` on `qc.test`, the product-specific test linking will be implemented directly in `quality_control_oca_product` using direct Many2one fields (`product_tmpl_id`, `product_id`).

**Rationale**: The generic `object_id` approach was removed in 17.0 because it was broken and unused. The maintainer confirmed that direct m2o links are simpler, clearer, and scoped to the module's actual purpose — avoiding future surprises from a generic mechanism that has proven to not be used by the community.

I will now implement `quality_control_oca_product` with:
- `product_tmpl_id` / `product_id` direct fields on `qc.test`
- Proper trigger line domain filtering
- Menu items and views scoped to product/template

Thanks @pedrobaeza for the guidance!

PR #1555, opened 2025-06-27
[17.0][MIG] mrp_product_generator_by_date_interval: Migration to 17.0

CarlosRoca13:
cc @Tecnativa TT52416

ping @pedrobaeza @pilarvargas-tecnativa

PR #1572, opened 2025-07-27
[17.0][MIG] mrp_production_generator_by_date_interval: Migration to version 17.0

pedrobaeza:
- Standard procedure.
- Replace `date_planned_start` by `date_start`.
- Remove onchange calls, as now everything is computed writable.

@Tecnativa TT57377


pedrobaeza:
Duplicated with #1555

PR #1576, opened 2025-07-29
[18.0][MIG] mrp_production_putaway_strategy: Migration to 18.0

ArnauCForgeFlow:
Standard migration to version 18.0


mrp_production_putaway_strategy/models/mrp_production.py
```diff
@@ -7,22 +7,27 @@
...
+
+        mos = super().create(vals_list)
+
+        for mo in mos:
             message = (
                 _(
-                    "Applied Putaway strategy to finished products.\n Finished "
-                    "Products Location: %s. "
+                    "Applied Putaway strategy to finished products.\n"
+                    "Finished Products Location: %s."
```
LoisRForgeFlow: This is not correct, you cannot post the message in all MOs that are created, it should be posted when the putaway was actually applied. For example in runbot, without any putaway, I create a MO and the message is added to the chatter:

image

LoisRForgeFlow (changes requested):
It works but there is a problem with the messages posted in chatter, see below

mrp_production_putaway_strategy/models/mrp_production.py
ArnauCForgeFlow: Could you check again? @LoisRForgeFlow 
Thanks

LoisRForgeFlow (approved):
LGTM :+1:

AaronHForgeFlow (approved):
Code review LGTM!

PR #1577, opened 2025-07-30
[18.0][MIG] `mrp_stock_rule_propagate_bom_line`: Migration to 18.0

yankinmax:


mrp_stock_rule_propagate_bom_line/tests/test_stock_rule_group_kit_line.py
```diff
@@ -0,0 +1,124 @@
...
+
+class TestStockMoveBomLinePropagate(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create({"name": "customer"})
+
+        # Set WH for multi step delivery
+        cls.warehouse = cls.env.ref("stock.warehouse0")
+        cls.warehouse.delivery_steps = "pick_pack_ship"
```
grindtildeath, line 17: We actually want to multiply the test with two warehouse having different settings:
- One with standard push rules as in Odoo 18
- One with pull rules that were standard with Odoo < 18

jbaudoux (changes requested):
This module can be dropped. The feature is now standard odoo.

yankinmax:
> This module can be dropped. The feature is now standard odoo.

Closing considering this comment

PR #1578, opened 2025-08-01
[15.0][ADD] mrp_production_location_picking_type

JoanSForgeFlow:
This module adds the ability to set a custom production location on manufacturing operation types, and it will be used that production location in manufacturing orders of that operation type and related stock moves.


mrp_production_location_picking_type/tests/test_mrp_production_location_picking_type.py
```diff
@@ -0,0 +1,145 @@
...
+        mo.action_confirm()
+        finished_moves = mo.move_finished_ids
+        self.assertTrue(finished_moves, "Finished product moves should exist")
+        for move in finished_moves:
+            self.assertEqual(
+                move.location_id,
+                self.custom_production_location,
+                f"Finished product move {move.product_id.name} should have "
+                f"custom production location as source",
+            )
```
JordiMForgeFlow: why not checking all this in the same test method?

mrp_production_location_picking_type/models/mrp_production.py
```diff
@@ -0,0 +1,57 @@
...
+        super()._compute_production_location()
+        for production in self:
+            if (
+                production.picking_type_id
+                and production.picking_type_id.production_location_id
+            ):
+                production.production_location_id = (
+                    production.picking_type_id.production_location_id
+                )
+        return True
```
JordiMForgeFlow:
```suggestion
        res = super()._compute_production_location()
        for production in self:
            if (
                production.picking_type_id
                and production.picking_type_id.production_location_id
            ):
                production.production_location_id = (
                    production.picking_type_id.production_location_id
                )
        return res
```
minor suggestion, in upper versions the CI will require super() calls to keep the result

AaronHForgeFlow (approved):
Code LGTM (aside from already existing comments)

JordiMForgeFlow (approved):
code review 👍🏼

LoisRForgeFlow (approved):
Functional test, works fine from procurement and from manual creation :+1:

PR #1579, opened 2025-08-05
[17.0][FIX] mrp_bom_attribute_match

eugenios73:
If a user tries to create a BOM and is not part of the **"Manage Multiple Units of Measure"** group, the **UoM** (Unit of Measure) option does not appear next to the **"Quantity"** field on the form. When they try to save the new BOM, the validation error from the screenshot pops up. This happens because when the **"product_uom_category_id"** field is overwritten, the **"_compute_product_uom_category_id()"** function attempts to modify the UoM based on the **"Product Template"** or **"Product Variant"**. However, the user in the current session does not have the permissions to modify this field, which triggers the error. The fact that the user is not in this group doesn't prevent them from being able to create a BOM.

Captura desde 2025-08-05 15-14-35

image

image

### DIFF
--- a/mrp_bom_attribute_match/models/mrp_bom.py
+++ b/mrp_bom_attribute_match/models/mrp_bom.py
@@ -27,7 +27,24 @@ class MrpBomLine(models.Model):
         "uom.category",
         related=None,
         compute="_compute_product_uom_category_id",
+        compute_sudo=True,
     )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for values in vals_list:
+            if (
+                not values.get("product_id")
+                and "product_uom_id" not in values
+                and "component_template_id" in values
+                and values["component_template_id"]
+            ):
+                values["product_uom_id"] = (
+                    self.env["product.template"]
+                    .browse(values["component_template_id"])
+                    .uom_id.id
+                )
+        return super().create(vals_list)
 
     @api.depends("product_id", "component_template_id")
     def _compute_product_uom_category_id(self):


eugenios73:
@mymage @LoisRForgeFlow @pedrobaeza please can you review this PR?

LoisRForgeFlow (approved):
Thanks for the good explanation

Code review :+1:

mymage:
@eugenios73 I apologise if I'm doing some wrong, but I found the same error.
I used the baseonly DB, installed the module without changing any settings, set up a component with a property, set up a product with same property, create a BOM and if I add the component directly as product-template it raised the same error (lost Field: Product Unit of Measure (product_uom_id)).
If I add the component, save and then change the component with the component-template, there is non error.
But, if I go to settings and activate the option Inventory->Products->"Units of Measure" the error don't appears.
(in the other DB the options is active so if you test there it doesn't rise the error)

eugenios73:
Hi @mymage. Don't apologize! You're right, I've reviewed it and managed to reproduce the error you mentioned. I'm going to check the functionality, because it's not the same case I fixed. Thank you very much for your tests!

eugenios73:
@mymage  I've solved the problem. When the original `create` method of `mrp.bom.line` was executed, it only modified the `product_uom_id` field. It did this by relying on the value of `product_id` and searching for the `uom_id` *only* in `product.product`.

Because of this, when components of the type `product.template` were added, the `uom_id` was never retrieved from those models, which caused the error you mentioned. The new modification now also takes the value of `component_template_id` into account to search for the `uom_id` in `product.template`, thus resolving the issue.

FYI @LoisRForgeFlow @ivantodorovich

mymage (approved):
Functional OK

LoisRForgeFlow:
Awesome, let's merge!

/ocabot merge patch

LoisRForgeFlow:
@eugenios73 could you check if this patch is needed in v18?

eugenios73:
Yes @LoisRForgeFlow , no problem, I will check in the v18.0

eugenios73:
> @eugenios73 could you check if this patch is needed in v18?

@LoisRForgeFlow Yes, the same problem exists in version 18.0. Fix for v18 https://github.com/OCA/manufacture/pull/1583

PR #1583, opened 2025-08-07
[18.0][FIX] mrp_bom_attribute_match

eugenios73:
If a user tries to create a BOM and is not part of the "Manage Multiple Units of Measure" group, the UoM (Unit of Measure) option does not appear next to the "Quantity" field on the form. When they try to save the new BOM, the validation error from the screenshot pops up. This happens because when the "product_uom_category_id" field is overwritten, the "_compute_product_uom_category_id()" function attempts to modify the UoM based on the "Product Template" or "Product Variant". However, the user in the current session does not have the permissions to modify this field, which triggers the error. The fact that the user is not in this group doesn't prevent them from being able to create a BOM.

image

image

image


eugenios73:
@LoisRForgeFlow Yes, the same problem exists in version 18.0.

LoisRForgeFlow (approved):
Thanks :+1:

LoisRForgeFlow:
@mymage can you review?

mymage (approved):
Functional ok

PR #1584, opened 2025-08-08
[16.0][ADD] quality_control_stock_oca: reminder popup for quality inspections

quirino95:
Implemented a popup asking for confirmation when trying to validate a receipt without completing related Quality Inspections before.
If Quality Inspections are triggered, and are not done (success or fail), the popup appears and user can choose between:
- Complete Inspections -> user will be redirected to the inspections page;
- Skip Inspections -> Pending inspections will be ignored and receipt will be validated;

Here's attached a screen of the new popup:
immagine


quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -419,3 +419,23 @@ def test_qc_inspection_lot(self):
...
+
+    def test_button_validate(self):
+        self.inspection1.write(
+            {
+                "name": self.picking1.move_ids[:1]._name + "inspection",
+                "object_id": "%s,%d" % (self.picking1._name, self.picking1.id),
+            }
+        )
+        # pylint: disable=W0104
+        self.picking1.qc_inspections_ids
```
SirPyTech: **question**: This smells, and looks like the test would work without it too, could you please check if it is really needed?

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,27 @@ def _create_backorder(self):
...
+    def button_validate(self):
+        res = self._pre_button_validate()
+        if res is not True:
+            return res
+        return super().button_validate()
+
+    def _pre_button_validate(self):
+        if (
+            self.created_inspections != self.done_inspections
+            and not self.env.context.get("skip_inspections", False)
```
SirPyTech: **suggestion**: Check the `context` first so we can short circuit (see https://docs.python.org/3/tutorial/datastructures.html#more-on-conditions) and avoid the computation of `created_inspections` and `done_inspections` if that's not needed.

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,27 @@ def _create_backorder(self):
         # To re-allocate backorder moves to the new backorder picking
         self.sudo().qc_inspections_ids._compute_picking()
         return res
+
+    def button_validate(self):
+        res = self._pre_button_validate()
+        if res is not True:
+            return res
+        return super().button_validate()
```
SirPyTech: **suggestion**: Please add some documentation about this new feature.

SirPyTech: **issue**: The default behavior of the method is being changed, and I can see how it will break many existing workflows where validating a picking simply results in the picking being validated.

Please enable this feature with a configuration somewhere (in the test, in the product or in the picking type for instance) so that the default behavior is preserved when no configuration is done (i.e. by default).

quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -419,3 +419,23 @@ def test_qc_inspection_lot(self):
         self.inspection1.onchange_object_id()
         self.assertEqual(self.inspection1.lot_id, self.lot)
         self.assertEqual(self.inspection1.product_id, self.lot.product_id)
+
+    def test_button_validate(self):
```
SirPyTech, line 460: **suggestion**: Please describe what is being tested in the docstring.

quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -419,3 +419,23 @@ def test_qc_inspection_lot(self):
...
+        move1 = self.picking1.move_ids[0]
+        move1.quantity_done = 2
+
+        res = self.picking1.button_validate()
+        self.assertNotEqual(res, True)
+
+        self.inspection1.write({"state": "success"})
+        res = self.picking1.button_validate()
+
+        self.assertEqual(res, True)
```
SirPyTech, line 482: **suggestion**: Please also check whether the picking has been validated.

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,27 @@ def _create_backorder(self):
         # To re-allocate backorder moves to the new backorder picking
         self.sudo().qc_inspections_ids._compute_picking()
         return res
+
+    def button_validate(self):
+        res = self._pre_button_validate()
+        if res is not True:
+            return res
```
SirPyTech: **note**: This works fine, but to me it seems tricky because it only works because we are checking that result **is** `not True`.
Please consider splitting it in two different methods, each with their responsibility:
```py
if self._needs_inspections_wizard():
    return self._show_inspections_wizard()
```

SirPyTech (changes requested):
Thanks for the PR!
I have reviewed the code and tested it, works fine.

Please see the concerns about documentation and preserving the existing behavior in the below comments.

quality_control_stock_oca/tests/test_quality_control_stock.py
quirino95: Tests also work without the two rows, so i removed them.

quality_control_stock_oca/models/stock_picking.py
quirino95: I created a new field on the product (False by default): it's in the inventory tab, quality inspection section.
Once selected, if needed, popup appears (updated also documentation).

quality_control_stock_oca/tests/test_quality_control_stock.py
SirPyTech: 👍

quality_control_stock_oca/tests/test_quality_control_stock.py
SirPyTech: :+1:

quality_control_stock_oca/models/stock_picking.py
SirPyTech: :+1:

quality_control_stock_oca/models/stock_picking.py
SirPyTech: :+1:

SirPyTech: :+1:

quality_control_stock_oca/tests/test_quality_control_stock.py
SirPyTech: :+1:

quality_control_stock_oca/models/stock_picking.py
SirPyTech: :+1:

quality_control_stock_oca/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,4 @@
 This module defines triggers that creates inspections when stock moves are done.
 
 It also adds some shortcuts on picking and lots to these inspections.
+Setting a flag on products, if inspections are not done (successfully or unsuccessfully), on picking confirmation a reminder popup appears.
```
SirPyTech: **suggestion**
```suggestion
Activating the "Remind Quality Control" flag on products, if inspections are not done (successfully or unsuccessfully), on picking confirmation a reminder popup appears.
```

quality_control_stock_oca/wizard/qc_check_wizard.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="qc_check_wizard_form" model="ir.ui.view">
+        <field name="name">QC Check</field>
+        <field name="model">qc.check.wizard</field>
+        <field name="arch" type="xml">
+            <form string="QC Check">
+                <group>
+                    <p colspan="2">
+                        Some inspections are not done or not completed.
```
SirPyTech: **suggestion**: neither "done" nor "completed" are states of an inspection: https://github.com/OCA/manufacture/blob/49f701eda70c2b049884e05837a992c908f977e1/quality_control_oca/models/qc_inspection.py#L81-L90

Although we can assume that an inspection is completed/done when it is `success` or `failed`, the comparison between "done" and "completed" makes little sense because they are the same _state_.

This would be more aligned with the added phrase in the README too:
```suggestion
                        Some inspections are not done (successfully or unsuccessfully).
```

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,27 @@ def _create_backorder(self):
...
+
+    def _needs_inspections_wizard(self):
+        if (
+            any(m.product_id.remind_qc for m in self.move_line_ids)
+            and not self.env.context.get("skip_inspections", False)
+            and self.created_inspections != self.done_inspections
+        ):
+            return True
+
+        return False
```
SirPyTech, line 137: **suggestion**

```suggestion
        return not self.env.context.get("skip_inspections")
            and any(m.product_id.remind_qc for m in self.move_line_ids)
            and self.created_inspections != self.done_inspections
```
Always check the easiest/fastest conditions first to take advantage of short circuiting, from the previous comment about short circuit
> Check the context first so we can short circuit (see https://docs.python.org/3/tutorial/datastructures.html#more-on-conditions)

(from https://github.com/OCA/manufacture/pull/1584#discussion_r2266083659)

SirPyTech (approved):
Thanks for implementing the suggested changes, I have added a few suggestions but they are not blocking

anusriNPS:
Validation of picking with `Complete Inspections` and `Skip Inspections` is verified. Code and Functional Review: LGTM.

anusriNPS (approved):

SirPyTech:
@OCA/manufacturing-maintainers this has been `ready to merge` for 5 months, can someone please have a look? Thanks!

Issue #1488, opened 2025-02-27
[16.0] mrp_bom_hierarchy: Permission error when viewing BoM with read-only access


marco-lucaf:
## Module
mrp_bom_hierarchy

## Describe the bug
When trying to display a BoM in read-only mode to a user with limited permissions, odoo raises a permission error stating that the user needs write access to mrp.bom.line, even though they're only trying to view the bom.

## To Reproduce
**Affected versions**:
- Odoo 16.0 (with mrp_bom_hierarchy module installed)

1. Create a user with read-only access to BoMs (without write permissions on mrp.bom.line)
2. Try to view a BoM with this user
3. An access error is raised when the system tries to compute the "has_bom" field

**Expected behavior**
Users with read-only access should be able to view the BoM hierarchy without encountering permission errors.

**Additional context**
The issue is in the `_compute_has_bom` method. When computing whether a BoM line has a sub-BoM, the system tries to access information that requires elevated permissions.

## Proposed solution
Adding `sudo()` to the `_compute_has_bom` method solves the issue by temporarily elevating the permissions for this specific operation:

```python
@api.depends("product_id", "bom_id")     
def _compute_has_bom(self):         
    self = self.sudo()         
    res = super()._compute_child_bom_id()
        for line in self:
            line.has_bom = bool(line.child_bom_id)
        return res
```

This allows read-only users to view the BoM hierarchy while maintaining proper security restrictions elsewhere.

LoisRForgeFlow:
Solved via https://github.com/OCA/manufacture/pull/1588

PR #1588, opened 2025-08-26
[16.0][FIX] mrp_bom_hierarchy: bom lines read permission

corredato:
FIX: #1488 w/ proposed solution

This fix uses sudo() in _compute_has_bom to bypass the access rights issue.
While it solves the problem, it is not the best approach as it elevates permissions more than strictly required.

A safer long-term solution would be to restrict access to the specific fields needed or explicitly check read permissions on mrp.bom.line

### DIFF
--- a/mrp_bom_hierarchy/model/mrp_bom_line.py
+++ b/mrp_bom_hierarchy/model/mrp_bom_line.py
@@ -12,6 +12,7 @@ class MrpBomLine(models.Model):
     has_bom = fields.Boolean(
         string="Has sub BoM",
         compute="_compute_has_bom",
+        compute_sudo=True,
     )
 
     @api.depends("product_id", "bom_id")


corredato:
@LoisRForgeFlow @pedrobaeza  review it please

pedrobaeza:
Put `compute_sudo=True` instead, but your issue report sounds weird. Anyway, I'm not using this module.

corredato:
> Put `compute_sudo=True` instead, but your issue report sounds weird. Anyway, I'm not using this module.

Was strange until I tested it. I created a user with adm permissions in the bom and only read permission in the bom line, and in fact the user could not access any bom to see the components.

I am going to test compute_sudo and make an amend

corredato:
@pedrobaeza merging this we can close #1488

LoisRForgeFlow (approved):
I can confirm the bug, and that the patch works :+1:

LoisRForgeFlow:
Thanks for fixing!

/ocabot merge patch

LoisRForgeFlow:
@corredato Could you forward port this patch to next versions if needed?

corredato:
> @corredato Could you forward port this patch to next versions if needed?

Sure

LoisRForgeFlow:
> > @corredato Could you forward port this patch to next versions if needed?
> 
> Sure

@corredato a reminder here

corredato:
> > > @corredato Could you forward port this patch to next versions if needed?
> > 
> > Sure
> 
> @corredato a reminder here

Hi @LoisRForgeFlow 

Lots of work to do currently, i'll probably have time over the weekend and create the PRs

LoisRForgeFlow:
@corredato I have done the ports myself this time, because it means less time from my side to do it than to check if you have done it or not. See https://github.com/OCA/manufacture/pull/1598 and https://github.com/OCA/manufacture/pull/1599. 

Helping with maintenance is an important part of collaborating with OCA and also helps your future self when you migrate the projects you did this for right now (nobody wants to have regressions when migrating to a new odoo version).

corredato:
> have

You're right, thanks for taking the initiative, I really am a bit short on time

PR #1592, opened 2025-08-27
[18.0][IMP] mrp_lot_number_propagation: Support mass production

grindtildeath:


mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -92,9 +91,112 @@ def _set_lot_number_propagation_data_from_bom(self):
...
+                else:
+                    productions_batch_ids.add(order.id)
+
+            batch_productions = self.browse(productions_batch_ids)
+            for order in batch_productions:
+                # If any production is not set to propagate, it will require user input
+                #  so return the wizard for that order instead of one with lot number
+                #  propagation
+                if not order.is_lot_number_propagated:
+                    res["context"] = {"default_production_id": order.id}
```
simahawk: I would not override the whole ctx unless wanted
```suggestion
                    res["context"].update({"default_production_id": order.id})
```

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -92,9 +91,112 @@ def _set_lot_number_propagation_data_from_bom(self):
...
+            return True
+        return super()._auto_production_checks()
+
     def pre_button_mark_done(self):
-        self._create_and_assign_propagated_lot_number()
-        return super().pre_button_mark_done()
+        if all(prod._auto_production_checks() for prod in self):
+            self._create_and_assign_propagated_lot_number()
+        res = super().pre_button_mark_done()
+        if isinstance(res, dict) and res["res_model"] == "mrp.batch.produce":
```
simahawk, line 119: this part can be maybe split in smaller methods

mrp_lot_number_propagation/tests/test_mrp_production.py
```diff
@@ -143,3 +175,87 @@ def test_confirm_with_variant_no(self):
...
+        )
+        wizard = wizard_form.save()
+        with self.assertRaisesRegex(
+            UserError, "set to propagate lot number from component"
+        ):
+            wizard.action_done()
+        wizard_form = Form(wizard)
+        wizard_form.production_text = "\n".join(
+            [
+                f"{comp_serial_name},{comp_serial_name},{component_lot_name}"
```
simahawk, line 216: why the serial should be duplicated?

mrp_lot_number_propagation/wizard/mrp_batch_produce.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _production_text_to_object(self, mark_done=False):
+        self._check_propagated_lot_number()
+        self = self.with_context(lot_propagation=True)
+        return super()._production_text_to_object(mark_done)
+
+    def _check_propagated_lot_number(self):
+        if not self.is_lot_number_propagated:
+            return
+        for production_line in deque(self.production_text.split("\n")):
```
simahawk, line 43: I'm not sure you gain that much using `deque` here... You are not consuming the whole list as a queue.
Here a `splitlines` should be enough since you iterate on every element and then `split(sep, 1)` will return only the 1st element. Just a question tho.

simahawk (commented):
minor remarks

mrp_lot_number_propagation/wizard/mrp_batch_produce.py
```diff
@@ -0,0 +1,58 @@
...
+                    f"Lot number must be the same between finished product "
+                    f"{wiz.production_id.product_id.display_name} and component "
+                    f"{propagated_component.display_name}"
+                )
+            wiz.lot_number_propagation_alert_msg = msg
+
+    def _production_text_to_object(self, mark_done=False):
+        self._check_propagated_lot_number()
+        self = self.with_context(lot_propagation=True)
+        return super()._production_text_to_object(mark_done)
```
sebalix:
```suggestion
        return super()._production_text_to_object(mark_done=mark_done)
```

mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -92,9 +91,112 @@ def _set_lot_number_propagation_data_from_bom(self):
...
+            for order in batch_productions:
+                batch_produce_wizard = (
+                    self.env["mrp.batch.produce"]
+                    .with_context(default_production_id=order.id)
+                    .create({})
+                )
+                propagate_move = order._get_propagating_component_move()
+                batch_produce_wizard.production_text = "\n".join(
+                    [
+                        f"{ml.lot_id.name},{ml.lot_id.name}"
```
sebalix: I don't know this `production_text` field, but here you should not use `self.component_separator` instead of comma, as you are splitting on this in the `mrp.batch.produce` wizard?

mrp_lot_number_propagation/models/mrp_production.py
grindtildeath: Actually the context return is exactly this, but I agree we can update to be nice with any other module that would override this

mrp_lot_number_propagation/tests/test_mrp_production.py
grindtildeath: The module is meant to propagate lot number from a component to the finished product. In that case we need to define it twice, first for the finished product, then for the component that is propagating.

mrp_lot_number_propagation/models/mrp_production.py
grindtildeath: I usually think it's good to split in smaller function, but there it's a single process of multiple small steps/checks, hence why I added lots of comment to explain what each is doing.

IMO splitting this doesn't improve the readability of the code

mrp_lot_number_propagation/wizard/mrp_batch_produce.py
grindtildeath: Actually I used `deque` because that is what is used in the super function. I guess I can add a comment about that and keep in line with what's done in odoo

mrp_lot_number_propagation/models/mrp_production.py
grindtildeath: Actually the comma is the default value for `component_separator`, but yes, let's be on the safe side in case it's changed :+1:

mrp_lot_number_propagation/wizard/mrp_batch_produce.py
```diff
@@ -0,0 +1,59 @@
...
+                if move_raw.product_id.tracking == "none":
+                    continue
+                if (
+                    move_raw.propagate_lot_number
+                    and components_line[i] != finished_line
+                ):
+                    raise exceptions.UserError(
+                        self.env._(
+                            f"As the product being produced is set to propagate lot number from component "  # noqa: E501
+                            f"{self.production_id._get_propagating_component_move().product_id.display_name},"  # noqa: E501
```
ivantodorovich: This translation string will not work.
Let gettext do the string interpolation

mrp_lot_number_propagation/tests/test_mrp_production.py
ivantodorovich: Maybe add a comment with this explanation

ivantodorovich (approved):

vvrossem (approved):
pre-approving: pre-commit needs to be executed

PR #1601, opened 2025-09-09
[16.0][ADD] mrp_production_unique_lot

MarinaAForgeFlow:
This module enforces the uniqueness of lot numbers for finished products in production orders. It prevents the same lot number from being used in multiple productions.

This can useful for businesses where each lot is strictly associated with the specific manufacturing order in which the product was produced (each production produces a different lot). By using this module, the user can avoid problems and errors in relation to an already existing lot being used twice.

The module follows the same logic as the serial number uniqueness feature and the code is based on the odoo code.


mrp_production_unique_lot/models/mrp_production.py
```diff
@@ -0,0 +1,140 @@
...
+            order._check_lot_uniqueness()
+        return res
+
+    def _check_lot_uniqueness(self):
+        """Alert the user if the lot number has already been produced"""
+        if not self.force_lot_uniqueness or not self.lot_producing_id:
+            return
+
+        if self.product_tracking == "lot" and self.lot_producing_id:
+            if self._is_finished_lot_already_produced(self.lot_producing_id):
```
JordiMForgeFlow: this could be joined into only one if statement

AaronHForgeFlow (approved):
Functional ok!

kaynnan (approved):

PR #1559, opened 2025-06-30
[16.0][IMP] quality_control_stock_oca: inspection mandatory on picking

Kev-Roche:
**Purpose**
This PR introduces a mandatory inspection step before confirming a picking.
Depending on the picking type, the inspection must be confirmed prior to confirming the picking itself.

**Additional Changes**
This PR also includes three minor improvements:
- The smart buttons for Created / Done / Passed / Failed inspections are now only visible when their respective counts are not null. Also,  Improved display of inspection names in the smart button.

- lot_id determination: When confirming a picking, the lot_id on stock moves now takes into account restricted lots from the stock_restrict_lot module in order to determine the correct lot if installed.

- add timing field in qc trigger line views.

PR #1606, opened 2025-09-12
[18]quality_control_stock_oca: inspection mandatory on picking

florian-dacosta:
Port of https://github.com/OCA/manufacture/pull/1559

Purpose
This PR introduces a mandatory inspection step before confirming a picking.
Depending on the picking type, the inspection must be confirmed prior to confirming the picking itself.

Additional Changes
This PR also includes three minor improvements:

    The smart buttons for Created / Done / Passed / Failed inspections are now only visible when their respective counts are not null. Also, Improved display of inspection names in the smart button.

    lot_id determination: When confirming a picking, the lot_id on stock moves now takes into account restricted lots from the stock_restrict_lot module in order to determine the correct lot if installed.

    add timing field in qc trigger line views.

### DIFF
--- a/quality_control_stock_oca/models/stock_picking.py
+++ b/quality_control_stock_oca/models/stock_picking.py
@@ -4,7 +4,7 @@
 # Copyright 2024 Quartile
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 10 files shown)


florian-dacosta:
@Kev-Roche

Kev-Roche (approved):

florian-dacosta:
Hello @AaronHForgeFlow @LauraCForgeFlow 
Would you mind to review this one (merged in v16 already)

florian-dacosta:
@bguillot Can you reopen it ?
Maybe review it also if you can

bguillot (approved):

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -30,6 +30,22 @@ class StockPicking(models.Model):
         compute="_compute_count_inspections", string="Inspections failed"
     )
 
+    inspection_required_message = fields.Html(
+        readonly=True, compute="_compute_inspection_required_message"
+    )
+
+    @api.depends("qc_inspections_ids")
+    def _compute_inspection_required_message(self):
+        message = _("Control quality is required to validate this " "Transfert.")
```
diggy128, line 39: Maybe better use self.env._() instead of importing and using _()

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -79,6 +95,17 @@ def action_cancel(self):
         return res
 
     def _action_done(self):
+        for picking in self:
+            picking_names = ""
+            if picking.inspection_required_message:
```
diggy128, line 100: Since inspection_required_message is an html field, you should use is_html_empty().
It may now as is but at some point it may lead to always evaluating True (e.g. odoo writing a default value on that field).
Furthermore, does it really need to be an html field?
Maybe consider Text to keep newline and have formatting done in the view.

diggy128 (changes requested):

PR #1610, opened 2025-09-20
[18.0][IMP] mrp_attachment_mgmt: use product.document search view for preview

dhx:
For product.template and product.product using the product_document_search view allows us to show attachment previews


dhx:
Hi @MarinaAForgeFlow would this PR also make sense to you?

MarinaAForgeFlow:
> Hi @MarinaAForgeFlow would this PR also make sense to you?

It looks correct to me at first glance. Could you rebase it so the runboat build runs and I can test it on runboat?

dhx:
@MarinaAForgeFlow thanks for looking into this and sorry for the delayed rebase - runboat build should be good now, please note, this only changes the view for product attachments as for bom attachments I'm not aware of a model specific action in odoo core.

PR #1611, opened 2025-09-22
[16.0][ADD] quality_control_stock_oca: automatic scrap creation on Inspection failure

quirino95:
Depends on https://github.com/OCA/manufacture/pull/1584

New flag "Scrap Automatically" on products.
When an inspection related to a picking fails, on picking validation, scraps are automatically created for all products whose inspection has failed and have the flag active.
If all products in a picking have the flag active, and all inspections are completed (both success or fail), validation is completed automatically (and of course needed scraps are created).


quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,54 @@ def _create_backorder(self):
...
+    def button_validate(self):
+        if self._needs_inspections_wizard():
+            return self._show_inspections_wizard()
+
+        super().button_validate()
+
+        if all(qc.product_id.auto_scrap for qc in self.qc_inspections_ids):
+            self.create_scraps()
+
+        return True
```
HekkiMelody: **issue**: tests of another module seem to be broken because of this. Can you test if this change fixes it?


```suggestion
        res = super().button_validate()

        if all(qc.product_id.auto_scrap for qc in self.qc_inspections_ids):
            self.create_scraps()

        return res
```

HekkiMelody (changes requested):

quality_control_stock_oca/models/qc_inspection.py
```diff
@@ -79,6 +79,17 @@ def _prepare_inspection_header(self, object_ref, trigger_line):
...
+    def action_approve(self):
+        super().action_approve()
+        if (
+            self.picking_id.created_inspections == self.picking_id.done_inspections
+            and all(m.product_id.auto_scrap for m in self.picking_id.move_ids)
+        ):
+
+            self.picking_id.button_validate()
+
+        return True
```
HekkiMelody: **suggestion**:


```suggestion
        res = super().action_approve()
        if (
            self.picking_id.created_inspections == self.picking_id.done_inspections
            and all(m.product_id.auto_scrap for m in self.picking_id.move_ids)
        ):

            self.picking_id.button_validate()

        return res
```

quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -419,3 +419,91 @@ def test_qc_inspection_lot(self):
...
+            {
+                "name": self.picking1.move_ids[:1]._name + "inspection",
+                "object_id": "%s,%d" % (self.picking1._name, self.picking1.id),
+                "state": "failed",
+                "qty": 2,
+            }
+        )
+        self.inspection1.action_approve()
+
+        self.assertEqual(self.picking1.state, "done")
```
anusriNPS, line 501: **Observation:** If there are two products in a receipt, one with scrap `True` and other with scrap `False`, picking is not in `done` and `scrap` not created for the product whose inspection `Failed`.

quality_control_stock_oca/models/stock_picking.py
```diff
@@ -93,3 +93,54 @@ def _create_backorder(self):
...
         self.sudo().qc_inspections_ids._compute_picking()
         return res
+
+    def button_validate(self):
+        if self._needs_inspections_wizard():
+            return self._show_inspections_wizard()
+
+        res = super().button_validate()
+
+        if all(qc.product_id.auto_scrap for qc in self.qc_inspections_ids):
```
anusriNPS: **Question:** Why do we create scraps, if `only all products` in a receipt has 'scrap automatically' as `True`? shouldn't be `any` one of the products with scrap?

quality_control_stock_oca/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,7 @@
 This module defines triggers that creates inspections when stock moves are done.
 
 It also adds some shortcuts on picking and lots to these inspections.
+Activating the "Remind Quality Control" flag on products, if inspections are not done (successfully or unsuccessfully), on picking confirmation a reminder popup appears.
+Activating "Scrap Automatically" flag on products, if inspections fail, picking:
+- Will be automatic validated if every product in picking has the flag checked;
```
anusriNPS, line 6: **Question:** I see in description it says: "Moves should be done once set inspection as "failed" (so the "validation" should be automatic).", So I think whenever inspections are done and failed, Validation has to be completed (i.e. Pickings are in `Done` state) and also create scrap if any one of the products has 'scrap automatically' as `True` in a receipt- Is my understanding right?

If so, here mentioning validation is automatic only if `every product` has checked with `scrap automatically` is valid?

quality_control_stock_oca/tests/test_quality_control_stock.py
```diff
@@ -419,3 +419,91 @@ def test_qc_inspection_lot(self):
...
+
+        self.assertFalse(self.picking1.has_scrap_move)
+
+    def test_auto_create_scraps(self):
+        """
+        Test scraps creation when inspection fails
+        if relative flag is on
+        """
+        self.product.auto_scrap = True
+        self.picking1.move_ids[:1].quantity_done = 2
```
anusriNPS, line 531: **Observation:** Same observation with two products(one with scrap and other without scrap) under a receipt, validation is not done and scrap not created

image

anusriNPS (commented):
Code and Functional review: Works fine. I would like to clarify some queries related to using 'scrap automatically' checkbox functionality.

quality_control_stock_oca/readme/DESCRIPTION.rst
quirino95: Yes, is valid. This is a way to keep the default behavior, but also a chance to have a different behavior (checking flag `auto_scrap`) to speed up picking validation.

quality_control_stock_oca/models/stock_picking.py
quirino95: This is a good suggestion: just fixed in a bit different way. Now every product is scrapped when needed, independently from the others

quality_control_stock_oca/tests/test_quality_control_stock.py
quirino95: Based on the comments above: both products need to have `auto_scrap` flag checked in order to automatic validate picking. With last fix (https://github.com/OCA/manufacture/pull/1611#discussion_r2419180025), now needed scraps are created on validation.

quality_control_stock_oca/tests/test_quality_control_stock.py
quirino95: Just fixed (see comment below)

anusriNPS: **Suggestion:** Can we add scrap creation validation as well for this testcase?

quality_control_stock_oca/tests/test_quality_control_stock.py
anusriNPS: Agree validation based on `auto_scrap` flag and scrap creation on failed inspections once after validation handled manually. However, I see an observation as below where after validation completion, for failed inspection a new inspection is triggered with same use case above which was not seen before this latest fix. Both trigger with timing as `before`
Inspection view:
Screenshot from 2025-10-14 13-11-55

Transfer view:
image

quality_control_stock_oca/readme/DESCRIPTION.rst
anusriNPS: Thank you for clarification. Implementation works intact for success scenarios but with respect to failed inspections after latest fix observed some improper behaviour as commented below. Could not figure out yet what makes this trigger of inspection when `auto_scrap` is checked and inspection is failed.

quality_control_stock_oca/models/stock_picking.py
anusriNPS: It looks good when handled in create_scraps() for inspection failure with auto_scrap is checked.

quality_control_stock_oca/tests/test_quality_control_stock.py
quirino95: Just added scrap creation validation, but not in that method since it tests inspection approval.
Later in the file there is `test_auto_create_scraps`, where i tested scrap creation, and now I added a check on scraps state.

quality_control_stock_oca/tests/test_quality_control_stock.py
quirino95: I think that behaviour is not related to this feature, since I was able to reproduce same issue with `auto_scrap` off on both products.
I've tested also with different products:
immagine

immagine

With `auto_scrap` set on both products, and no new Inspections were created. Maybe other products have different inventory rules, or Inspection settings.

quality_control_stock_oca/readme/DESCRIPTION.rst
quirino95: Please see comment below

quality_control_stock_oca/tests/test_quality_control_stock.py
anusriNPS: I tried to use `runboat` build instance on this PR to check issue reproduction. Agree, as you have mentioned `auto_scrap` as `off` on both products, `auto_scrap` as `on` on both products and `inspection` failed on any one of the products, it works as expected. No new inspection created.

However, when `auto_scrap` is set for one of the products and `inspection fails` for product whose `auto_scrap=true`, new inspection is created after validation.
RunBoat Instance view:
http://oca-manufacture-16-0-pr1611-c823028a1f8e.runboat.odoo-community.org/web#id=27&cids=1&menu_id=120&action=238&model=stock.picking&view_type=form

Also, when `auto_scrap=false` on products in the Pickings, i don't observe this new inspection creation, which is also verified in runboat instance. If its happening on runboat instance, it should not be settings related issue. What do you think?

anusriNPS: Based on discussion, it was identified that when a `scrap is created` for a product whose `inspection is failed`, there is a new inspection created after validation. This is an existing behaviour and no issues with any feature introduced with this PR.

As per first patchset, when one of the products with `auto_scrap=true` , scrap was not created, hence new inspection was not seen. This is the reason for difference in behaviour observed between patchsets.

anusriNPS:
Code and Functional Review: LGTM

anusriNPS (approved):

quality_control_stock_oca/models/qc_inspection.py
```diff
@@ -79,6 +79,17 @@ def _prepare_inspection_header(self, object_ref, trigger_line):
                 res["qty"] = object_ref.quantity_done
         return res
 
+    def action_approve(self):
+        res = super().action_approve()
+        if (
+            self.picking_id.created_inspections == self.picking_id.done_inspections
+            and all(m.product_id.auto_scrap for m in self.picking_id.move_ids)
+        ):
```
HekkiMelody: **question**: should we add an `if self.picking_id` here? We could approve an inspection with no related pickings, and in that case we don't need to do any of this.

**suggestion**(non-blocking): move this check

```py
self.picking_id.created_inspections == self.picking_id.done_inspections
            and all(m.product_id.auto_scrap for m in self.picking_id.move_ids)
```
to a separate method (maybe something like `_check_validate_picking`?) to make it easier to customize in future/private modules.

quirino95: Good idea, updated `if` statement and moved in a separate method.

HekkiMelody (approved):
Code review only, LGTM

SirPyTech:
@OCA/manufacturing-maintainers this has been `ready to merge` for 4 months, can someone please have a look? Thanks!

PR #1616, opened 2025-09-26
18.0 FIX mrp_bom_section_and_note: consu instead of product in ctx

bealdav:
"product" no more exist as product type entry

default_detailed_type field no more exist in 18


quentinDupont (commented):
What was the problem ? :)

quentinDupont (approved):
code review and tested on runboat with success :heavy_check_mark:

hparfr (approved):

clementmbr (approved):

PR #1618, opened 2025-09-29
[18.0][FIX] mrp_lot_number_propagation: Raise error if no component reserved fo…

grindtildeath:
…r mass produce


pedrobaeza:
Please don't abbreviate the module name, as doing searches over commit names won't match with your abbreviations. No matter if the message is a bit long. It's a trade-off that we have, but it's better than non being able to locate things.

grindtildeath:
Well it's been happening on other repos but whatever :shrug:

henrybackman (approved):

sebalix (approved):

PR #1626, opened 2025-10-10
[16.0][FIX] mrp_multi_level: MRP areas in stock location groups

LauraCForgeFlow:


LoisRForgeFlow (approved):
To clarify: without this, a user without access to manufacturing cannot access stock.location form

PR #1631, opened 2025-10-16
[IMP] mrp_production_back_to_draft - Translation Portuguese Brazil.

rodmad85:


quentinDupont:
I'm not sure but i think we have to go to weblate now to handle translation. Maybe not when it's a new language, i'm note sure

dreispt (approved):

dreispt:
Translations are better handled here: https://translation.odoo-community.org/

PR #1632, opened 2025-10-17
[19.0][MIG] mrp_bom_attribute_match: Migration to 19.0.

hieulucky111:
[MIG] mrp_bom_attribute_match 19.0
[FIX] Error when clicking "Overview" on a Manufacturing Order if the BOM has Attribute Match
[FIX] fail pipeline dueto tests in mrp_multi_level setting the wrong value for the field.


ferran-S73 (approved):
code review

Vang-NguyenPhu (approved):
LGTM

huan-tq (approved):
LGTM

DucTruongKomit (approved):
Tested and review code.
It seems good to me.

Quan-nhm (approved):
code review LGTM

WesleyOliveira98 (approved):

BhaveshHeliconia (approved):
LGTM!

mrp_bom_attribute_match/models/mrp_bom.py
```diff
@@ -0,0 +1,334 @@
...
+                    [
+                        ("product_tmpl_id", "=", comp.id),
+                        ("attribute_id", "=", ptav.attribute_id.id),
+                        (
+                            "product_attribute_value_id",
+                            "=",
+                            ptav.product_attribute_value_id.id,
+                        ),
+                    ]
+                )
```
ferran-S73, line 326: We should either check that the ptav is active in the component template or only check the attributes that are configured in the bom line.

I'll expose my use case:

1. Create a product with attributes A, B and C.
2. Create a component with attributes A, B and C.
3. Remove C from the component attributes. This sets the field `ptav_active=False` instead of deleting the ptav.
4. Set up a bom to produce our product with a bom line for the component template matching on attributes A and B.
5. When this iteration is performed it will search for a ptav for attribute C and will find one that isn't active if it had the same value as the the product. Thus resulting in an impossible combination by the current component config and eventually returning False.

vangnp3702: I have fixed the issue. Could you please help to test again? Thank you

PR #1633, opened 2025-10-19
[17.0][MIG] mrp_bom_widget_section_and_note_one2many

ValentinVinagre:
standard migration
T-9042


ValentinVinagre:
@luis-ron could u review? ty

Jaimermaccione (changes requested):
@ValentinVinagre **Functional review:** The module works perfectly and the feature for adding **notes and sections** to Bills of Materials behaves as expected. Tested this in local environment.

---

#### Observation

The only point to note is related to the README statement:

> “It adds a page in the BoM form view and hides the original page (you can see it as admin in debug mode).”

From what I understand, this means the original **“Components”** tab should be hidden and replaced by the new **“Components and Sections”** tab, visible only for administrators in debug mode.

However, both tabs (**“Components”** and **“Components and Sections”**) remain visible, regardless of whether the user is in debug mode or has admin rights.

BoM Tabs Screenshot

---

#### Suggestion

The term **“admin”** in the README is somewhat ambiguous — it would be clearer to specify which exact access group it refers to (e.g. *Manufacturing Administrator*).

Perhaps I’m misinterpreting the intended behavior, but **this same situation seems to occur in other versions as well (16.0 or 18.0 tested on runboat)**.

If this behavior is correct — either because it’s intended or because I may have misunderstood the README statement — please let me know so I can mark it as functionally approved. Thanks!

ValentinVinagre:
@Jaimermaccione I don't see anything in the code under the tab that says it's hidden. Are you sure that's in the migrated version and not a later one? If it needs to be done, it's not relevant to this migration; it would have to be in a later IMP.
Regarding the Readme file, I can change it without any problem. Could you tell me how you want to specify it so we don't have any review issues?
Thanks.

Jaimermaccione:
@ValentinVinagre  Thanks for the clarification!

It’s possible that this behavior (hiding the original tab) existed in earlier versions but wasn’t carried over during one of the previous migrations.
I’ve **tested it in versions 16.0 and 18.0**, and the behavior is the same — **both tabs remain visible**.

However, **in version 12.0**, which is likely the original one, **the behavior is different**: At least, when not in debug mode, only the “Components and Sections” tab is visible, and enabling debug mode reveals both tabs (“Components” and “Components and Sections”).

So, if that functionality was intentionally removed in later versions, I’d simply update the README so it reflects the current behavior.

In that case, I’d leave it like this:

> In a similar way to invoices, this module adds the possibility to include sections and notes in Bills of Materials.
> It adds a new page in the BoM form view that includes both components and sections.
> 
> This module depends on the account module, since the widget section_and_note is defined in section_and_note_fields_backend.js within the account module.

That way the documentation stays consistent with the actual implementation and avoids confusion in future reviews.

Ty!!

Jaimermaccione:
@ValentinVinagre To wrap this up, please update the README so that the introductory paragraph is as follows:

In a similar way to invoices, this module adds the possibility to include **sections and notes** in Bills of Materials.  
It adds a new page in the BoM form view that includes both components and sections.  

This module depends on the account module, since the widget `section_and_note` is defined in `section_and_note_fields_backend.js` within the `account` module.


Once you’ve made the change, please let me know so I can review it.
Thanks!

quentinDupont:
> I’ve tested it in versions 16.0 and 18.0, and the behavior is the same — both tabs remain visible.
> 
> However, in version 12.0, which is likely the original one, the behavior is different: At least, when not in debug mode, only the “Components and Sections” tab is visible, and enabling debug mode reveals both tabs (“Components” and “Components and Sections”).
>
>  So, if that functionality was intentionally removed in later versions, I’d simply update the README so it reflects the current behavior.

Hello !
Thank you for the work !
That's a pity if we can see the two tabs if we are not admin, it's confusing and not useful imho. In any case, it was my goal to hide it in version 12 that I developed, as you noticed when you tested it.

ValentinVinagre:
> > I’ve tested it in versions 16.0 and 18.0, and the behavior is the same — both tabs remain visible.
> > However, in version 12.0, which is likely the original one, the behavior is different: At least, when not in debug mode, only the “Components and Sections” tab is visible, and enabling debug mode reveals both tabs (“Components” and “Components and Sections”).
> > So, if that functionality was intentionally removed in later versions, I’d simply update the README so it reflects the current behavior.
> 
> Hello ! Thank you for the work ! That's a pity if we can see the two tabs if we are not admin, it's confusing and not useful imho. In any case, it was my goal to hide it in version 12 that I developed, as you noticed when you tested it.

Hi, Completely understandable. What hasn't been indicated previously is that "IMP" and, in order not to stop the migration, it can be done later.

ValentinVinagre:
> @ValentinVinagre To wrap this up, please update the README so that the introductory paragraph is as follows:
> 
> In a similar way to invoices, this module adds the possibility to include **sections and notes** in Bills of Materials. It adds a new page in the BoM form view that includes both components and sections.
> 
> This module depends on the account module, since the widget `section_and_note` is defined in `section_and_note_fields_backend.js` within the `account` module.
> 
> Once you’ve made the change, please let me know so I can review it. Thanks!

it's okay?

Jaimermaccione (approved):
@ValentinVinagre LGTM! Changes reviewed and everything looks fine from my side.

ValentinVinagre:
@LoisRForgeFlow could u review? is a standard migration. Ty

HaraldPanten:
@etobella @JordiBForgeFlow  @pedrobaeza   Could you merge it, please?

THX in advance.

HaraldPanten (approved):

etobella (approved):

PR #1635, opened 2025-10-20
[18.0][IMP] mrp_multi_level: user_has_groups call is deprecated

AaronHForgeFlow:
Not causing errors in v18, but better to anticipate errors in newer versions.

Method user_has_groups() calls self.env.user.has_group


LoisRForgeFlow (approved):
Absolutely, thanks!

PR #1636, opened 2025-10-27
[16.0][FIX] mrp_unbuild_restore_origin: fix when unbuild MO with lot_producing_id

nobuQuartile:
@qrtl
QT5868

**Issue**
When executing button_unbuild, the lot assigned to the Manufacturing Order (MO) is set in the context as default_lot_id. As a result, when creating related material stock_move_line records, the Unbuild(MO)’s lot is automatically used as the default lot_id.
This causes an error:　"This lot ~~ is incompatible with this product ~~"

**Solution**
Replaced the custom _get_move_line_vals usage with the standard _prepare_move_line_vals, so that lot_id is now correctly set.
_prepare_move_line_vals: https://github.com/OCA/OCB/blob/52bec03/addons/mrp/models/mrp_unbuild.py#L146C5-L155C10

**Improvement (New Logic Added)**
Introduced a taken_quantity calculation following the logic in
`mrp_unbuild.py` L189–L207: https://github.com/OCA/OCB/blob/52bec03/addons/mrp/models/mrp_unbuild.py#L189-L207
This is newly added logic—separate from the fix above—to improve this module.

### DIFF
--- a/mrp_unbuild_restore_origin/models/mrp_unbuild.py
+++ b/mrp_unbuild_restore_origin/models/mrp_unbuild.py
@@ -29,21 +29,6 @@ class MrpUnbuild(models.Model):
             vals["location_dest_id"] = origin_move_line.location_id.id
         return vals
 
-    def _get_move_line_vals(self, move, move_line):
-        vals = {
-            "move_id": move.id,
-            "owner_id": move_line.owner_id.id,
-            "qty_done": min(move.product_uom_qty, move_line.qty_done),
-            "product_id": move.product_id.id,
-            "product_uom_id": move.product_uom.id,
-            "location_id": move.location_id.id,
-            "location_dest_id": move.location_dest_id.id,
-        }
-        if self.env.context.get("exact_location"):
-            vals["location_id"] = move_line.location_dest_id.id
-            vals["location_dest_id"] = move_line.location_id.id
-        return vals
-
     def _generate_produce_moves(self):
         """This logic may seem a bit complex, but it's necessary due to how the following
         standard code works:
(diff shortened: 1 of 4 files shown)


mrp_unbuild_restore_origin/models/mrp_unbuild.py
```diff
@@ -75,8 +61,12 @@ def _generate_produce_moves(self):
...
                     vals_list = []
                     for move_line in raw_move.move_line_ids:
-                        vals = self._get_move_line_vals(move, move_line)
-                        vals_list.append(vals)
+                        taken_quantity = min(move.product_uom_qty, move_line.qty_done)
+                        if taken_quantity:
+                            vals = self._prepare_move_line_vals(
+                                move, move_line, taken_quantity
+                            )
+                            vals_list.append(vals)
```
yostashiro: I think it should be like this to handle the case where an origin move came with multiple move lines.
```.py
                    remaining_qty = move.product_uom_qty
                    for move_line in raw_move.move_line_ids:
                        if remaining <= 0:
                            break
                        take_qty = min(remaining_qty, move_line.qty_done)
                        if not take_qty:
                            continue
                        vals = self._prepare_move_line_vals(move, move_line, take_qty)
                        vals_list.append(vals)
                        remaining_qty -= take_qty
```

mrp_unbuild_restore_origin/models/mrp_unbuild.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2025 Quartile (https://www.quartile.co)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+
```
yostashiro:
```suggestion
```

yostashiro (changes requested):
Can you please try improving the commit message, as the rationale is not obvious?

yostashiro (commented):
> preventing incorrect field assignments from the original raw moves

Sorry, this part of the commit message doesn't sound correct.

The incorrect field assignment (for `lot_id`) was coming from the `default_lot_id` context from the standard setup, and using `_prepare_move_line_vals()` would nullify this incorrect assignment as the method specifically takes `lot_id` from the origin move line.

yostashiro (approved):
Code review. LGTM.

AungKoKoLin1997 (approved):

PR #1637, opened 2025-10-28
[18.0] [ADD] sale_mrp_bom_menu

ivantodorovich:
This module adds a menu in **Sales > Products > Bills of Materials** to easily navigate to the Bills of Materials from the Sales app.


yankinmax (approved):
Thanks

sale_mrp_bom_menu/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2025 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```
vvrossem:
```suggestion
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```

or simply remove the copyright from this file

vvrossem (approved):

PR #1638, opened 2025-10-29
[18.0][ADD] mrp_component_no_consumption

MarinaAForgeFlow:
This module allows to configure a product to not be consumed in manufacturing orders. 

This could be useful, for example, when working with molds. You need to have traceability of when and where a mold was used, and you need to ensure the mold is available before using it in a production. However, the mold should not be consumed. By making it storable and checking the new field "skip_mo_consumption", you can do this.

To avoid the consumption of the component, the move has been marked as done before the _apply_inventory super call, as inside this method the done moves are skipped. However, there is a limitation with this, which is that the mold will not be taken into account for the finished product cost calculation. 

The logic should be moved somewhere else, perhaps in the _action_done of the stock.move, if we want to take into account the mold cost. Any suggestion on if this should be done, or any other alternative, is welcomed.


quentinDupont:
Open question : should it be handle with product's type ?

diggy128:
We had many similar issues.
The problem in our case was that most of these "accessories" did not make sense to have as products and additionally there are other parameters (time, machine settings, etc) so instead we created an additional model in maintenance.equipment for these setup parameters.

In your case you mention that you mark the move as done (guessing overriding all code that actually moves the product)?
This can and will inevitably have complications during migrations in the future (I've been bitten by this kind of juggling).

The best handling that comes to mind is as follows:

1. You make each mold storable. This allows to keep an inventory of qty.
2. You add the mold in the BOM consume line
3. You add the mold as a by-product in BOM.
4. When you start production, you consume the mold
5. When you finish the production, you produce the mold as a by-product

This gives you the following advantages:

1. You know which mold was used in each production
2. You reserve molds to use in production
3. You know at every time which molds are not used in production (available quantity)
4. When you complete the production (and complete the by-product move) the mold is returned back to the inventory
5. In case a mold breaks during production, you cancel the by-product move so the quantity for that mold is updated (aka not changed as the mold never returned to stock)
6. This will also allow to take into account the mold cost, but I don't know how it is handled for the by-product.

MarinaAForgeFlow:
> We had many similar issues. The problem in our case was that most of these "accessories" did not make sense to have as products and additionally there are other parameters (time, machine settings, etc) so instead we created an additional model in maintenance.equipment for these setup parameters.
> 
> In your case you mention that you mark the move as done (guessing overriding all code that actually moves the product)? This can and will inevitably have complications during migrations in the future (I've been bitten by this kind of juggling).
> 
> The best handling that comes to mind is as follows:
> 
> 1. You make each mold storable. This allows to keep an inventory of qty.
> 2. You add the mold in the BOM consume line
> 3. You add the mold as a by-product in BOM.
> 4. When you start production, you consume the mold
> 5. When you finish the production, you produce the mold as a by-product
> 
> This gives you the following advantages:
> 
> 1. You know which mold was used in each production
> 2. You reserve molds to use in production
> 3. You know at every time which molds are not used in production (available quantity)
> 4. When you complete the production (and complete the by-product move) the mold is returned back to the inventory
> 5. In case a mold breaks during production, you cancel the by-product move so the quantity for that mold is updated (aka not changed as the mold never returned to stock)
> 6. This will also allow to take into account the mold cost, but I don't know how it is handled for the by-product.

Thanks for the suggestion! It does make sense your solution and, as you say, is much more easy to maintain and migrate. Is this something that you are already using? Also, how you prevent human errors (imagine the consumed quantity does not match the by-product produced quantity)? I assume that a check or warning should be added when marking the production as done.

PR #1640, opened 2025-11-03
[19.0][MIG] mrp_multi_level

JordiMForgeFlow:
- [x] https://github.com/OCA/manufacture/pull/1641
- Includes FW of #1655 

Migration:
- Procurement group no longer exists, methods are now on stock rule model.
- Pylint error when searching without specified domain, we add comments to ignore it.
- Groupby filters in views require a domain.
- Purchase UoM field no longer exists on products. The UoM in the PO is either manually set on the PO or via supplier pricelist configuration.
- Odoo no longer runs tests with demo mode so all demo data is not available in tests. New method added to create all required data as it was in V18.
- Buy route has now higher sequence than manufacture route, test adapted accordingly.
- MRP security groups no longer have implicit Product modification rights, needs to be added explicitly in the test user setup.


MiquelRForgeFlow:
You have to add https://github.com/OCA/manufacture/pull/1655.

AaronHForgeFlow (approved):
Code review 👍

LoisRForgeFlow (approved):
Code and functional review 👍

PR #1645, opened 2025-11-05
[18.0][ADD] mrp_bom_product_allergen

christian-ramos-tecnativa:
Module to extend BoM with product allergens
Depends on:

- [x] https://github.com/OCA/community-data-files/pull/248

@Tecnativa TT58394


mrp_bom_product_allergen/static/img/allergen_icons/Celery.png
pedrobaeza, line 1: Why putting again the images?

christian-ramos-tecnativa: Oh! I completely overlooked that, just removed them. Thanks for pointing it out!

mrp_bom_product_allergen/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def button_bom_allergens(self):
+        """Button to compute allergens from BoM."""
+        templates = self.filtered(
+            lambda t: t.product_variant_count == 1 and t.bom_count > 0
+        )
+        if templates:
+            return templates.mapped("product_variant_id").button_bom_allergens()
```
pedrobaeza:
```suggestion
            return templates.product_variant_id.button_bom_allergens()
```

mrp_bom_product_allergen/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+    _inherit = "product.product"
+
+    def get_bom_lines(self, bom=None):
+        """Get BoM lines recursively for the product."""
+        bom_lines_ids = self.bom_ids.bom_line_ids.ids
+        if not bom:
+            bom = []
+        bom += self.bom_ids.ids
+        if self.bom_ids.bom_line_ids.product_id.bom_ids.filtered(
+            lambda b: b.id not in bom
```
pedrobaeza: Do a set and you won't need this filtering.

mrp_bom_product_allergen/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def get_bom_lines(self, bom=None):
+        """Get BoM lines recursively for the product."""
+        bom_lines_ids = self.bom_ids.bom_line_ids.ids
```
pedrobaeza, line 12: Why using here `ids` instead of records? I suppose it's because what we talked about not iterating and adding records to recordsets, but it's not the same to go record per record, as this case that is recordset per recordset, so I think it's OK to add them as recordsets.

christian-ramos-tecnativa:
Ready for review/merge

christian-ramos-tecnativa:
Hmm CI not finding the `product_allergen` module

christian-ramos-tecnativa:
CI fixed

christian-ramos-tecnativa:
Doing some UI improves

christian-ramos-tecnativa:
image
Added Allergens column to the preview

pedrobaeza:
Please check CI

christian-ramos-tecnativa:
Test adapted to the new approach

eduezerouali-tecnativa (approved):
Code LGTM and tested on runboat

BhaveshHeliconia (approved):
Code and functional review LGTM!

quentinDupont (dismissed):
Hello,
Thanks for this work :)
That's funny, I write a similar module with the exact same name in v16 ;)
https://github.com/grap/grap-odoo-business/tree/16.0/mrp_bom_product_allergen
(Based on an other module to handle allergen)

I like the way this module works. I think it lacks the BoM part like in my module : in a "mrp_bom_product_allergen", i think we can add for the BoM itself the concatenation of the allergens in the BoM lines

What do you think ?

quentinDupont:
By the way, If you use BoM to handle recipes, this is the whole heart of our cooperative business, so you can check our modules. Some of them are already in OCA manufacture 16.0, or in PR https://github.com/OCA/manufacture/pulls/quentinDupont and especially in our custom repo : https://github.com/grap/grap-odoo-custom/tree/16.0

christian-ramos-tecnativa:
Hello @quentinDupont, what a coincidence! 😄 I added allergens to the BOM lines because I think it makes perfect sense.

Now that this is covered in OCA, we can join forces on migrations and further improvements!

flotho (approved):
code review, LGTM

christian-ramos-tecnativa:
Using the allergen records instead of the img to avoid adding the same one multiple times

pedrobaeza:
CI is red

christian-ramos-tecnativa:
CI pending of https://github.com/OCA/manufacture/pull/1817

pedrobaeza:
Please explore the alternative of `freeze_time`.

PR #1646, opened 2025-11-06
[17.0][ADD] quality_control_oca_test_question_formula module

arielbarreiros96:
This module serves doing automatic checks by defining a python logic at test line level.

It is useful for example for this use case:
We have to do a quality check for a picking, which has a very long list of products on it, we need to verify if all the lots reserved has at least X days remaining before end of life.

This same logic of automating the validation can be done for some other things as well.


edescalona (approved):
LGTM

diggy128:
Great work! However, allow me a few comments:
I've mentioned some of them in another PR where they add image as a variable type.

Variable types in QC can be one of the following two:
1) Quantitative
2) Qualitative or categorical

In the first case, the variable can take any numerical value. In the second, it can take a value from a predefined set of values (thus categories).

In the example you set of the picking, where there is a check for BBE dates for all products, it can either be true or false. This case is clearly a qualitative type variable. To be more precise, since it can take one of two possible values, it is a binary qualitative value.

Now, since you want to add automation to the calculation of the value (great feature, btw), my opinion is it should be implemented in another way:
- Keep the variable types as is
- Set the variable type as qualitative and add two values: True, False
- Add a check box "value_is_computed" in test.line that, when set, shows another "formula" text field that you type the custom formula into.
- The rest of the logic should be as you made it already.

This way, you can also create a computed value for a quantitative variable. E.g. "Average days to BBE from receiving" where you can get a numerical value that allows you to monitor the trend and prevent out of limits values beforehand. In this case, you can set an upper and lower limit in the test line itself (and show as valid values in inspection).

This is part of SPC that can be applied to many more areas and get valuable insights about a process's output metrics that can increase value or cause problems.

arielbarreiros96:
> Great work! However, allow me a few comments: I've mentioned some of them in another PR where they add image as a variable type.
> 
> Variable types in QC can be one of the following two:
> 
> 1. Quantitative
> 2. Qualitative or categorical
> 
> In the first case, the variable can take any numerical value. In the second, it can take a value from a predefined set of values (thus categories).
> 
> In the example you set of the picking, where there is a check for BBE dates for all products, it can either be true or false. This case is clearly a qualitative type variable. To be more precise, since it can take one of two possible values, it is a binary qualitative value.
> 
> Now, since you want to add automation to the calculation of the value (great feature, btw), my opinion is it should be implemented in another way:
> 
> * Keep the variable types as is
> * Set the variable type as qualitative and add two values: True, False
> * Add a check box "value_is_computed" in test.line that, when set, shows another "formula" text field that you type the custom formula into.
> * The rest of the logic should be as you made it already.
> 
> This way, you can also create a computed value for a quantitative variable. E.g. "Average days to BBE from receiving" where you can get a numerical value that allows you to monitor the trend and prevent out of limits values beforehand. In this case, you can set an upper and lower limit in the test line itself (and show as valid values in inspection).
> 
> This is part of SPC that can be applied to many more areas and get valuable insights about a process's output metrics that can increase value or cause problems.

Thanks for the feedback @diggy128 , I think this makes even more sense, I'll give it a though and come back around when it is ready.

I'm also thinking on a way for letting the user know why the test failed, basically adding a message. I mean, for example this example I mentioned, checking expiration date of a bunch of lots from a picking, I would become handy to have somewhere to see why the test is failing (if so), otherwise someone would have to manually check all of them to see what's going on, with a few only it may be acceptable, but I think with more than 20 recorords this starts to become impractical to handle by hand

diggy128:
> I'm also thinking on a way for letting the user know why the test failed, basically adding a message. I mean, for example this example I mentioned, checking expiration date of a bunch of lots from a picking, I would become handy to have somewhere to see why the test is failing (if so), otherwise someone would have to manually check all of them to see what's going on, with a few only it may be acceptable, but I think with more than 20 recorords this starts to become impractical to handle by hand

A couple more points here: 
1. Do you want to know why the inspection failed OR that specific inspection.line?
2. Do you REALLY want that consolidated result?

As we've been using quality control since v7.0 (back when NaN contributed, if I'm not mistaken), and now on v16, we tried many ways to make data entry  easier and the results more usable.

On point 2 above, I would strongly recommend automating the generation of inspections with as much more data as you can because: a) You can always consolidate data later, but you can't break up data that were consolidated .b) You get a MUCH better use of the data you already have entered (aka lot BBE) c) You can automate further actions depending on the result.

In your scenario, consider the following:
1) You receive 2 different products with 20 different lots each.
2) Out of them 2 lots of the second product are near the expiry date
3) As with your company policy, you can not use these products and want to return them to the supplier.

With the approach I mentioned above, the flow would be like so:
1) You receive the products
2) You enter the details in the picking and validate
3) An automated action creates one inspection for each lot received (2*20=40 tests), each one with one  line (BBE days to date), and runs the test validating the inspection
Now you know which lot is near expiry and you have  traceability per lot.

You could also augment the receive process further with automated actions. Eg
1) You can block using the lots failing inspection.
2) You can create a return picking to the supplier (so that this lot is reserved and not used in any other picking)
3) You can automatically print "DO NOT USE" stickers to apply on the packaging of the failing lots.
4) You can send an automated email (from the picking) to the supplier, taking data from the failing inspections to inform the supplier of the failing inspections.
5) You have COMPLETE traceability
6) You can use that test results to create KPIs for each supplier and use them for supplier evaluation
7) You are an ISO master and you use, consume, and keep data that proves that you OWN the process and you know what you do with them (more than just recording for compliance)

We apply all the above in our company, except n6, which will be done in a spreadsheet after moving to v18. 
It takes a lot of time, but it creates data you trust and data you can use in ways you didn't think of in the beginning (eg, points 3,4 above).

arielbarreiros96:
Hey @diggy128 take look at this new approach, now all questions remain either qualitative or quantitative.
But questions have a formula for self-computing its value. The formulas are also capable of returning a message, which gets posted on the internal notes, since the message is created on the formula, you ged to decide its format, you can craft whatever makes sense for you.

For the use case we've been trying to cover, this is fully functional the way it is now, what do you think?

diggy128:
Getting to it. At this point I've made a functional review in runboat, didn't get into the code.
In qc.test:
Auto-compute value code field gets very wide breaking layout. Other than that, everything OK. 
Excellent work on default code field content!

In qc.inspection:
Inspection lines that are auto-computed:
1. Should be visually separated from the ones that are to be manually filled (both in inspections and inspection lines actions).
2. Should be read only (Not even QC manager should be able to manually edit them, it beats the purpose of computing the value in the first place, doesn't it?). If there is a use case that should be editable please mention it. If truly needed, there should also be a boolean field in qc.test.line that would allow the result to be overridden.
3. Auto-computed results don't update when the values it depends are updated (eg when updating the quantity in qc.inspection)

Given points 2,3 above:
Maybe you should rethink "Recompute Manually Adjusted Lines" and dump all manual override in these lines. Allowing to override the computed values sets these values as auto-generated NOT computed.
Tracking all possible fields that could be used in computing the results is not possible. Maybe you should override write in qc.inspection and recompute the auto-computed lines. That way any change in qc.inspection will trigger the recompute.

Consider the following scenario:
1. You receive a picking
2. You create the inspection and mark as todo (state-> ready)
3. You discover you incorrectly entered the BBE in the lot and you correct it
4. You confirm the inspection
5. The lines are computed again with the correct BBE

Unfortunately nothing can be done if the inspection is done and THEN a value that the compute depends on gets changed (actually it can, but it in a context of a separate module as it will require additional logic).

arielbarreiros96:
> Getting to it. At this point I've made a functional review in runboat, didn't get into the code. In qc.test: Auto-compute value code field gets very wide breaking layout. Other than that, everything OK. Excellent work on default code field content!
> 
> In qc.inspection: Inspection lines that are auto-computed:
> 
> 1. Should be visually separated from the ones that are to be manually filled (both in inspections and inspection lines actions).
> 2. Should be read only (Not even QC manager should be able to manually edit them, it beats the purpose of computing the value in the first place, doesn't it?). If there is a use case that should be editable please mention it. If truly needed, there should also be a boolean field in qc.test.line that would allow the result to be overridden.
> 3. Auto-computed results don't update when the values it depends are updated (eg when updating the quantity in qc.inspection)
> 
> Given points 2,3 above: Maybe you should rethink "Recompute Manually Adjusted Lines" and dump all manual override in these lines. Allowing to override the computed values sets these values as auto-generated NOT computed. Tracking all possible fields that could be used in computing the results is not possible. Maybe you should override write in qc.inspection and recompute the auto-computed lines. That way any change in qc.inspection will trigger the recompute.
> 
> Consider the following scenario:
> 
> 1. You receive a picking
> 2. You create the inspection and mark as todo (state-> ready)
> 3. You discover you incorrectly entered the BBE in the lot and you correct it
> 4. You confirm the inspection
> 5. The lines are computed again with the correct BBE
> 
> Unfortunately nothing can be done if the inspection is done and THEN a value that the compute depends on gets changed (actually it can, but it in a context of a separate module as it will require additional logic).

Hello  @diggy128 thanks for the feedback. I'll see what I can do on the UI.
In relation to functionality, there is a field (hidden) called manual_override that keeps track of the fields changed manually at line level.

My logic is that, at any point, the user should be able to override any computation, that's particularly useful when working with qualitative options with blurry/overlaping edges, there human judgement should play a role in the decision. If at any point the user overrides a computed result, there is a button at the top for recomputing, that's the idea at least :sweat_smile: if you do not see the button or something let me know

There is not really any tracking of dependencies if anything changes, basically there is only 2 ways to trigger the computation: by setting the qc to ready, and manually

arielbarreiros96:
@BinhexTeam T14770

PR #1649, opened 2025-11-11
[18.0][MIG] mrp_packaging_default: Migration to 18.0

Xino61122:
Migration at v16 to v18.

Video:
https://www.loom.com/share/b1c4f3096ca34bdb8a62b665e3f7030a

@moduon MT-11814


mrp_packaging_default/models/mrp_bom_line.py
```diff
@@ -27,11 +27,6 @@ class MrpBomLine(models.Model):
     def _compute_product_packaging(self):
         """Set the appropriate packaging for the product qty."""
         for one in self:
-            one.product_packaging_id = (
-                one.product_id.packaging_ids._find_suitable_product_packaging(
-                    one.product_qty, one.product_uom_id
-                )
-            )
```
yajo: Instead of moving this logic to product onchange, could you please test this?

```suggestion
            one.product_packaging_id = (
                one.product_packaging_id._find_suitable_product_packaging(
                    one.product_qty, one.product_uom_id
                ) or
                one.product_id.packaging_ids._find_suitable_product_packaging(
                    one.product_qty, one.product_uom_id
                )
            )
```

If this works, it's better to keep logic inside the compute method, because it works for multi-records and works for direct ORM operations.

mrp_packaging_default/models/mrp_bom_line.py
```diff
@@ -65,3 +60,8 @@ def _onchange_product_set_qty_from_packaging(self):
         if default_packaging:
             self.product_uom_id = default_packaging.product_uom_id
             self.product_qty = default_packaging.qty
+            self.product_packaging_id = (
+                self.product_id.packaging_ids._find_suitable_product_packaging(
+                    self.product_qty, self.product_uom_id
+                )
+            )
```
yajo: If the above comment works, then remove these lines.

yajo (changes requested):
Code review.

Functionally looks OK, but please @Gelojr review.

mrp_packaging_default/models/mrp_bom_line.py
Xino61122: It works perfectly with your changes.
I'll upload the changes.
Thank you very much. :+1:

Gelojr (approved):
Congratulations on the work done.
The following tests have been executed successfully.

- Test 1: Products correctly configured with multiple packaging options and saved without issues.
- Test 2: Default packaging proposed on BoM line and quantities recalculated correctly when changing packaging or packaging qty.
- Test 3: Kit exploded on delivery and component moves inherited packaging and packaging qty as defined in the BoM.
- Test 4: Manufacturing order created by replenishment rule inherited packaging and packaging qty from the BoM.
- Test 5: Changing the production quantity recalculated component quantities and packaging qty proportionally and correctly.
- Test 6: Manually created manufacturing order showed packaging and packaging qty on components as defined in the BoM.

yajo (changes requested):
Commits history is lost. Could you restore it please?

yajo:
This was replaced by https://github.com/OCA/manufacture/pull/1676 and can be closed.

PR #1653, opened 2025-11-12
[17.0][FIX] mrp_production_back_to_draft: not move_raw_ids

david-s73:
- This case occurs when you try to revert a 'mrp.production' to draft when you don't have components or 'move_raw_ids', since then it doesn't modify the status of the 'mrp.production' and it remains as confirmed, then the raise UserError is triggered because 'rec.state != ‘draft’'.

image


Reyes4711-S73 (approved):
LGTM

sbiosca-s73 (approved):
LGTM

PR #1659, opened 2025-11-18
[18.0][MIG] mrp_primecontractor_raw_material: migration to 18.0

FranzPoize:


FranzPoize:
Test are passing locally not sure which modules makes it fail in CI

PR #1660, opened 2025-11-18
[18.0][MIG] mrp_subcontractor: Migration to 18.0

FranzPoize:


Kev-Roche (approved):
LGTM

FranzPoize:
can you take a look @clementmbr

clementmbr:
@FranzPoize sorry for the delay.
The module only adds a `has_owner` boolean on bom lines?

It doesn't seem useful, isn't it better to just remove the module?
Or at least it needs explanations on the readme to understand how it could be used by other modules.

But it looks like the boolean `primecontractor_raw_material` in module mrp_primecontractor_raw_material does the same job as `has_owner` here, but being used to receive the raw material from the prime contractor location.

PR #1662, opened 2025-11-19
[15.0][ADD] mrp_bom_assign_auto: New module

eduezerouali-tecnativa:
cc @Tecnativa TT59293
ping @sergio-teruel @victoralmau 

This module verifies all available BoMs for a given product and automatically selects the first BoM for which all required components are currently in stock. If no BoM meets the availability criteria, the module gracefully falls back to Odoo’s default BoM selection logic.


pedrobaeza:
Check CI

mrp_bom_assign_auto/models/mrp_bom.py
```diff
@@ -0,0 +1,61 @@
...
+        boms = self.search(domain, order="sequence, product_id, id")
+        quant = self.env["stock.quant"]
+        location = self.env.ref("stock.stock_location_stock")
+        for bom in boms:
+            if not bom:
+                continue
+            for line in bom.bom_line_ids:
+                required = line.product_qty
+                component = line.product_id
+                available = quant._get_available_quantity(
```
carlosdauden: It may be better to load the availability of all products at once to obtain better performance.
See how this is done in the production order:
https://github.com/odoo/odoo/blob/fb4c85e007ee70e95e95197d358f1cde19b82338/addons/stock/models/stock_move.py#L440

Then decide (either you or add an option for the user to do so) whether to take into account free_qty, virtual_available, or another option.

carlosdauden (changes requested):

mrp_bom_assign_auto/models/mrp_bom.py
```diff
@@ -0,0 +1,64 @@
...
+        boms = self.search(domain, order="sequence, product_id, id")
+        bom_products = list(set(boms.bom_line_ids.mapped("product_id").ids))
+        product_available = (
+            self.env["product.product"].browse(bom_products).read(["virtual_available"])
+        )
+        product_available_map = {
+            rec["id"]: rec["virtual_available"] for rec in product_available
+        }
+        for bom in boms:
+            if not bom:
```
carlosdauden: When does this condition occur within the for?

carlosdauden (changes requested):

mrp_bom_assign_auto/models/mrp_bom.py
eduezerouali-tecnativa: will never occur, that was code left behind.

mrp_bom_assign_auto/models/mrp_bom.py
eduezerouali-tecnativa: At the moment, virtual_available is hard-coded, but it might be useful to make this a configurable setting for the user. What do you think?

carlosdauden: Yes, it is better to add an option for the user to decide.
The default option should be free_qty.

eduezerouali-tecnativa:
ping @sergio-teruel @carlosdauden

eduezerouali-tecnativa:
@carlosdauden @sergio-teruel please could you review?

carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza:
Trying after #1815

/ocabot merge nobump

PR #1673, opened 2025-12-03
[18.0][IMP] mrp_restrict_lot: add restrict_lot to finished_move_ids

hparfr:
Before this change: when a MO is created with a lot_producing_id, restrict_lot_id is not propagated to the move_finished_ids

### DIFF
--- a/mrp_restrict_lot/models/mrp_production.py
+++ b/mrp_restrict_lot/models/mrp_production.py
@@ -7,14 +7,25 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     lot_producing_id = fields.Many2one(
-        compute="_compute_lot_producing_id", store=True, readonly=False
+        compute="_compute_lot_producing_id",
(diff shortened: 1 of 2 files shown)


StefanRijnhart (approved):
Nice one!

hparfr:
@florian-dacosta please review

florian-dacosta (approved):

mrp_restrict_lot/models/mrp_production.py
```diff
@@ -18,3 +21,11 @@ def _compute_lot_producing_id(self):
             ).restrict_lot_id
             if finish_move_lot:
                 order.lot_producing_id = finish_move_lot
+
+    def _inverse_lot_producing_id(self):
+        for order in self:
+            finish_move_lot = order.move_finished_ids.filtered(
```
florian-dacosta: Would be better to rename this variable finish_move

PR #1454, opened 2025-01-13
[17.0][MIG] mrp_subcontracting_skip_no_negative: Migration to 17.0

david-s73:

PR #1372, opened 2024-10-23
[14.0][IMP] mrp_subcontracting_skip_no_negative: Show negative stock error of subcontracting component

victoralmau:
Show negative stock error of subcontracting component

@Tecnativa TT50668

PR #1674, opened 2025-12-04
[18.0][MIG] mrp_subcontracting_skip_no_negative: Migration to 18.0

AungKoKoLin1997:
@qrtl QT6162


AungKoKoLin1997:
I found that the commit introduced during the 17.0 migration (https://github.com/OCA/manufacture/pull/1454#issuecomment-2901826580) not only affects partial subcontracting receipts, but also skips the check from that PR (https://github.com/OCA/manufacture/pull/1372) when the component quantity is insufficient. Therefore, in my latest commit, I decided to remove the code I originally added in my first PR (https://github.com/OCA/manufacture/pull/1025), since it is not related to the no negative stock module and the scenario that motivated it no longer occurs in Odoo 18 (https://github.com/OCA/manufacture/pull/1025#discussion_r1361740224).

kanda999 (approved):
Functional review: It works as expected in my local env

yostashiro (commented):
@AungKoKoLin1997 Would you consider adding a test where subcontracting picking receipt fails when component stock is insufficient?

AungKoKoLin1997:
In my last commit, I bring back the fix from odoo 16 that was not included in17.0 migration and improve tests.
https://github.com/OCA/manufacture/commit/19343f6b78ddc2323186323f0f76ed410b86c9ea

I believe my latest two commit should be back ported to 17.0.

mrp_subcontracting_skip_no_negative/tests/test_mrp_subcontracting_skip_no_negative.py
```diff
@@ -0,0 +1,152 @@
...
+        self.env["stock.quant"].create(
+            {
+                "product_id": product.id,
+                "location_id": self.subcontractor_location.id,
+                "quantity": qty,
+            }
+        )
+
+    @mute_logger("odoo.models.unlink")
+    def test_mrp_subcontracting_skip_no_negative_01(self):
```
yostashiro: We may perhaps make test methods a bit more descriptive instead of having trailing numbers.

yostashiro (approved):
Code review. LGTM.

kanda999 (approved):

AungKoKoLin1997:
@victoralmau @fdesmottes @Raul-S73 
Please let me know if you have a different opinion on the comment above.
https://github.com/OCA/manufacture/pull/1674#issuecomment-3610481862

fdesmottes:
> @victoralmau @fdesmottes @Raul-S73 Please let me know if you have a different opinion on the comment above. [#1674 (commentaire)](https://github.com/OCA/manufacture/pull/1674#issuecomment-3610481862)

Hi @AungKoKoLin1997 it's ok for me.

mrp_subcontracting_skip_no_negative/tests/test_mrp_subcontracting_skip_no_negative.py
```diff
@@ -0,0 +1,152 @@
...
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import Command
+from odoo.exceptions import ValidationError
+from odoo.tests import Form, TransactionCase
+from odoo.tools import mute_logger
+
+
+class TestMrpSubcontractingSkipNoNegative(TransactionCase):
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon
from odoo.tools import mute_logger


class TestMrpSubcontractingSkipNoNegative(BaseCommon):
```

victoralmau (approved):
Code review OK

Can you merge the extra commits into the migration commit?

ejemplo

AungKoKoLin1997:
> Can you merge the extra commits into the migration commit?

@victoralmau 
I want to keep them as separate commits because I plan to backport these commits to 17.0 later. Also, the last commit already exists as a separate commit in Odoo 16.
https://github.com/OCA/manufacture/pull/1674#issuecomment-3635346332

mrp_subcontracting_skip_no_negative/tests/test_mrp_subcontracting_skip_no_negative.py
AungKoKoLin1997: My understanding is that `BaseCommon` is an internal helper built on top of `TransactionCase`, mainly used to standardize core tests (such as a mail-disabled context and common user/company setup).
Since our test doesn’t rely on those features, I prefer using `TransactionCase` to keep it simpler and more explicit.

pedrobaeza: The suggestion is for speeding up tests. When you create a record, some chatter messages are added, some followers are auto-added, etc. Using this class, all that things are disabled by default. It's true there's a little overhead with the creation of some users, but it usually worths. Alternatively, you can do `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`.

AungKoKoLin1997: @pedrobaeza Thank you for the explanation.
I applied `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))` in the tests.

victoralmau: Sorry, it's not just those context values, it's more https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/tests/common.py#L11. That's why we always suggest using that `BaseCommon` class.

AungKoKoLin1997: I used `BaseCommon` class.

KDRdev (approved):

victoralmau (approved):
Code review OK, changes seem coherent

PR #1677, opened 2025-12-10
[18.0][FIX] mrp_lot_number_propagation: Clean outdated finished serial number

grindtildeath:
When a manufacturing order is split through the produce all wizard, the serial numbers of the finished product are created and written on each backorder MO, before move lines are unlinked from the original MO and created again on the backorders.

As long as nothing changes on the propagating component, we are fine. But in case the component is scrapped or unreserved, and another component serial number was assigned, we had a discrepancy between the propagating lot number from the component and the lot number of the finished product.

Since this module creates and assigns the finished serial number during the call to button_mark_done, we could have wiped the serial number as soon as the move line propagating its serial number was unassigned. But in that case Odoo would still have created a stock.lot record for each backorder that would have been correct in most cases, just to be recreated afterwards.

Instead, we will clean the outdated serial number of the finished product only in case a propagating move with a different serial number is assigned to the manufacturing order, as well as protecting against any discrepancy during the when calling button_mark_done.

### DIFF
--- a/mrp_lot_number_propagation/models/mrp_production.py
+++ b/mrp_lot_number_propagation/models/mrp_production.py
@@ -186,7 +186,11 @@ class MrpProduction(models.Model):
 
     def _create_and_assign_propagated_lot_number(self):
         for order in self:
-            if not order.is_lot_number_propagated or order.lot_producing_id:
+            if (
+                not order.is_lot_number_propagated
+                or order.lot_producing_id
+                and order.lot_producing_id.name == order.propagated_lot_producing
+            ):
                 continue
             finish_moves = order.move_finished_ids.filtered(
                 lambda mv, mo=order: mv.product_id == mo.product_id
--- a/mrp_lot_number_propagation/models/stock_move.py
+++ b/mrp_lot_number_propagation/models/stock_move.py
@@ -11,3 +11,25 @@ class StockMove(models.Model):
         default=False,
         readonly=True,
     )
+
+    def _action_assign(self, force_qty=False):
+        res = super()._action_assign(force_qty=force_qty)
+        self._ensure_production_propagating_finished_lot()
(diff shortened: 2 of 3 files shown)


mrp_lot_number_propagation/models/mrp_production.py
```diff
@@ -186,7 +186,11 @@ def pre_button_mark_done(self):
 
     def _create_and_assign_propagated_lot_number(self):
         for order in self:
-            if not order.is_lot_number_propagated or order.lot_producing_id:
+            if (
+                not order.is_lot_number_propagated
+                or order.lot_producing_id
+                and order.lot_producing_id.name == order.propagated_lot_producing
```
florentx, line 192: I wonder if this test could be:

```suggestion
                or order.lot_producing_id.name == order.propagated_lot_producing
```

florentx (approved):

mrp_lot_number_propagation/models/stock_move.py
```diff
@@ -11,3 +11,18 @@ class StockMove(models.Model):
...
+            finished_lot = move.raw_material_production_id.lot_producing_id
+            if (
+                move.propagate_lot_number
+                and finished_lot
+                and len(move.move_line_ids) == 1
+                and move.move_line_ids.lot_id.name != finished_lot.name
+            ):
+                move.raw_material_production_id.write({"lot_producing_id": False})
+                if not finished_lot.quant_ids:
+                    finished_lot.unlink()
```
sebalix, line 35: Should we move that part in a dedicated method called from `_action_assign()`?

sebalix (approved):
Small comment, LG

henrybackman (approved):

twalter-c2c (approved):

PR #1682, opened 2025-12-19
[MIG] `mrp_multi_level_consume_safety_stock`: Migration to 18.0

Camille0907:
Migration of `mrp_multi_level_consume_safety_stock` to 18.0 (from https://github.com/OCA/manufacture/pull/1169)

---

Also add a small fix into `mrp_multi_level`: Ensure the safety stock is rebuilt as soon as the target date is reached, even if a supply comes a few days after (Needed for `mrp_multi_level_consume_safety_stock` tests to pass)

This completes https://github.com/OCA/manufacture/commit/57cba495:

If I have 0 units, my safety stock is 5 units and its target date is tomorrow, the procurement recommendation should be 5 units for tomorrow, even if I have a resupply of 5 the day after tomorrow.


mrp_multi_level/wizards/mrp_multi_level.py
LoisRForgeFlow, line 1: @Camille0907 This is outside of the migration, can you add a test case for this?

vvrossem: From the [commit](https://github.com/OCA/manufacture/pull/1682/commits/8fc61e3d05eb9c9290f4ec46345c1f70649587e5) message, a test case already covers this, but was previously failing (TBC)

vvrossem (approved):

mrp_multi_level/wizards/mrp_multi_level.py
Camille0907: Hello, yes I confirm there is a test case inside ``mrp_multi_level_consume_safety_stock``, which fails without this change

I understand we might want a test case directly in ``mrp_multi_level``, and I tried to add one, but I have not been able to find a relevant one: the only thing I manage to do was to have nearly the entire ``mrp_multi_level_consume_safety_stock`` methods mocked into the tests of ``mrp_multi_level`` to trigger a relevant case... That looked very ugly to me so I got rid of it :shrug:

BhaveshHeliconia (approved):
Functional review LGTM!

mrp_multi_level/wizards/mrp_multi_level.py
LoisRForgeFlow: Hi, thanks for checking. I have reviewed and I could not find a change in the behavior of the base module, so let's move forward with this change as is.

LoisRForgeFlow (approved):

LoisRForgeFlow:
Fwport of the base module patch in https://github.com/OCA/manufacture/pull/1762.

PR #1691, opened 2026-01-13
[19.0][MIG] account_move_line_mrp_info

JasminSForgeFlow:
Standard Migration

Remove dependency of account_move_line_stock_info as Odoo 19 has refactored stock valuation

@ForgeFlow


account_move_line_mrp_info/hooks.py
```diff
@@ -0,0 +1,52 @@
+import logging
+
+from psycopg2 import sql
+
+
+def pre_init_hook(env):
```
AaronHForgeFlow: It seems to me this hook is not necessary, it is referencing the stock_move table and there is no account move line related to stock moves

AaronHForgeFlow (approved):
Functional + code LGTM

account_move_line_mrp_info/hooks.py
JasminSForgeFlow: You are right, data is already there.

I have removed hook

BhaveshHeliconia (approved):
LGTM!

JordiMForgeFlow (approved):

PR #1693, opened 2026-01-14
[18.0][IMP] mrp_multi_level_estimate: allow to ignore estimates

LoisRForgeFlow:
This module is auto-installed but sometimes you need to ignore estimates, so adding the possibility via configuration is useful.


PaulaMForgeFlow (approved):
Functional test OK

PR #1706, opened 2026-01-31
[18.0] [FIX] quality_control_oca: Move div inside group 

diggy128:
in qc.test.question.form

Having the div just inside the form makes adding extra fields in the form view ugly. This fixes this by moving the quantitative related fields inside a group that preserves div for the existing fields and allows to properly add extra fields. Breaking change: The same attribute value for name (quantitative) that was used for the div is now used for the group so if there where any ugly hacks to add extra fields inside the div, these will require changing the path

### DIFF
--- a/quality_control_oca/views/qc_test_view.xml
+++ b/quality_control_oca/views/qc_test_view.xml
@@ -96,37 +96,39 @@
                         </list>
                     </field>
                 </group>
-                <div
+                <group
                     name="quantitative"
-                    align="center"
+                    colspan="4"
                     invisible="type != 'quantitative'"
                 >
-                    <h1 name="quantitative-data">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Screenshot of the before/after?

diggy128:
**BEFORE**
before
**AFTER**
after

You can see that the two extra fields sample_size and c that are placed inside the quantitative group, in the after image are properly displayed inside the group, but are without label and centered (as the div has align="center") before applying this change.

pedrobaeza (approved):

flotho (approved):
LGTM,
Thanks

diggy128:
@pedrobaeza 
Can you please merge this too?

PR #1711, opened 2026-02-06
[18.0][ADD] mrp_bom_line_uom_rounding: UoM rounding validation on BoM

JoanSForgeFlow:
When a component is added to a BoM, the system validates that the quantity respects the rounding precision defined in the Unit of Measure.


LoisRForgeFlow (approved):
👍

ThiagoMForgeFlow (approved):

PR #1716, opened 2026-02-10
[18.0][IMP] quality_control_oca: remove unused object_id on qt_test

remi-filament:
Forward port of #1568 from 17.0 to 18.0
As reported in issue https://github.com/OCA/manufacture/issues/1378 qc_test type field allows for display of object_id for which is selection is set(), creating invalid display.
Since this field does not seem to be used anywhere (none of the depending modules are using it neither), I propose with this PR to remove both object_id and type from qc_test.

Note : a field with same name object_id is used on qc_inspection model and make a lot of sense there !

### DIFF
--- a/quality_control_oca/demo/quality_control_demo.xml
+++ b/quality_control_oca/demo/quality_control_demo.xml
@@ -2,7 +2,6 @@
 <odoo noupdate="1">
     <record model="qc.test" id="qc_test_1">
         <field name="name">Generic Test (demo)</field>
-        <field name="type">generic</field>
         <field name="active" eval="True" />
         <field name="category" ref="qc_test_template_category_generic" />
     </record>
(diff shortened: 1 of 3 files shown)


diggy128 (approved):
LGTM!
I think this field was left over from the original NaN module.

Hugo-Trentesaux (approved):

Saran440 (approved):
Code Review 👍

dreispt (approved):

Saran440:
@dreispt Hi, this PR is ready to merge. Could you please help merge it?

pedrobaeza (changes requested):
Please put the proper label, as `[REM]` is for removing an addon, which is not the case. If you think it, in any PR you may remove some lines of code as part of your patch. You can use `[IMP]`, as it's not either a bugfix.

pedrobaeza (approved):
/ocabot merge minor

remi-filament:
Thanks @pedrobaeza I just renamed both PR and commit

PR #1569, opened 2025-07-21
[17.0][IMP] quality_control_oca: display archived ribbon

remi-filament:
Display archived ribbon for coherency with other Odoo views instead of active checkbox in form view

PR #1717, opened 2026-02-10
[18.0][IMP] quality_control_oca: display archived ribbon

remi-filament:
Forward port of #1569 from 17.0 to 18.0
Display archived ribbon for coherency with other Odoo views instead of active checkbox in form view


pedrobaeza (approved):
Straight fw-port, so let's fast-track it:

/ocabot merge minor

PR #1719, opened 2026-02-10
[18.0][IMP] quality_control_oca: select test from inspection form view

remi-filament:
Forward port of #1571 

The form view was using a button on top to set test on inspection.
The same can be done by having test field settable on the form view, improving a lot the UX.

This PR allows to set the test directly on the view, setting inspection lines depending on the test selected, and thus removing useless wizard.


quality_control_oca/views/qc_inspection_view.xml
```diff
@@ -76,11 +66,17 @@
...
+                            <field name="test" readonly="state != 'draft'" />
                             <field name="user" />
                             <field name="object_id" readonly="state != 'draft'" />
                             <field name="qty" />
                             <field name="product_id" />
+                            <field
+                                name="company_id"
+                                readonly="state != 'draft'"
+                                groups="base.group_multi_company"
+                            />
```
diggy128: Since there are already many fields on the left hand group, consider moving the field to the right-hand side group along with user field, below the date fields to be similar to other Odoo views (mrp.production, repair.order, etc)

diggy128:
@remi-filament Maybe take a look at this too if you find some time..

PR #1725, opened 2026-02-13
[17.0][REF] mrp_production_serial_matrix

DavidJForgeFlow:
Moving the matrix from a wizard to an standard model. This way we can fill it up in multiple sessions and check previous executions. Also added hooks to be used in the glue modules (for example queue_jobs). We also notify now all the errors during the execution in the chatter, this way we ensure that the user has the required information to fix all the issues.

Increased test coverage to ensure the correct behavior. 

Some of the changes were done thinking in https://github.com/OCA/manufacture/pull/1726 and https://github.com/OCA/manufacture/pull/1727


LoisRForgeFlow (approved):
Functional review, LGTM

JoanSForgeFlow (approved):

PR #1729, opened 2026-02-17
[18.0][MIG] mrp_progress_button: Migration to 18.0

Killance:


mrp_progress_button/tests/test_progress_button.py
```diff
@@ -0,0 +1,75 @@
+# Copyright (C) 2017 Akretion (http://www.akretion.com). All Rights Reserved
+# @author Florian DA COSTA <florian.dacosta@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields
+from odoo.tests.common import TransactionCase
+
+
+class TestProgressButton(TransactionCase):
+    def setUp(self, *args, **kwargs):
```
Tisho99, line 10: I think this should be changed by setupClass

mrp_progress_button/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# @author Florian DA COSTA <florian.dacosta@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Mrp Progress Button",
+    "summary": "Add a button on MO to make the MO state 'In Progress'",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/manufacture",
+    "category": "Manufacturing",
+    "version": "18.0.2.0.0",
```
Tisho99, line 11: Version should be "18.0.1.0.0"

mrp_progress_button/tests/test_progress_button.py
```diff
@@ -2,32 +2,31 @@
...
-        self.stock_location_stock = self.env.ref("stock.stock_location_stock")
+        self.stock_location_stock = self.env.ref("stock.stock_location_locations")
         self.manufacture_route = self.env.ref("mrp.route_warehouse0_manufacture")
         self.uom_unit = self.env.ref("uom.product_uom_unit")
 
         self.product_manuf = self.env["product.product"].create(
             {
                 "name": "Manuf",
-                "type": "product",
+                "type": "consu",
```
Tisho99, line 21: I think that the is_storable should also be added, shouldn't it?

mrp_progress_button/models/mrp_production.py
```diff
@@ -2,38 +2,48 @@
...
+        "move_raw_ids.quantity",
         "move_finished_ids.state",
         "workorder_ids.state",
         "product_qty",
         "qty_producing",
         "date_start",
     )
     def _compute_state(self):
-        res = super()._compute_state()
+        super()._compute_state()
```
Tisho99, line 22: I think this return should not be removed because you should return what you get from super()

mrp_progress_button/models/mrp_production.py
```diff
@@ -2,38 +2,48 @@
...
         for production in self:
-            if production.state == "confirmed" and production.date_start:
+            ctx = dict(production.env.context)
+            if ctx.get("previous_state") == "confirmed" and production.date_start:
                 production.state = "progress"
-        return res
+            elif (
+                ctx.get("previous_state") == "progress"
```
Tisho99, line 28: I haven't found this context key in standard Odoo. 

I think that the reason to add this logic should be expliained, as it seems very custom

Tisho99 (changes requested):
Hello @Killance 

I have been reviewing this and found a few lines of code that could be improved.

The migration is also missing the "[IMP] $module: pre-commit auto fixes" commit.

Most of the guidelines are explained here: https://github.com/OCA/maintainer-tools/wiki#migration

As this PR has been inactive for 2 months, I will try to finish it in another PR unless you let me know that you would prefer to continue working on it.

Thank you for the migration.

PR #1740, opened 2026-02-25
[19.0][IMP] mrp_multi_level: add procure button in list header

LoisRForgeFlow:
Fwport of #1739


LoisRForgeFlow:
merging with nobump because we just merged the migration

/ocabot merge nobump

PR #1745, opened 2026-03-02
[18.0] [FIX] quality_control_oca: Handle multiple unlink

diggy128:
- Fix singleton error when selecting multiple records to delete.
- Provide a clear reason for failing along with record names that produced the error


diggy128:
@OCA/manufacturing-maintainers 
A quick review?

quality_control_oca/models/qc_inspection.py
```diff
@@ -119,14 +119,28 @@ def write(self, vals):
...
     @api.ondelete(at_uninstall=False)
     def _unlink_except_autogenerated_and_non_draft(self):
-        if self.auto_generated and not self.env.is_superuser():
+        auto_generated_records = self.env["qc.inspection"]
+        non_draft_records = self.env["qc.inspection"]
+        for record in self:
+            if record.auto_generated and not record.env.is_superuser():
+                auto_generated_records |= record
+            if record.state != "draft":
+                non_draft_records |= record
```
remi-filament: instead of a loop you could get both list with a .filtered() and do only once self.env.is_superuser() instead of doing it for each record

quality_control_oca/models/qc_inspection.py
```diff
@@ -119,14 +119,28 @@ def write(self, vals):
...
+                    ", ".join(auto_generated_records.mapped("display_name")),
+                )
             )
-        if self.state != "draft":
+        if len(non_draft_records) > 0:
             raise exceptions.UserError(
                 self.env._(
-                    "You cannot remove an inspection that is not in draft state."
+                    "Inspections %s are not in draft state and can only be deleted "
+                    "by superuser.",
```
remi-filament: Actually I do not think these could even be deleted by superuser since there is no check whether we are superuser or not so it will probably ends up with the same UserError even if superuser...

remi-filament (changes requested):
Hi @diggy128 thanks for your PR,
a few comments inline

remi-filament:
Maybe you could also join the 2 errors in one so that the user get the full list of failing inspection tests (otherwise he may remove the auto-generated ones from this list to still get an error with non draft ones)

quality_control_oca/models/qc_inspection.py
diggy128: Done

quality_control_oca/models/qc_inspection.py
diggy128: True, indeed. Fixed

diggy128:
@remi-filament 
Done. I also updated the tests. Can you please review?

diggy128:
Kind reminder @remi-filament

DantePereyra (approved):
Only code review. LGTM

PR #1754, opened 2026-03-16
Draft: [14.0][IMP] Make the serial matrix able to produce multiple quantities partially

ddejong-therp:
Hello,

Here are some changes to the `mrp_production_serial_matrix` module that make it possible to use the serial matrix wizard to produce multiple quantities at once, each with one of the selected serials, but allowing a part of the quantity to be produced when one fails.

So as an example, when you would like to produce an MO with a quantity of 10, and the 7th product fails to be produced, you would still end up with 6 produced orders, and a backorder for the remaining 4 products.
A message about the exception that just occurred is posted on the backorder.

Now, since this may not be a feature everyone would want to use, I've added a checkbox on the mrp settings page so that it can be enabled if desired.

The real reason that we wanted this implemented, is actually so that the transaction that produces multiple products (with serials), which can become pretty large, doesn't have to lock up Odoo because db tables are locked up for a long period of time.
So after each order is produced for a single product, the db transaction is committed.

### DIFF
--- a/mrp_production_serial_matrix/models/res_config_settings.py
+++ b/mrp_production_serial_matrix/models/res_config_settings.py
@@ -1,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    mrp_serial_matrix_allow_exceptions = fields.Boolean(
+        "Allow manufacturing exceptions",
+        config_parameter="mrp_production_serial_matrix.mrp_serial_matrix_allow_exceptions",
+    )
--- a/mrp_production_serial_matrix/static/description/index.html
+++ b/mrp_production_serial_matrix/static/description/index.html
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
@@ -9,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
@@ -275,7 +275,7 @@ pre.literal-block, pre.doctest-block, pre.math, pre.code {
   margin-left: 2em ;
   margin-right: 2em }
 
-pre.code .ln { color: grey; } /* line numbers */
+pre.code .ln { color: gray; } /* line numbers */
 pre.code, code { background-color: #eeeeee }
 pre.code .comment, code .comment { color: #5C6576 }
 pre.code .keyword, code .keyword { color: #3B0D06; font-weight: bold }
@@ -301,7 +301,7 @@ span.option {
 span.pre {
   white-space: pre }
 
-span.problematic {
+span.problematic, pre.problematic {
   color: red }
 
 span.section-subtitle {
@@ -450,7 +450,9 @@ If you spotted it first, help us to smash it by providing a detailed and welcome
 <div class="section" id="maintainers">
 <h2><a class="toc-backref" href="#toc-entry-7">Maintainers</a></h2>
 <p>This module is maintained by the OCA.</p>
-<a class="reference external image-reference" href="https://odoo-community.org"><img alt="Odoo Community Association" src="https://odoo-community.org/logo.png" /></a>
+<a class="reference external image-reference" href="https://odoo-community.org">
+<img alt="Odoo Community Association" src="https://odoo-community.org/logo.png" />
+</a>
 <p>OCA, or the Odoo Community Association, is a nonprofit organization whose
 mission is to support the collaborative development of Odoo features and
 promote its widespread use.</p>
--- a/mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
+++ b/mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
@@ -1,9 +1,13 @@
 # Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import logging
+import threading
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.float_utils import float_compare, float_is_zero
+
+_logger = logging.getLogger(__name__)
 
 
 class MrpProductionSerialMatrix(models.TransientModel):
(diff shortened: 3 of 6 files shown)


mrp_production_serial_matrix/models/res_config_settings.py
```diff
@@ -0,0 +1,25 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    allow_manufacturing_exceptions = fields.Boolean("Allow manufacturing exceptions")
```
thomaspaulb: This field name needs to be prefixed by something like `mrp_serial_matrix_` so that it won't conflict with other modules. Eg `mrp_serial_matrix_allow_exceptions`

mrp_production_serial_matrix/models/res_config_settings.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def get_values(self):
+        res = super().get_values()
+        value = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("mrp_production_serial_matrix.allow_manufacturing_exceptions")
+        )
+        res["allow_manufacturing_exceptions"] = value
+        return res
```
thomaspaulb: You don't have to do this, there is a shortcut:

```
    my_custom_setting = fields.Boolean(
        string="Enable My Custom Feature",
        config_parameter='my_module.my_custom_setting'
    )
```

mrp_production_serial_matrix/static/description/index.html
```diff
@@ -301,7 +301,7 @@
 span.pre {
   white-space: pre }
 
-span.problematic {
+span.problematic, pre.problematic {
```
thomaspaulb, line 304: Changes like this need to be branched into a separate `pre-commit fixes` commit, and the functional changes in another squashed commit

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
```diff
@@ -223,61 +226,96 @@ def button_validate(self):
...
+                # somewhere, since it is ok to only produce a part of the
+                # quantity.
+                self.env.cr.commit()  # pylint: disable=invalid-commit
+                finished_count += 1
+                if no_backorders:
                     break
+        except Exception as e:
+            exceptions_allowed = self.env["ir.config_parameter"].get_param(
+                "mrp_production_serial_matrix.allow_manufacturing_exceptions"
+            )
```
thomaspaulb: This config parameter should already be fetched outside of the loop, not upon catching the exception itself.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
```diff
@@ -223,61 +226,96 @@ def button_validate(self):
...
+                    return {
+                        "res_id": backorder_ids[0].id,
+                        "name": _("Manufacturing Order"),
+                        "view_mode": "form",
+                        "res_model": "mrp.production",
+                        "type": "ir.actions.act_window",
+                    }
             else:
-                break
+                raise e
```
thomaspaulb: Cleaner is to do 

```
if not exceptions_allowed:
    raise
# handle exception
```

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
```diff
@@ -223,61 +226,96 @@ def button_validate(self):
...
+                    no_backorders = True
+
+                # Commit changes in case processing any of the MO's fails
+                # somewhere, since it is ok to only produce a part of the
+                # quantity.
+                self.env.cr.commit()  # pylint: disable=invalid-commit
+                finished_count += 1
+                if no_backorders:
                     break
+        except Exception as e:
```
thomaspaulb, line 302: Actually I think it's wrong to catch exceptions outside the whole loop - we should:

1. loop
2. inside the loop, place a try that catches exception
3. if exception, handle
4. if no exception, commit

thomaspaulb (changes requested):

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
```diff
@@ -223,61 +226,96 @@ def button_validate(self):
...
+            exceptions_allowed = self.env["ir.config_parameter"].get_param(
+                "mrp_production_serial_matrix.allow_manufacturing_exceptions"
+            )
+            if exceptions_allowed and finished_count > 0:
+                _logger.error(e)
+                message = _(
+                    "Not all orders where produced because an exception occurred: "
```
thomaspaulb, line 312: were

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
```diff
@@ -223,61 +226,96 @@ def button_validate(self):
...
+                self.env.cr.commit()  # pylint: disable=invalid-commit
+                finished_count += 1
+                if no_backorders:
                     break
+        except Exception as e:
+            exceptions_allowed = self.env["ir.config_parameter"].get_param(
+                "mrp_production_serial_matrix.allow_manufacturing_exceptions"
+            )
+            if exceptions_allowed and finished_count > 0:
+                _logger.error(e)
```
thomaspaulb: We actually also need a `self.env.cr.rollback()` here because otherwise, the modifications of the last call of which the exception was caught, may still get committed, since we are catching the exception.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
ddejong-therp: Alright.
Can you explain what the benefit of this is though?
Thanks.

thomaspaulb: Hmmm, I'm actually not sure if it's needed to place the try/except inside the loop.

I guess it's only needed if you want to know on which serial or which MO it failed.

If the except is outside of the loop block those variables are lost.

thomaspaulb: In any case we do need to `break` the loop once an exception has happened. In the way you've done it currently, an explicit `break` is not necessary

mrp_production_serial_matrix/models/res_config_settings.py
ddejong-therp: Alright, I have now renamed the config parameter.

mrp_production_serial_matrix/models/res_config_settings.py
ddejong-therp: Ah yes I didn't realize this was already possible in 14.0
Changed it now.

mrp_production_serial_matrix/static/description/index.html
ddejong-therp: Alright, I have squashed all the commits, and removed these pre-commit changes.
I will add these later again when all other changes are done.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
ddejong-therp: Good one. Fixed that.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
ddejong-therp: Indeed.
Fixed this as well.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
ddejong-therp: woops, fixed it.

mrp_production_serial_matrix/wizards/mrp_production_serial_matrix.py
ddejong-therp: Yes, very fair. Fixed it.

PR #1758, opened 2026-03-26
[19.0][MIG] mrp_production_allow_recursive

PieterPaulussen:


CasVissers-360ERP:
@PieterPaulussen should we not inherit this as well? https://github.com/odoo/odoo/blob/e242b99b190ed060abc785ba0f3ef23f89f8b090/addons/mrp/models/mrp_bom.py#L136

AungKoKoLin1997:
I don't think we need to use this module anymore in odoo 19 because `_onchange_product_id` is removed and there is no more restriction to use the finished product as a component.

PR #1759, opened 2026-03-30
[16.0][ADD] mrp_subcontracting_serial_mass_produce

nobuQuartile:
## Summary

- When using Mass Produce on a subcontracting MO with serial-tracked finished
  products, the standard flow splits the MO and assigns serials but does not
  call `subcontracting_record_component()` on each split MO. As a result,
  `subcontracting_has_been_recorded` remains False and the finished serials are
  never synced to the receipt move lines.
- This module extends `stock.assign.serial._assign_serial_numbers()` to
  automatically call `subcontracting_record_component()` on each split MO
  after Mass Produce, enabling Mass Produce to work end-to-end for
  subcontracting orders.

@qrtl
QT6138


AungKoKoLin1997 (commented):
I think the current module is missing a `CONTEXT.md`. It should explain how this module fills a gap in standard Odoo behavior when the subcontracting operation type is unarchived, even though that is not the default Odoo setup.

I also do not think the module should be limited to mass production with multiple serial number entries. While the module mainly targets clunky serial assignment, it should also support manual serial number entry in the MO form, likely through an extension of _onchange_lot_producing.

Also, shouldn’t we propose this to OCB instead of creating a new custom module? Although the subcontracting operation type is archived by default, the issue still appears when a user unarchives it and performs operations on a subcontracted MO.

AungKoKoLin1997:
@nobuQuartile Please review the PR changes. After you review, please squash the commits and adjust PR title.

nobuQuartile:
LGTM: Code review and functional review

@AungKoKoLin1997 
Could you review this PR?

AungKoKoLin1997 (approved):
LGTM 👍

mrp_subcontracting_serial_mass_produce/wizard/stock_assign_serial_numbers.py
```diff
@@ -0,0 +1,39 @@
...
+    _inherit = "stock.assign.serial"
+
+    def _assign_serial_numbers(self, cancel_remaining_quantity=False):
+        if (
+            not self.production_id._get_subcontract_move()
+            or self.production_id.product_id.tracking != "serial"
+        ):
+            return super()._assign_serial_numbers(cancel_remaining_quantity)
+        serial_numbers = set(self._get_serial_numbers())
+        res = super()._assign_serial_numbers(cancel_remaining_quantity)
```
yostashiro:
```suggestion
        res = super()._assign_serial_numbers(cancel_remaining_quantity)
        if (
            not self.production_id._get_subcontract_move()
            or self.production_id.product_id.tracking != "serial"
        ):
            return res
        serial_numbers = set(self._get_serial_numbers())
```

mrp_subcontracting_serial_mass_produce/wizard/stock_assign_serial_numbers.py
```diff
@@ -0,0 +1,39 @@
...
+                and not mo.subcontracting_has_been_recorded
+                and mo._get_subcontract_move()
+                and mo.lot_producing_id
+                and mo.lot_producing_id in assigned_lots
+            )
+        )
+        for production in productions:
+            production.with_context(
+                cancel_backorder=False
+            ).subcontracting_record_component()
```
yostashiro, line 39: Otherwise, some MOs might be silently skipped in case strict consumption is set on BoM and there is qty discrepancy on a component. We want to process production anyway in such cases under the mass produce context.

```suggestion
            # Skip consumption warning since qty is set by _split_productions
            production.with_context(
                cancel_backorder=False, skip_consumption=True
            ).subcontracting_record_component()
```

Can you please add tests to confirm the intended behavior of these contexts?

yostashiro (changes requested):

mrp_subcontracting_serial_mass_produce/wizard/stock_assign_serial_numbers.py
AungKoKoLin1997: Updated!

mrp_subcontracting_serial_mass_produce/wizard/stock_assign_serial_numbers.py
AungKoKoLin1997: Done!

mrp_subcontracting_serial_mass_produce/tests/test_mrp_subcontracting_serial_mass_produce.py
```diff
@@ -0,0 +1,113 @@
...
+            "Serial numbers should be synced to receipt move lines",
+        )
+
+    def test_mass_produce_skips_strict_consumption_warning(self):
+        # Force a qty discrepancy by adding a 3rd BoM line AFTER the parent MO
+        # was created. The existing move_raw_ids does not include comp3, but
+        # _get_consumption_issues computes expected qty from the current BoM,
+        # so with consumption='strict' it would otherwise return an issue and
+        # subcontracting_record_component would silently return the
+        # consumption wizard without recording the MO.
```
yostashiro: To make the intent clearer.
```suggestion
        # Mass Produce records each split MO with skip_consumption=True, since the
        # consumed qty is set by _split_productions rather than the BoM. Without it,
        # a strict-consumption discrepancy makes subcontracting_record_component
        # return a consumption-wizard action that this automated flow discards,
        # silently leaving the MO unrecorded.
        # Here we fabricate that discrepancy by adding a BoM line after MO creation.
```

AungKoKoLin1997: Updated!

yostashiro (approved):
Code review. 👍

PR #1760, opened 2026-04-06
[18.0][ADD] mrp_tier_validation

CILC98:
This module introduces a tiered approval workflow for Manufacturing Orders (MRP) in Odoo. With this feature, manufacturing orders can require validation from multiple users or managerial levels before confirmation or processing, based on configurable rules and thresholds.

Key advantages:

Adds multi-level approval to the manufacturing process, increasing control and reducing unauthorized or erroneous order confirmations.
Ensures compliance with internal policies and audit requirements by enforcing structured approval steps.
Provides a clear audit trail of all approval actions, improving transparency and traceability.
Offers flexibility to adapt approval flows to different organizational structures and business needs.
This enhancement helps organizations strengthen their manufacturing process governance and reduce operational risks.


CILC98:
@LoisRForgeFlow Hi! Could you please help me by checking?

CILC98:
@LoisRForgeFlow Hi! Could you please help me by checking?

CILC98:
@pedrobaeza  Hi! Could you please help me by checking?

pedrobaeza:
You can review other PRs and ask in exchange that they review yours.

mrp_tier_validation/data/mrp_tier_validation_exception_data.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo noupdate="1">
+    <record id="mrp_tier_validation_exception" model="tier.validation.exception">
+        <field name="name">MRP EXCEPTION</field>
+        <field name="model_id" ref="mrp.model_mrp_production" />
+        <field
+            name="field_ids"
+            eval="[(6,0,[
```
celm1990: It’s more readable; see
https://github.com/odoo/odoo/blob/ed27aa65484aa3c36a5be5dfdd5b118666d41a18/odoo/fields.py#L4282
```suggestion
            eval="[Command.set([
```

mrp_tier_validation/models/mrp_production.py
```diff
@@ -0,0 +1,32 @@
...
+                    raise ValidationError(
+                        _(
+                            "This action needs to be validated for at least "
+                            "one record. \nPlease request a validation."
+                        )
+                    )
+            if rec.review_ids and not rec.validated:
+                raise ValidationError(
+                    _("A validation process is still open for at least one record.")
+                )
```
celm1990, line 33: I don’t understand why you did it.

The `base_tier_validation` module should handle this, right?

I imagine you need to add the attribute `_tier_validation_state_field_is_computed = True` because the state in mrp.production is computed. Please try it this way and let me know if it works in this case.

celm1990 (changes requested):
I got this error when trying to validate a production order. Please check.

Image

mrp_tier_validation/models/mrp_production.py
CILC98: For example, if I configure a tier definition to enforce that a product's lot/serial number is always specified, the validation is bypassed due to Odoo's automatic sequence assignment. A similar issue occurs if I create a validation requiring users to manually allocate times for operation types. Because these background processes inject data directly into the vals dictionary of the write method, relying solely on _tier_validation_state_field_is_computed = True is insufficient to trigger the validation.

Nevertheless, I will implement the recommended changes while retaining the proposed solution that prevents this specific inconvenience.

CILC98:
@celm1990 I've made the recommended adjustments. Your help with the review, please.

CILC98:
@celm1990 Hi!. Can you help me please?

mrp_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "MRP Tier Validation",
+    "summary": "Extends the functionality of Productions to "
+    "support a tier validation process.",
+    "version": "18.0.1.0.0",
+    "category": "Manufacturing/Manufacturing",
+    "website": "https://github.com/OCA/manufacture",
+    "author": "TRESCLOUD, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["mrp", "mrp_workorder", "base_tier_validation"],
```
celm1990: The `mrp_workorder` module is Enterprise. Please keep Community dependencies only; for this reason, the CI fails.

celm1990 (changes requested):
Please squash the commits

mrp_tier_validation/__manifest__.py
CILC98: Thank you very much Carlos, you are absolutely right

CILC98:
Hi @celm1990. Can you help me with the review, please?

CILC98:
Hi @celm1990. Can you help me with the review, please?

PR #1281, opened 2024-06-07
[14.0][ADD] mrp_kit_drop_ship

KKamaa:
A forward port from ForgeFlow's [mrp_kit_drop_ship](https://github.com/OCA/manufacture/pull/655)

PR #1761, opened 2026-04-08
[14.0][MIG] mrp_kit_drop_ship

thomaspaulb:
Without this module, if a product is being procured and has an existing Bill of Materials of type Kit, the components of the product are going to be procured, instead of the main product.

This module introduces the capability to ignore the existence of the kit during procurement, when the procurement is associated with a dropship. That is, when the product is going to be sourced from a third party vendor to the customer.

Upon confirmation of the purchase order the system will still explode the product into components within the drop ship picking.

Continuation of https://github.com/OCA/manufacture/pull/1281

Forward port of https://github.com/OCA/manufacture/pull/655


thomaspaulb:
Inferring approvals from the previous PR and merging:

/ocabot merge nobump

PR #1769, opened 2026-04-17
[18.0][FIX] mrp_production_location_picking_type: preserve picking type production location on recompute

JoanSForgeFlow:


AaronHForgeFlow (commented):
If the locations are computed why the methods _get_move_raw_values and _get_move_finished_values are needed?

JoanSForgeFlow:
Hi @AaronHForgeFlow,

  - _get_move_raw_values / _get_move_finished_values set the correct location when the moves are
  created. Core Odoo passes an explicit location there (from the product's property_stock_production),
   which wins over the compute. Without these overrides, new moves would be born with the wrong
  location.
  - _compute_location_id / _compute_location_dest_id keep existing moves in sync when the picking
  type (and thus the production location) changes afterwards.

AaronHForgeFlow (approved):
LGTM 👍  Thanks for the explanation

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1774, opened 2026-04-28
[16.0][FIX] mrp_bom_product_price_margin: uom change must change standard and total price

quentinDupont:
Bug raises by @SandieFavre 
If user changes uom on bom_line, it doesn't change its standard and total price
This PR fix that by calcuting new price with core function `_compute_price`

### DIFF
--- a/mrp_bom_product_price_margin/models/mrp_bom_line.py
+++ b/mrp_bom_product_price_margin/models/mrp_bom_line.py
@@ -22,15 +22,23 @@ class MrpBomLine(models.Model):
         digits=(16, 3),
     )
 
-    @api.depends("standard_price_unit", "product_qty")
+    @api.depends(
+        "standard_price_unit",
+        "product_qty",
+    )
     def _compute_standard_price_subtotal(self):
         for line in self:
             line.standard_price_subtotal = line.standard_price_unit * line.product_qty
 
-    @api.depends("product_id")
+    @api.depends("product_id", "product_id.uom_id", "product_uom_id")
     def _compute_standard_price_unit(self):
         for line in self:
-            line.standard_price_unit = line.product_id.standard_price
+            if not (line.product_id and line.product_id.uom_id and line.product_uom_id):
+                line.standard_price_unit = 0.0
+            else:
+                line.standard_price_unit = line.product_id.uom_id._compute_price(
+                    line.product_id.standard_price, line.product_uom_id
+                )
 
     @api.depends(
         "standard_price_subtotal", "bom_id.bom_line_ids.standard_price_subtotal"


mrp_bom_product_price_margin/models/mrp_bom_line.py
```diff
@@ -22,15 +22,29 @@ class MrpBomLine(models.Model):
...
     def _compute_standard_price_unit(self):
         for line in self:
-            line.standard_price_unit = line.product_id.standard_price
+            if not (line.product_id and line.product_id.uom_id and line.product_uom_id):
+                line.standard_price_unit = 0.0
+                continue
+            else:
+                line.standard_price_unit = line.product_id.uom_id._compute_price(
+                    line.product_id.standard_price, line.product_uom_id
+                )
```
CILC98, line 41: Hi! I think the "else" is no longer necessary because of the "continue" clause that was added earlier.

mrp_bom_product_price_margin/models/mrp_bom_line.py
```diff
@@ -22,15 +22,29 @@ class MrpBomLine(models.Model):
...
     )
 
-    @api.depends("standard_price_unit", "product_qty")
+    @api.depends(
+        "standard_price_unit",
+        "product_qty",
+        "product_id",
+        "product_id.uom_id",
+        "product_uom_id",
+    )
```
CILC98, line 28: I think adjusting the standard_price_unit calculation is sufficient. What do you think?

CILC98 (changes requested):

mrp_bom_product_price_margin/models/mrp_bom_line.py
```diff
@@ -22,15 +22,29 @@ class MrpBomLine(models.Model):
...
     def _compute_standard_price_subtotal(self):
         for line in self:
             line.standard_price_subtotal = line.standard_price_unit * line.product_qty
 
-    @api.depends("product_id")
+    @api.depends(
+        "standard_price_unit", "product_id", "product_id.uom_id", "product_uom_id"
```
CILC98:
```suggestion
        "product_id", "product_id.uom_id", "product_uom_id"
```
Hi! I believe it shouldn't depend on the same field that is being calculated.

CILC98 (changes requested):

mrp_bom_product_price_margin/models/mrp_bom_line.py
quentinDupont: yep ! at first i change this compute but you're right, this code is not useful

CILC98:
LGTM

CILC98 (approved):

quentinDupont:
Bug revealed by colleagues, i need to check later

quentinDupont:
> Bug revealed by colleagues, i need to check later

Bug was linked to other custom module so I un-draft this one

legalsylvain (approved):

BhaveshHeliconia (approved):

PR #1780, opened 2026-05-07
[17.0][FIX] mrp_production_serial_matrix_import_xlsx

DavidJForgeFlow:
If the imported file contains or matrix lines or the finished_lot_ids, the system should allow to import the data.

In line 202 the finished lots are updated, so we want to enter here if we have something to write.

### DIFF
--- a/mrp_production_serial_matrix_import_xlsx/models/mrp_production_serial_matrix.py
+++ b/mrp_production_serial_matrix_import_xlsx/models/mrp_production_serial_matrix.py
@@ -196,7 +196,7 @@ class MrpProductionSerialMatrix(models.Model):
             finished_lot_ids, component_templates, data_map
         )
 
-        if new_lines_vals:
+        if new_lines_vals or finished_lot_ids:
             self.line_ids = [(5, 0, 0)]
             self.line_ids = [(0, 0, v) for v in new_lines_vals]
             self.finished_lot_ids = [(6, 0, finished_lot_ids.ids)]


mrp_production_serial_matrix_import_xlsx/models/mrp_production_serial_matrix.py
```diff
@@ -196,7 +196,7 @@ def action_import_template(self):
             finished_lot_ids, component_templates, data_map
         )
 
-        if new_lines_vals:
+        if new_lines_vals or finished_lot_ids:
             self.line_ids = [(5, 0, 0)]
             self.line_ids = [(0, 0, v) for v in new_lines_vals]
```
AaronHForgeFlow, line 201: won't this line fail if new_lines_vals is None/Flase?

DavidJForgeFlow: _prepare_matrix_lines_from_data in L195 will return [] in that case

RicardCForgeFlow: Hi @AaronHForgeFlow I think that new_lines_vals can only be a list (empty in the worst case) 
https://github.com/OCA/manufacture/blob/fa56cc9d1e612638d8f1109ed44d2a4c3c23c5a8/mrp_production_serial_matrix_import_xlsx/models/mrp_production_serial_matrix.py#L147

Right?

RicardCForgeFlow (approved):

mrp_production_serial_matrix_import_xlsx/models/mrp_production_serial_matrix.py
AaronHForgeFlow: ahh ok, thanks,

AaronHForgeFlow (approved):
👍

BhaveshHeliconia (approved):

LoisRForgeFlow (approved):
👍

PR #1781, opened 2026-05-09
[IMP] quality_control_oca, various: Get translated model name in object_id

diggy128:
quality_control_oca, quality_control_stock_oca:
Refactor object_selection_values() in qc.inspection to get model and name from ir.model using env.ref
Model refs are taken from a dependent module.
This makes the selection dropdown in the UI translated (using the model name in ir.model).

As the reimplementation just changes how the models in these modules are inserted in the selection field, any overrides in custom modules still work.


diggy128:
@OCA/manufacturing-maintainers a quick review?

PR #1784, opened 2026-05-14
[IMP] mrp_multi_level: add date filters on planned order search view

soniak05:
Add predefined date filters (Today, Yesterday, Last 7 Days, Last 30 Days, Last 365 Days) on order_release_date field in the planned order search view.

### DIFF
--- a/mrp_multi_level/views/mrp_planned_order_views.xml
+++ b/mrp_multi_level/views/mrp_planned_order_views.xml
@@ -93,6 +93,35 @@
                     domain="[('mrp_planner_id', '=', uid)]"
                 />
                 <filter string="Fixed" name="fixed" domain="[('fixed','=',True)]" />
+                <separator />
+                <filter
+                    string="Released Today"
+                    name="release_today"
+                    domain="[('order_release_date', '=', context_today().strftime('%Y-%m-%d'))]"
+                />
+                <filter
+                    string="Released Yesterday"
+                    name="release_yesterday"
+                    domain="[('order_release_date', '=', (context_today() - datetime.timedelta(days=1)).strftime('%Y-%m-%d'))]"
+                />
+                <filter
+                    string="Released (Last 7 Days)"
+                    name="release_last_7_days"
+                    domain="[('order_release_date', '&gt;=', (context_today() - datetime.timedelta(days=6)).strftime('%Y-%m-%d')),
+                            ('order_release_date', '&lt;=', context_today().strftime('%Y-%m-%d'))]"
+                />
+                <filter
+                    string="Released (Last 30 Days)"
+                    name="release_last_30_days"
+                    domain="[('order_release_date', '&gt;=', (context_today() - datetime.timedelta(days=29)).strftime('%Y-%m-%d')),
+                            ('order_release_date', '&lt;=', context_today().strftime('%Y-%m-%d'))]"
+                />
+                <filter
+                    string="Released (Last 365 Days)"
+                    name="release_last_365_days"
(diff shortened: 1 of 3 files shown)


mrp_multi_level/views/mrp_planned_order_views.xml
aleboldro02, line 1: I don't think defining 5 time filters is the best approach. I recommend adding a grouping that does the same thing with less code.

aleboldro02 (changes requested):

mrp_multi_level/views/mrp_planned_order_views.xml
AdrianaSaiz: Hello @aleboldro02 we want filters, not grouping, those are diferent points of view and different usages.

Whats wrong with adding 5 filters? It doesn't impact performance, and we will have a better usability.

Your comment makes no sense at this point

AdrianaSaiz (approved):
LGTM great job, this will improve usage and help with record searches

alvaro-gmz (approved):
LGTM

mrp_multi_level/views/mrp_planned_order_views.xml
aleboldro02: Hello @AdrianaSaiz,
I think my comment makes perfect sense from a maintainability perspective, which is also one of the general OCA principles.

My point is not that filters are technically wrong, but that introducing 5 separate filters for a similar purpose increases XML verbosity and future maintenance effort, while a grouping could achieve a comparable result with less code.

If the decision is to prioritize UX over code simplicity, that's absolutely fine, but that is a tradeoff discussion — not a matter of the comment “making no sense”.

soniak05: Hello @aleboldro02 ,
A groupby isn't a valid alternative here because it doesn't do the same thing as a filter.  One groups, the other restricts the dataset. It's not a UX vs maintainability tradeoff, they're just different functionalities.

aleboldro02 (approved):

AdrianaSaiz:
Good morning @LoisRForgeFlow , could you have look at this pull request?
We consider it a good visual and usage improvement and its just one xml file

mrp_multi_level/views/mrp_planned_order_views.xml
```diff
@@ -93,6 +93,35 @@
...
                 />
                 <filter string="Fixed" name="fixed" domain="[('fixed','=',True)]" />
+                <separator />
+                <filter
+                    string="Released Today"
+                    name="release_today"
+                    domain="[('order_release_date', '=', context_today().strftime('%Y-%m-%d'))]"
+                />
+                <filter
+                    string="Released Yesterday"
```
LoisRForgeFlow, line 103: First, "released today/yesterday/etc." implies that the move has been released and this field is meant to display the deadline to release the requirement in the lines calling for procurement. So it would be more accurate "To release today"

mrp_multi_level/views/mrp_planned_order_views.xml
```diff
@@ -93,6 +93,35 @@
...
+                    name="release_yesterday"
+                    domain="[('order_release_date', '=', (context_today() - datetime.timedelta(days=1)).strftime('%Y-%m-%d'))]"
```
LoisRForgeFlow, line 105: Second, You will never get release dates in the past, that means you haven't run MRP today and even if that's the case it doesn't matter, anything that is due is late and you are going to miss the plan, so these filters are not really useful from my PoV

LoisRForgeFlow (changes requested):
I think there is a confusion on the purpose of that field and you might be misusing it. See my comments below.

PR #1783, opened 2026-05-13
[19.0][MIG] quality_control_oca: Migration to 19.0

Hotdgo9:
Standard Migration to 19.0.

Adjusted for Odoo 19.0 breaking changes:

* `uom.category` removed; `test_uom_category` rewired as a computed `Many2one('uom.uom')` walking up `relative_uom_id` to the root, with `uom_id` constrained via `child_of`.
* `res.groups.category_id` replaced by `privilege_id` via the new `res.groups.privilege` model. A Quality Control privilege record preserves the User/Manager radio in the user form.
* `res.groups.users` → `user_ids`; `res.users.groups_id` → `group_ids`.
* `<search>` view syntax: fields no longer allowed inside `<group>`; group-by `<group>` no longer accepts `expand=` or `string=`.
* `product.template.categ_id` no longer has a default; the test product now explicitly references `product.product_category_goods`.
* Lazy translation form for `self.env._()` calls; one f-string fix.

12 tests collected, 10 ran, 0 failures locally on `odoo:19.0` with and without demo data.

cc: @pedrobaeza @anajuaristi @LoisRForgeFlow @SimoneRubino

PR #1785, opened 2026-05-14
[19.0][MIG] quality_control_stock_oca: Migration to 19.0

Hotdgo9:
Standard Migration to 19.0.

## Dependencies

Depends on #1783 (`quality_control_oca` migration). The `test with Odoo` / `test with OCB` checks on this PR will fail until #1783 merges into 19.0, because the base module isn't yet available in OCA's 19.0 branch. `pre-commit`, `runboat/build`, and codecov checks should pass and can be reviewed in parallel. After #1783 merges, this PR rebases cleanly and CI flips green.

## 18 → 19 adjustments

* `read_group` is deprecated in 19.0 (`@api.deprecated`). Refactored both `_compute_count_inspections` call sites (`stock_picking.py` and `stock_production_lot.py`) to `_read_group(domain=..., groupby=[...], aggregates=["__count"])`, with tuple-result unpacking replacing the old dict-result iteration.
* `odoo.fields.first` is removed. Replaced `first(recordset)` with `recordset[:1]` slicing in `qc_inspection.py:_compute_lot`.
* `stock.move._action_confirm` gained a `create_proc=True` parameter in 19.0. The override now accepts and passes it through; otherwise Odoo core's internal `_create_backorder` call hits a `TypeError` when dispatched to the override.
* Search-view XPath simplified: `<group expand="0" position="inside">` → `<group position="inside">`, matching the parent `quality_control_oca` search view's 19.0-compliant shape (no `expand=`).
* Test-only: `move_ids_without_package` was renamed to `move_ids` on the stock.picking form view in 19.0; one `Form()` usage updated. Plus ruff UP031 f-string fixes for two existing `%`-format expressions.

19 tests collected, 17 ran, 0 failures locally on `odoo:19.0`, both with and without demo data.

Deployed and validated on a production Odoo 19 Community instance running alongside the rest of OCA/manufacture's 19.0 modules. Manually exercised the full auto-trigger workflow: configured a `qc.trigger.product_template_line` linking the auto-created `My Company: Receipts` trigger to a test, created and confirmed a receiving picking, observed auto-generated `qc.inspection` in `ready` state, completed the inspector workflow, verified `_compute_count_inspections` counters updated correctly.

cc: @pedrobaeza @anajuaristi @LoisRForgeFlow @SimoneRubino


Hotdgo9:
Heads-up: the `test with Odoo`, `test with OCB`, and `runboat/build` failures are all from the same root cause — the test environment's `pip install odoo-addon-quality_control_oca==19.0.*` resolves no matching version on PyPI, because #1783 hasn't merged yet (so the OCA bot hasn't published the 19.0 wheel for the base module). `pre-commit` and `Detect unreleased dependencies` pass as expected.

When #1783 lands and the OCA bot publishes `odoo-addon-quality_control_oca-19.0.1.0.0`, I'll rebase this branch on the updated `19.0` and the test jobs should turn green. Locally on `odoo:19.0` the test suite is 19 collected / 17 ran / 0 failed both with and without demo data, and the module has been running in a production Odoo 19 instance with the full trigger workflow exercised (auto-inspection on receiving picking confirm, inspector flow to `success`, `_compute_count_inspections` counters tracking correctly).

BhaveshHeliconia:
@Hotdgo9 can you add test-requirement.txt file for dependent pr.

BhaveshHeliconia (commented):
Tested locally functional and code review LGTM!

dalonsod:
> @Hotdgo9 can you add test-requirement.txt file for dependent pr.

@Hotdgo9 could you attend @BhaveshHeliconia 's suggestion? And also, please add #1803 to commit history

Hotdgo9:
@BhaveshHeliconia @dalonsod — both done:

**`test-requirements.txt`** added at the repo root so CI installs the dependency from #1783 until it merges:

```
odoo-addon-quality_control_oca @ git+https://github.com/OCA/manufacture@refs/pull/1783/head#subdirectory=quality_control_oca
```

I'll drop this file in the rebase once #1783 lands on `19.0`.

**#1803** — rebased onto current `18.0`, so `[IMP] quality_control_stock_oca: added picking origin to inspections` is replayed in history with authorship intact. The refresh also brought in `[IMP] ... add context to actions from lot and picking` and updated translations (both merged to 18.0 after this branch was cut). Neither needed a 19.0 change — `picking_origin` is a plain `related` Char, and the search-view additions reconcile cleanly with the `expand=` removal.

Local on `odoo:19.0`, no demo: `quality_control_stock_oca: 19 tests, 0 failed, 0 error(s) of 17 tests`; `pre-commit` green.

dalonsod:
Thanks! There's still an action in red, could you check which the problem is?

Hotdgo9:
The red check is **`Detect unreleased dependencies`**, and it's expected here — it's the direct result of the `test-requirements.txt` that lets the test jobs run. That job rejects any non-comment line containing a `/` in `requirements.txt`/`test-requirements.txt`:

```
grep "^[^#].*/" ${reqfile}   →   Unreleased dependencies found in test-requirements.txt.
```

so the `git+https://…@refs/pull/1783/head` line trips it by design. For a dependent PR this is the standard, unavoidable tradeoff while the parent is open: CI can either install the dependency from #1783 and run `test with Odoo` / `test with OCB` (green now), **or** pass `Detect unreleased dependencies` — not both, since the only line that wouldn't trip it is a released PyPI wheel, which won't exist until #1783 merges.

As a `Depends on #1783` PR this isn't mergeable before then anyway, and `test-requirements.txt` is a standalone commit that I'll drop in the rebase once #1783 lands — at which point every check, including this one, turns green.

I'd suggest leaving it as-is so the suite keeps running against the real dependency during review, but I'm happy to remove the file if you'd rather see `Detect unreleased dependencies` green in the meantime.

BhaveshHeliconia (approved):

Issue #1286, opened 2024-06-14
qc.test.object_id selection only offers invalid choices


cparadis777:
## Module
quality_control_oca

## Describe the bug
The first selection menu of the qc.test.object_id field has 5 "empty" (no selection text) choices: "S", "e", "t", "(", ")".
No matter which value is selected, trying to select a model from the second selection menu causes an Odoo server error to appear.

## To Reproduce
**Affected versions**: 14, 15, 16, 17
The same bug was reported previously, but marked stale:
V14: #823 
V12: #562 
I have reproduced the bug in runboat for V14, V15, V16 and V17 myself.

1. Create a new related test
2. Set the first selection of the object_id field
3. Try to selected a related model

**Expected behavior**
The user should be able to select a model from the second selection menu, and the options in the first selection menu should be visible.

**Additional context**

I don't have much experience with the module, but I checked the code and it seems the error would be located in QcTest.objectSelectionValue() method: 
https://github.com/OCA/manufacture/blob/4d44182ab4e5c5cbe5bb63e25d2281ed15f472e2/quality_control_oca/models/qc_test.py#L21

Odoo server error trace:
Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1748, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1775, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1979, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 222, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 725, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 24, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 20, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 534, in __getitem__
    return self.registry[model_name](self, (), ())
  File "/opt/odoo/odoo/modules/registry.py", line 213, in __getitem__
[...]

PR #1797, opened 2026-05-21
[17.0][FIX] quality_control_oca: Restore object_id on qc.test with dynamic model selection

miquelrosell99:
This PR restores the `object_id` field on `qc.test` that was present in 14.0/16.0 but removed in 17.0 due to a broken selection implementation (see #1286).

## What changed

- **qc.test**: Re-introduced `object_id` as a `Reference` field with a dynamic, extensible selection mechanism via `_get_object_id_models()`.
- **Trigger lines**: Added domain filtering in `qc.trigger.product_line` and `qc.trigger.product_template_line` so only generic tests (empty `object_id`) or tests linked to the same product/template can be selected.
- **Views**: Added `object_id` to qc.test form and tree views.

## Architecture

- `object_id` empty = generic test (can be used for any object).
- `object_id` set = product-specific drawing inspection plan.
- Submodules can extend `_get_object_id_models()` to add more model types.

Closes #1568

### DIFF
--- a/quality_control_oca/models/qc_test.py
+++ b/quality_control_oca/models/qc_test.py
@@ -32,6 +32,31 @@ class QcTest(models.Model):
         default=lambda self: self.env.company,
     )
 
+    @api.model
+    def _get_object_id_models(self):
+        """Return the list of models selectable for the object_id field.
+        Override this method to add more models.
+        :return: List of model technical names.
+        """
+        return ["product.product", "product.template"]
+
+    @api.model
+    def object_selection_values(self):
+        """Build the selection list for the Reference field from ir.model."""
+        res = []
+        for model_name in self._get_object_id_models():
+            model = self.env["ir.model"]._get(model_name)
+            if model:
+                res.append((model_name, model.name))
+        return res
+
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why not directly putting this into `quality_control_oca_product`? Do you see any other thing where this can be useful?

miquelrosell99:
> Why not directly putting this into `quality_control_oca_product`? Do you see any other thing where this can be useful?

Mainly to keep the "general" approach the 'object_id' field gave us. If it would be better to simply add many2one fields linking to products, that's ok for me since we only use quality controls scoped to products.

Please, confirm what is more aligned with OCAs approach.

pedrobaeza:
As this generic has proven to not be used, I would go for the direct m2o links in the menu. Simpler and limited to the scope of the module you use for avoiding new surprises in the future.

miquelrosell99:
> As this generic has proven to not be used, I would go for the direct m2o links in the menu. Simpler and limited to the scope of the module you use for avoiding new surprises in the future.

Understood. In that case I will close this PR. Thanks a lot!

PR #1807, opened 2026-06-09
[16.0][FIX] mrp_subcontracting_skip_no_negative: fix backorder creation

JordiMForgeFlow:
With the current logic, we are calling separately _action_done on stock move for subcontracted moves and other moves.

In a scenario where we process the subcontracted move but not the other moves, _action_done is called for stock moves with no qty done. In the standard method, no moves are considered to be done (https://github.com/odoo/odoo/blob/19027e4986a469eee7a9781c444bbd2a1e7aaea5/addons/stock/models/stock_move.py#L1833) and consequently no picking is assigned to be backordered (https://github.com/odoo/odoo/blob/19027e4986a469eee7a9781c444bbd2a1e7aaea5/addons/stock/models/stock_move.py#L1867).

In simple words, Odoo will not trigger the backordering logic if all moves that enter _action_done have no qty done.

To fix it we must consolidate the call to _action_done including all moves. However, in order to keep the no negative check split evaluation, we do it in the stock move line _action_done as it is eventually called in the stock move method (https://github.com/odoo/odoo/blob/19027e4986a469eee7a9781c444bbd2a1e7aaea5/addons/stock/models/stock_move.py#L1857)

The fix that was added in https://github.com/OCA/manufacture/pull/1467 was not covering this problem.

### DIFF
--- a/mrp_subcontracting_skip_no_negative/models/__init__.py
+++ b/mrp_subcontracting_skip_no_negative/models/__init__.py
@@ -1,2 +1,3 @@
 from . import stock_move
+from . import stock_move_line
 from . import stock_picking
--- a/mrp_subcontracting_skip_no_negative/models/stock_move.py
+++ b/mrp_subcontracting_skip_no_negative/models/stock_move.py
@@ -11,9 +11,7 @@ class StockMove(models.Model):
     _inherit = "stock.move"
 
     def _action_done(self, cancel_backorder=False):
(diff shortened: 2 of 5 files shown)


AungKoKoLin1997 (approved):
Thanks for the fix.
Code and Functional Review: LGTM

BhaveshHeliconia (approved):
LGTM!

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1812, opened 2026-06-15
[18.0][ADD] mrp_bom_synchronization_group

ThiagoMForgeFlow:
New module introducing the concept of a **BoM Synchronization Group**: a set of Bills of Materials that must share the same consumed components, regardless of their finished product or their operations/routings.

Features:
- Group model with its views and configuration.
- Global default setting, applied to new groups and overridable per group, to choose the behaviour when members diverge: warn and fix manually, or synchronize automatically.
- Manual mode: discrepancies are flagged on the group and on each BoM, with a wizard that previews the changes per BoM before applying them.
- Automatic mode: editing the components of any member propagates the change to the rest of the group.

Synchronization only touches components (operations are preserved) and exposes a hook to resolve the operation of synced lines in downstream modules.


quentinDupont:
Thanks for the work. What are the use cases, can you give some example ? :)

ThiagoMForgeFlow:
> Thanks for the work. What are the use cases, can you give some example ? :)

Thanks! The use case is keeping the components of several BoMs aligned when they only differ in their operations.
  
Sometimes the same product is produced through different operations and work centers, so you need separate BoMs with the same components but different operations. The problem: if you edit the components of one BoM and forget the other, they silently drift apart.

A BoM Synchronization Group links those BoMs and keeps only their components in sync (operations are ignored).

PR #1813, opened 2026-06-15
[17.0][ADD] mrp_bom_component_mass_change

AndreuOForgeFlow:
This module allows you to replace or remove a component in several bills of materials at once.

When a component must be changed, the wizard shows the bills of materials where it is used (excluding, when replacing, those that already contain the new component to avoid duplicated lines), lets the user select the ones to update, and then either removes the component or replaces it with another product (setting the new quantity) in a single step.


mrp_bom_component_mass_change/models/mrp_bom_line.py
```diff
@@ -0,0 +1,16 @@
...
+
+class MrpBomLine(models.Model):
+    _inherit = "mrp.bom.line"
+
+    def action_bom_component_mass_change(self):
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "mrp_bom_component_mass_change.mrp_bom_component_mass_change_action"
+        )
+        action["context"] = {"default_component_id": self.product_id.id}
```
GuillemCForgeFlow: my only thought of a possible improvement would be that you could access the wizard in two different ways:
1. by clicking on the exchange button in the bom line. it that's the case, then we should restrict so that the component to change should be fixed to the component it has been chosen for.
2. an additional menu where you can call the wizard with no restrictions, probably under `Manufacturing > Operations`. you'd have no limitations on which component to change
WDYT?

mrp_bom_component_mass_change/views/mrp_bom_views.xml
```diff
@@ -0,0 +1,22 @@
...
+    <record id="mrp_bom_form_view" model="ir.ui.view">
+        <field name="name">mrp.bom.form - mrp_bom_component_mass_change</field>
+        <field name="model">mrp.bom</field>
+        <field name="inherit_id" ref="mrp.mrp_bom_form_view" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='bom_line_ids']/tree/field[@name='manual_consumption']"
+                position="after"
+            >
+                <button
```
GuillemCForgeFlow, line 13: maybe, to simplify, it'd be interesting to only show the button if the component appears in another BoM?

GuillemCForgeFlow (commented):
really great structure and test coverage 🤖 
apart from suggestions we can debate, LGTM 👍🏿

mrp_bom_component_mass_change/models/mrp_bom_line.py
quentinDupont: I agree. Before reading the README, I was looking for a menu like this

quentinDupont (approved):
Tested on runboat with success, great idea, great work 👌

mrp_bom_component_mass_change/models/mrp_bom_line.py
AndreuOForgeFlow: @GuillemCForgeFlow @quentinDupont, just added the new menu under `Manufacturing > Operations`:

image

And made the selected component on the wizard fixed (readonly True) when selected from the new button on the BoM line

mrp_bom_component_mass_change/views/mrp_bom_views.xml
AndreuOForgeFlow: Done :+1: Good idea

BhaveshHeliconia (approved):
LGTM!

quentinDupont:
before merging, i prefer to ask : @GuillemCForgeFlow you are good with your comments right ?

mrp_bom_component_mass_change/models/mrp_bom_line.py
```diff
@@ -0,0 +1,34 @@
...
+    @api.depends("product_id")
+    def _compute_component_in_multiple_boms(self):
+        for line in self:
+            line.component_in_multiple_boms = (
+                bool(line.product_id)
+                and self.env["mrp.bom"].search_count(
+                    [("bom_line_ids.product_id", "=", line.product_id.id)]
+                )
+                > 1
+            )
```
GuillemCForgeFlow: instead of doing this search for each line: `bom_line_ids.product_id`, you could search via a `read_group` using the `count` aggregator outside the loop and also doing: `env["mrp.bom.line"].search(["product_id", "in", self.ids])`
this is important in large databases where you may have lots of lines per BoM

GuillemCForgeFlow (changes requested):

GuillemCForgeFlow:
> before merging, i prefer to ask : @GuillemCForgeFlow you are good with your comments right ?

I see one final thing we should adjust for performance purposes before merging, then good for me 👍🏿

mrp_bom_component_mass_change/models/mrp_bom_line.py
AndreuOForgeFlow: Done, @GuillemCForgeFlow !

GuillemCForgeFlow (approved):
lgtm 👍🏿  code and functional review, great work 🤖

GuillemCForgeFlow:
@quentinDupont fine for me to merge 😄

PR #1817, opened 2026-06-22
[18.0][FIX] mrp_multi_level: make scheduling tests date-independent

bosd:
## Problem

`mrp_multi_level`'s scheduling tests (`TestMrpMultiLevel`) compute their expected
MRP inventory buckets with `calendar.plan_days(...)` (working days) anchored on
`datetime.today()`. Whether a given offset lands on the asserted bucket depends on
which weekday the suite runs on, so the suite is **not deterministic across run dates**.

This makes CI flap with the calendar:

- Green on Wed **2026-06-17**, red on Mon **2026-06-22** — with **no code change** in between.
- Failure: `test_04_mrp_multi_level` → `AssertionError: 0 != 1` (`len(fp_2_line_2)`), because a
  one-working-day shift moves the FP-2 supply out of the asserted `date_10` bucket.

The `18.0` branch has been red since 2026-06-22 for this reason, and it red-lights
unrelated PRs (the whole repo suite runs per PR).

## Fix

Freeze the shared test setup — and, via `addClassCleanup`, the test methods — to a fixed
working day (`2024-06-12`, a Wednesday), so `plan_days` math is deterministic regardless
of when CI runs. No production code is touched; this is a test-only change.

`freezegun` is already used in this repo (`mrp_stock_move_actual_date`).

## Test

`oca_run_tests` locally (via `act`) on `18.0`:
- Before: `1 failed, 0 error(s) of 218 tests` (`test_04_mrp_multi_level`).
- After: `0 failed, 0 error(s) of 218 tests` ✅

### DIFF
--- a/mrp_multi_level/tests/common.py
+++ b/mrp_multi_level/tests/common.py
@@ -3,14 +3,27 @@
 
 from datetime import datetime, timedelta
 
+from freezegun import freeze_time
+
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
+
+# The scheduling assertions in these tests compare MRP inventory buckets against
+# dates computed with ``calendar.plan_days`` (working days). Whether a given
+# offset lands on a particular bucket depends on which weekday "today" is, so the
+# suite is not deterministic across run dates (e.g. it passes mid-week but fails
+# on Mondays). Freeze the whole suite to a fixed working day so both the setup
+# and the planning run see the same reference date.
+FREEZE_DATE = "2024-06-12"  # Wednesday
 
 
 class TestMrpMultiLevelCommon(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        freezer = freeze_time(FREEZE_DATE)
+        freezer.start()
+        cls.addClassCleanup(freezer.stop)
         cls.mo_obj = cls.env["mrp.production"]
         cls.po_obj = cls.env["purchase.order"]
         cls.product_obj = cls.env["product.product"]


mrp_multi_level/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "MRP Multi Level",
-    "version": "18.0.1.4.2",
+    "version": "18.0.1.4.3",
```
BhaveshHeliconia: I don’t think we need to bump the version manually, as it should be updated automatically when the merge command is applied.

JordiMForgeFlow:
@bosd why not using the freeze_time decorator?

```
from freezegun import freeze_time

@freeze_time("2024-06-12")
class TestClass(BaseCommon):
```

mrp_multi_level/__manifest__.py
bosd: Good point — reverted the manual version bump; the merge bot will handle it.

bosd:
@JordiMForgeFlow I tried that first, but the `@freeze_time` class decorator breaks the run here.

freezegun's class decorator wraps the whole class lifecycle as roughly:

```python
setUpClass    = lambda: (self.start(), orig_setUpClass())
tearDownClass = lambda: (orig_tearDownClass(), self.stop())
```

So time is frozen *around* `TransactionCase.setUpClass()` (which opens the class
cursor/savepoint) and around the teardown/rollback. In this suite that leaves a lock on
`mrp_production` that isn't released in time, and the **next** module to install
(`mrp_stock_move_actual_date`) dies while altering that table:

```
ERROR odoo.sql_db: bad query: ALTER TABLE "mrp_production" ADD COLUMN "actual_date" date ...
ERROR: canceling statement due to lock timeout
CRITICAL odoo.service.server: Failed to initialize database `odoo`.
```

I reproduced this deterministically with `oca_run_tests` locally:

- `@freeze_time` decorator → `Failed to load registry` on `mrp_stock_move_actual_date` (lock timeout).
- explicit `freezer.start()` after `super().setUpClass()` + `addClassCleanup(freezer.stop)`
  → `0 failed, 0 error(s) of 218 tests`.

The difference is that starting the freeze *after* `super().setUpClass()` and stopping it via
`addClassCleanup` (which runs **before** `TransactionCase`'s own rollback, LIFO) keeps the DB
setup/teardown outside the frozen window, so the lock is released normally. That's why I kept
the explicit start/stop form rather than the decorator.

BhaveshHeliconia (approved):

JordiMForgeFlow (approved):

mrp_multi_level/tests/common.py
```diff
@@ -3,14 +3,27 @@
...
+from freezegun import freeze_time
+
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
 
+# The scheduling assertions in these tests compare MRP inventory buckets against
+# dates computed with ``calendar.plan_days`` (working days). Whether a given
+# offset lands on a particular bucket depends on which weekday "today" is, so the
+# suite is not deterministic across run dates (e.g. it passes mid-week but fails
+# on Mondays). Freeze the whole suite to a fixed working day so both the setup
```
LoisRForgeFlow, line 15: @bosd  I'm testing today (Monday) locally the test without this patch, and they passed. Any other condition to reproduce?

bosd:
@LoisRForgeFlow yes — the trigger is **running the full repo test suite in one database** (the way `oca_run_tests` / CI does), not running `mrp_multi_level`'s tests on their own. In isolation they pass on any day (that's what you saw); the failure only shows up when the other addons are co-installed and perturb the shared demo-data MRP dates, which — combined with the working-day `plan_days` math — pushes the FP-2 supply out of the `date_10` bucket on certain weekdays.

I just reproduced it on **current `origin/18.0` (136dccdcb), today (Mon 2026-07-20), no patch**, via `oca_run_tests`:

```
FAIL: TestMrpMultiLevel.test_04_mrp_multi_level
  File ".../mrp_multi_level/tests/test_mrp_multi_level.py", line 99, in test_04_mrp_multi_level
    self.assertEqual(len(fp_2_line_2), 1)
AssertionError: 0 != 1
1 failed, 0 error(s) of 221 tests
```

It's also weekday-dependent: green mid-week, red on Mondays (the branch was red on Mon 2026-06-22 for exactly this — the green push-CI runs since have all landed on non-Mondays, so it hasn't been re-triggered on a "bad" date until now).

So to reproduce: install the whole repo (or at least `mrp_multi_level` together with the other `mrp_multi_level*` / mrp addons) and run the tests on a Monday. The patch fixes it independently of *which* co-installed addon perturbs the dates, by freezing the suite's reference date so `plan_days` is deterministic. I'm happy to bisect the exact triggering addon if that's useful for you.

LoisRForgeFlow (approved):
Thanks for the additional details 👍 

/ocabot merge patch

LoisRForgeFlow:
@bosd can you forward port?

bosd:
Forward port to 19.0: #1834 (verified locally with `oca_run_tests`: mrp_multi_level 27 tests, 0 failed of 74).

PR #1823, opened 2026-07-07
[18.0][FIX] mrp_production_check_bom_alignment: tolerate rounding noise in qty

ThiagoMForgeFlow:
The quantity checks recomputed the expected component/by-product demand and compared it to the move demand with a strict equality. But when the MO/BoM quantity ratio is not an integer, the scaled quantity is a repeating decimal, and Odoo rounds the stored move demand differently depending on the path that wrote it:  rounds up to the UoM rounding, while  (the standard Update BoM) stores an unrounded value later rounded to the Product Unit of Measure precision. These differ by one rounding unit, producing a false quantities are not aligned warning that even the wizard's Update MO and Confirm cannot clear. Allow a tolerance of one unit of the coarsest applicable rounding.

@ForgeFlow

### DIFF
--- a/mrp_production_check_bom_alignment/models/mrp_production.py
+++ b/mrp_production_check_bom_alignment/models/mrp_production.py
@@ -91,13 +91,8 @@ class MrpProduction(models.Model):
             expected_qty = bom_line.product_uom_id._compute_quantity(
                 bom_line.product_qty * factor, move.product_uom
             )
-            if (
-                float_compare(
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
Nice corner case, makes sense 👍

JoanSForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

PR #1828, opened 2026-07-15
[18.0][IMP] mrp_production_check_bom_alignment: detail misaligned items in warning

ArnauCForgeFlow:
The BoM alignment warnings only stated that the MO was not aligned with its BoM, without saying which items were off, forcing the user to compare the MO against the BoM by hand. Now the message adds which component is not aligned


BhaveshHeliconia (approved):

cgarcia-solvos (approved):
Code review. LGTM!

LoisRForgeFlow (approved):
/ocabot merge patch

PR #1839, opened 2026-08-03
[18.0][ADD] mrp_workorder_pause_note: New module

cgarcia-solvos:
Adds a field to the MRP Workorder model to store
the last note given by the operator when pausing that workorder.


BhaveshHeliconia (approved):
Looks good!

PR #1841, opened 2026-08-07
[18.0][FIX] mrp_production_picking_type_from_route: resolve the route on create

kanda999:
`mrp.production.create` assigns the default operation type and takes the order reference from its sequence before the compute methods run. A manufacturing order created outside of the form therefore kept the default operation type and its sequence, while the components location had already been computed from the operation type the route points to. The stored operation type and the reference then disagreed with the flow the order actually followed.

Resolve the manufacture rule in `create` as well, so orders generated by procurement, by the scheduler or from an operation type kanban get the same operation type as the ones created from the form.

@qrtl QT6792

Assisted-by: Claude Opus 5.0


mrp_production_picking_type_from_route/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "MRP Production Picking Type From Route",
     "summary": "Updates the operation type creating MO based on the product",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
```
AungKoKoLin1997, line 6: Let ocabot handles that.

AungKoKoLin1997 (approved):
LGTM

PR #1842, opened 2026-08-11
[18.0][ADD] mrp_workcenter_productivity_calendar: add new module

CristianoMafraJunior:
## Summary

Work order time tracking (Manufacturing Order > work order > "Time Tracking" tab) always counts the raw elapsed time between `date_start` and `date_end` for productive/performance entries, even when the work center has a resource calendar with breaks configured. An activity logged from 10:00 to 14:00 counts as 4 hours even if the work center's calendar has a 12:00-13:00 lunch break.

Odoo core already uses the calendar for non-productive loss types (blocking reasons); this module extends the same logic to productive/performance time, opt-in per work center.

## Configuration

1. Manufacturing > Configuration > Work Centers, open a work center.
2. Set "Working Hours" to a resource calendar with the shift/break periods to apply.
3. Enable "Use Calendar for Time Tracking Duration".
4. Double check the calendar's Timezone field matches where the work center actually operates — a mismatch shifts the computed duration by the timezone offset.

cc @marcelsavegnago @douglascstd @kaynnan @WesleyOliveira98


mrp_workcenter_productivity_calendar/models/mrp_workcenter_productivity_loss.py
```diff
@@ -0,0 +1,34 @@
...
+                )
+            )
+            if use_calendar:
+                # get_work_hours_count uses the calendar's own tz; the
+                # resource-based alternative used for non-productive types
+                # relies on the work center's resource tz instead, which
+                # is hidden from the UI and easy to misconfigure.
+                hours = workcenter.resource_calendar_id.get_work_hours_count(
+                    date_start, date_stop
+                )
```
WesleyOliveira98: Could you keep the core call (super) for the non-productive case and only use get_work_hours_count for the new productive/performance case? Or if the tz unification is intentional, worth calling out explicitly in the README as a behavior change, not just as a "double-check your tz" config tip.

WesleyOliveira98 (changes requested):

mrp_workcenter_productivity_calendar/models/mrp_workcenter_productivity_loss.py
```diff
@@ -0,0 +1,34 @@
...
+                duration = max(duration, hours * 60)
+            else:
+                duration = max(
+                    duration, (date_stop - date_start).total_seconds() / 60.0
+                )
+        return round(duration, 2)
```
BhaveshHeliconia, line 40:
```suggestion
    def _convert_to_duration(self, date_start, date_stop, workcenter=False):
        use_calendar = (
            workcenter
            and workcenter.resource_calendar_id
            and workcenter.use_calendar_for_duration
        )
        calendar_records = (
            self.filtered(lambda p: p.loss_type in ("productive", "performance"))
            if use_calendar
            else self.browse()
        )
        other_records = self - calendar_records

        duration = 0
        if calendar_records:
            hours = workcenter.resource_calendar_id.get_work_hours_count(date_start, date_stop)
            duration = hours * 60

        if other_records:
            res = super(
                MrpWorkcenterProductivityLoss, other_records
            )._convert_to_duration(date_start, date_stop, workcenter=workcenter)
            duration = max(duration, res)

        return round(duration, 2)
```

BhaveshHeliconia (commented):
Hi @CristianoMafraJunior, to address @WesleyOliveira98's point, we can refactor _convert_to_duration to delegate non-productive loss types back to super() and restrict the calendar calculation to productive/performance types when enabled.

mrp_workcenter_productivity_calendar/models/mrp_workcenter_productivity_loss.py
CristianoMafraJunior: done

mrp_workcenter_productivity_calendar/models/mrp_workcenter_productivity_loss.py
CristianoMafraJunior: done

BhaveshHeliconia (approved):
Code review LGTM!

WesleyOliveira98 (approved):

marcelsavegnago (approved):
LGTM

PR #1848, opened 2026-08-18
[18.0][FIX] mrp_production_picking_type_from_route: keep op type once confirmed

ThiagoMForgeFlow:
The compute reassigned the operation type whenever `bom_id` changed, so Update BoM on a confirmed MO repointed it and its component locations while the transfers and reservations already created stayed on the old ones.


LauraCForgeFlow (approved):
Makes sense! 👍

JoanSForgeFlow (approved):

BhaveshHeliconia (approved):

PR #1851, opened 2026-08-21
[18.0][IMP] quality_control*: refresh and unify module icons

maiolif:
## Before / after

`quality_control_oca` was still on the pre-v16 icon template (gradient square,
diagonal drop-shadow band). Its two extension modules,
`quality_control_mrp_oca` and `quality_control_stock_oca`, had already moved
to a different icon set back in 2015 (4e7886f) and were never brought back in
sync with the base module — today the three don't match each other.

This redraws all three onto a single flat icon (gear + check, no background
tile) shared across the family, in the same "flat overlapping shapes, no
container" style Odoo itself uses for its own app icons (Repair, Email
Marketing, Manufacturing, Inventory, ...).

## What changed

- `quality_control_oca/static/description/{icon.svg,icon.png}` — replaced
- `quality_control_mrp_oca/static/description/{icon.svg,icon.png}` — icon.svg added (didn't exist before), icon.png replaced
- `quality_control_stock_oca/static/description/{icon.svg,icon.png}` — icon.svg added (didn't exist before), icon.png replaced

`icon.png` is 128×128 for all three now (the base module's was previously 140×140).

Artwork is original, hand-built SVG — no stock/third-party assets involved.


pedrobaeza (commented):
Thanks for the icon refreshing. The new one seems OK, but I still prefer to have a slight variation of the derivative ones, combining the main icon with the subarea - also refreshed - (stock, mrp, etc), in small in one corner.

Can you please do it this way?

PR #1852, opened 2026-08-27
[17.0][FIX] mrp_production_serial_matrix: keep component reservations sane

DavidJForgeFlow:
Two fixes for the stock the matrix left behind.

**A failed serial number was committed.** The errors raised while validating an MO reach the matrix from a flush, so `except UserError` caught them with the stock changes already applied and no rollback; writing the exception state and posting it to the chatter flushed again and committed them. `ValidationError` derives from `UserError`, so negative stock constraints were caught this way too: the operation was reported as an error and persisted at the same time, and the next serial number started from the already negative quantity. Each serial number now runs inside its own savepoint, with the recursion outside it so the ones that succeeded are kept.

Seen in the field with `stock_no_negative` installed: a component lot walked down to -4, -6, -8, -10, -12, -14 over six consecutive failed validations, each one recorded in the chatter and committed.

**Surplus component lines were consumed.** `_reserve_lot_in_move` accepted any availability above zero and ignored what it actually managed to reserve; a selected lot already on the move but short of the needed quantity was never topped up; and a leftover line had its lot cleared, which made Odoo consume a lot-tracked component as untracked stock. The needed quantity is now required and verified, the shortfall is reserved per lot, and any line not on a selected lot is dropped instead of consumed.

### DIFF
--- a/mrp_production_serial_matrix/models/mrp_production_serial_matrix.py
+++ b/mrp_production_serial_matrix/models/mrp_production_serial_matrix.py
@@ -432,28 +432,50 @@ class MrpProductionSerialMatrix(models.Model):
         to_remove = lots_current - lots_desired
         if to_remove:
             current_lines.filtered(lambda ml: ml.lot_id in to_remove).unlink()
-        to_add = lots_desired - lots_current
-        for lot in to_add:
-            if move.product_id.tracking == "serial":
-                qty_needed = 1.0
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
sounds sensible 👍

MarinaAForgeFlow (approved):

PR #1836, opened 2026-07-25
[18.0][FIX] stock_whole_kit_constraint: remove dead code with deprecated qty_done

rrebollo:
The `_check_backorder_moves` method in stock_whole_kit_constraint contained two unreachable loops iterating over move_line_ids filtered with `move_id=False`. Since move_line_ids is a One2many inverse field, it never returns lines without a parent move — making both loops dead code.

The second loop also referenced `pack.qty_done` on `stock.move.line`, a field removed in Odoo 17 (replaced by `stock.move.line.quantity`). Removing the dead code eliminates this deprecated field reference.

The method's remaining logic (computing quantity_todo and quantity_done from move.product_uom_qty and move.quantity) is functionally intact.

Reference: OCA OpenUpgrade 17.0 upgrade_analysis.txt for stock module.

@BinhexTeam T23802

PR #1856, opened 2026-09-07
[18.0][ADD] mrp_production_cancel_confirm

DavidJForgeFlow:
Cancelling a manufacturing order only asks for a plain confirmation in the form view, and for nothing at all in the list view, so an intended update is easily turned into a cancellation, with no trace of who did it or why.

The wizard is opened only when the cancellation comes from the Cancel button of the form or list views, so the flows that cancel orders on their own, such as deleting an order or cancelling a subcontracted receipt, keep working.


mrp_production_cancel_confirm/models/mrp_production.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import fields, models
+
+
+class MrpProduction(models.Model):
+    _name = "mrp.production"
+    _inherit = ["mrp.production", "base.cancel.confirm"]
+
+    _has_cancel_reason = "optional"  # ["no", "optional", "required"]
+
+    cancel_confirm = fields.Boolean(default=False)
```
BhaveshHeliconia:
```suggestion
    cancel_confirm = fields.Boolean(default=False, copy=False)
```

BhaveshHeliconia (approved):
Functional and code review LGTM!

Just minor suggestion

rrebollo (approved):
Code Review. LGTM!

Would you be so kind to review mine in return? #1836

mrp_production_cancel_confirm/models/mrp_production.py
DavidJForgeFlow: Hi @BhaveshHeliconia, copy is inherited from https://github.com/OCA/server-ux/blob/18.0/base_cancel_confirm/model/base_cancel_confirm.py#L22

PR #1858, opened 2026-09-09
[18.0][IMP] mrp_production_date_planned_finished: cover the security lead time with a test

carlosdauden:
Until v15 the onchange that back-computes the start date from the manually edited end date also subtracted `company_id.manufacturing_lead`, mirroring core, which added that security lead when computing the end date from the start one. Since v16 the end date is a stored compute that only adds the BoM manufacturing lead time, so dropping that subtraction on the way to 18.0 was right: putting it back makes core recompute the end date from the newly shifted start date, and the order ends `manufacturing_lead` days before the date the user asked for, which is precisely what this module exists to avoid. Reproduced on 18.0 with a security lead time of 5 days and a BoM lead time of 2 days: a requested end date of 2026-10-10 10:00 comes back as 2026-10-05 10:00.

No behaviour changes here. This only adds the regression test that was missing, since no existing test configured a non-zero security lead time and the question was therefore untested, plus a comment on the method and a note in the usage documentation explaining why the company security lead time is deliberately left out.

@Tecnativa
ping @sergio-teruel @juancarlosonate-tecnativa @CarlosRoca13

### DIFF
--- a/mrp_production_date_planned_finished/models/mrp_production.py
+++ b/mrp_production_date_planned_finished/models/mrp_production.py
@@ -11,6 +11,14 @@ class MrpProduction(models.Model):
     _inherit = "mrp.production"
 
     def _get_date_start_using_delays(self):
+        # This has to stay the exact inverse of core's `_compute_date_finished`,
+        # which is a stored compute depending on `date_start`: anything else
+        # subtracted here is undone right away by that recompute, and the order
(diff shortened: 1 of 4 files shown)


juancarlosonate-tecnativa (approved):

ivantodorovich (approved):
/ocabot merge patch

PR #1865, opened 2026-09-14
[ADD] mrp_exception: introduce base_exception to manufacturing

rfindling-gpc:
[ADD] mrp_exception: introduce base_exception to manufacturing

This commit introduces a new module integrating `base_exception` with `mrp.production`.

When a user attempts to complete a Manufacturing Order (via `button_mark_done`), the framework evaluates all active exception rules for the model. If exceptions are triggered, the MO is blocked and a resolution wizard opens. Exception Managers can bypass non-blocking exceptions, which correctly resumes the interrupted completion workflow.

Includes test coverage verifying rule detection, reverse relational mapping, wizard lifecycle, and blocking constraints.


rfindling-gpc:
forgot oca_dependencies.txt file. closing to resubmit.

PR #1868, opened 2026-09-17
[16.0][IMP] mrp_subcontracting_skip_no_negative: prevent considering linked MOs that will not be done

JordiMForgeFlow:


mrp_subcontracting_skip_no_negative/models/stock_move.py
```diff
@@ -21,7 +21,14 @@ def _action_done(self, cancel_backorder=False):
...
             )
-            productions = production_moves.production_id
+            # Only a production that is about to produce can consume components
+            # here. On a partial receipt where the MO has already been split off
+            # and validated, the backorder MO should be filtered out from the check.
+            productions = production_moves.production_id.filtered(
+                lambda p: not float_is_zero(
+                    p.qty_producing, precision_rounding=p.product_uom_id.rounding
+                )
+            )
```
BhaveshHeliconia, line 31:
```suggestion
            productions = production_moves.production_id
            recorded_productions = productions.filtered(lambda p: p._has_been_recorded())
            if recorded_productions:
                productions = recorded_productions.filtered(
                    lambda p: not float_is_zero(
                        p.qty_producing, precision_rounding=p.product_uom_id.rounding
                    )
                )
```
In standard subcontracting, when components are not pre-recorded through the Record Components wizard, stock.picking._action_done() calls super()._action_done() before setting qty_producing on the subcontracted MOs.

So when stock_move._action_done() runs, production.qty_producing is still 0.0. The float_is_zero() check then filters out those productions, so the component negative stock check is skipped for standard subcontracting receipts.

